PR #50, opened 2018-11-21
uninstall for installer syncer

ilackarms:
also should prevent reinstalling the same mesh over and over
untested


pkg/install/install_syncer.go
```diff
@@ -243,3 +264,13 @@ func getMeshObject(install *v1.Install) (*v1.Mesh, error) {
 	}
 	return mesh, err
 }
+
+func uninstallHelmRelease(releaseName string) error {
+	helmClient, err := helm.GetHelmClient()
+	if err != nil {
+		return err
+	}
```
rickducott: This doesn't clean up any additional resources, such as CRDs, CRBs

api/v1/install.proto
```diff
@@ -34,6 +34,10 @@ message Install {
     HelmChartLocator chartLocator = 6;
 
     Encryption encryption = 7;
+
+    // whether or not this install should be enabled
+    // if disabled, corresponding resources will be uninstalled
+    bool enabled = 12;
```
rickducott: this should default to true

ilackarms: there's a proto tag for that

pkg/install/install_syncer.go
ilackarms: i left this to a post/pre install method on the interface

api/v1/install.proto
ilackarms: fixed

ilackarms:
added an assertion in the e2e that  tests uninstall

pkg/install/consul/consul_installer.go
```diff
@@ -45,6 +45,14 @@ func getOverrides(encryption *v1.Encryption) string {
 	return strings.Replace(overridesYaml, "@@MTLS_ENABLED@@", strBool, -1)
 }
 
+func (c *ConsulInstaller) DoPreHelmUninstall() error {
+	return nil
+}
+
+func (c *ConsulInstaller) DoPostHelmUninstall() error {
+	return nil
```
rickducott: Needs to remove CRB

PR #67, opened 2018-11-21
wip: cli fixes

ilackarms:
Signed-off-by: Scott Weiss <sdw35@cornell.edu>

### DIFF
--- a/cli/pkg/cmd/create/routing_rule.go
+++ b/cli/pkg/cmd/create/routing_rule.go
@@ -2,6 +2,7 @@ package create
 
 import (
 	"fmt"
+	"github.com/solo-io/supergloo/cli/pkg/setup"
 	"strings"
 
 	"github.com/solo-io/supergloo/cli/pkg/common"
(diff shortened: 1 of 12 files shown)


cli/pkg/cmd/create/routing_rule.go
```diff
@@ -24,12 +25,15 @@ func RoutingRuleCmd(opts *options.Options) *cobra.Command {
 		Short: `Create a route rule with the given name`,
 		Long:  `Create a route rule with the given name`,
 		Args:  cobra.ExactArgs(1),
-		Run: func(c *cobra.Command, args []string) {
+		RunE: func(c *cobra.Command, args []string) error {
+			if err := setup.InitKubeOptions(opts); err != nil {
```
mitchdraft: Looks like we're `InitKubeOptions` in each of our subcommands. Can we instead call it once from the root command?

I believe you are doing this because we want to delegate initialization to a specific command and not automatically try to set up the supergloo process every time (and thereby the first time) the user calls a command.

As an alternative, we can trigger initialization by a flag passed by the user at the root command. Then we only do initialization when they run `supergloo --init`

cli/pkg/cmd/top.go
```diff
@@ -39,12 +38,18 @@ func App(version string) *cobra.Command {
...
 	)
 
-	setup.InitCache(&opts)
+	return app
+}
 
-	err := setup.Init(&opts)
-	if err != nil {
-		panic(errors.Wrap(err, "Error during initialization."))
+func withInit(command *cobra.Command, initFunc func() error) {
```
mitchdraft: this isn't used anywhere

hack/install/supergloo.yaml
```diff
@@ -42,4 +42,26 @@ spec:
       containers:
       - name: supergloo
         image: soloio/supergloo:dev
-        imagePullPolicy: IfNotPresent
+        imagePullPolicy: Always
```
mitchdraft: was this intended for a different pr? seems unrelated to the cli

rickducott:
I think we should be closing this at this point? @ilackarms @mitchdraft

hack/install/supergloo.yaml
rickducott: This is useful for testing locally but we don't want to merge it in. We can now use the script @EItanya made in `hack/install/get_latest.sh`

PR #72, opened 2018-11-22
minor fixes for copy/paste comments

joeholley:
Not sure if L#73 should reference linkerd or consul, but it probably needs to be changed too?


ilackarms:
thanks for the catch! :sweat_smile:

Issue #71, opened 2018-11-22
Linkerd installation: defaulting to a namespace


ilevine:
Linkerd installation ignore user choice of namespace and installing always on the same namespace.

PR #81, opened 2018-11-26
Finish uninstall

rickducott:


rickducott:
Finishes up and tests install/uninstall. 
1. Uninstall removes resources specific to the mesh, so install -> uninstall -> install should now work for all mesh types.
2. Refactors the e2e tests. The installer tests just test install/uninstall per mesh, and the e2e tests test everything else (but don't need to test uninstall). 

Some things that aren't included in this PR:
1. Multiple installs of the same mesh type are unsupported. Install has a few known issues, in particular the helm chart for Istio deploys CRDs and we are looking at a solution for moving that out of there. Install also uses a default name for certain configs (like cluster role binding or webhook configs), so uninstalling one mesh will uninstall some resources needed by other meshes of the same type. 
2. Linkerd uses a hard-coded namespace, so a temporary method on MeshInstaller has been exposed to account for this, so that the correct namespace is created/cleaned up. This will be removed when the hard-coded namespace is no longer used.

pkg/install/consul/consul_installer_test.go
```diff
@@ -50,8 +48,8 @@ var _ = Describe("Consul Installer", func() {
 								},
 							},
 						},
-						Encryption: &v1.Encryption{
-							TlsEnabled: mtls,
+						Enabled: &types.BoolValue{
```
rickducott: These tests only test install/uninstall, the tests inside test/e2e test encryption (and other features)

pkg/install/linkerd2/linkerd2_installer.go
```diff
@@ -16,6 +15,10 @@ func (c *Linkerd2Installer) GetDefaultNamespace() string {
 	return defaultNamespace
 }
 
+func (c *Linkerd2Installer) UseHardcodedNamespace() bool {
+	return true
```
rickducott: Temporary solution until https://github.com/solo-io/supergloo/issues/71 is fixed.

pkg/install/install_syncer.go
```diff
@@ -276,12 +279,28 @@ func getMeshObject(install *v1.Install, releaseName string) (*v1.Mesh, error) {
...
 	helmClient, err := helm.GetHelmClient(ctx)
 	if err != nil {
 		return err
 	}
 	_, err = helmClient.DeleteRelease(releaseName, helmlib.DeletePurge(true))
 	helm.Teardown()
-	return err
+	// Install may be into ns that can't be deleted, don't propagate error if delete fails
+	syncer.tryDeleteInstallNamespace(getInstallNamespace(install, meshInstaller.GetDefaultNamespace()))
+	// TODO: this will break if there are more than one installs of a given mesh that depend on the CRB
```
rickducott: #82

pkg/install/istio/istio_installer.go
```diff
@@ -91,10 +100,30 @@ func (c *IstioInstaller) AddSccToUsers(users ...string) error {
 	return err
 }
 
-func (c *IstioInstaller) DoPreHelmUninstall() error {
+func (c *IstioInstaller) DoPostHelmUninstall() error {
+	// TODO: this will break if there are more than one installs using these CRDs
```
rickducott: #82

test/e2e/istio_e2e_test.go
```diff
@@ -127,52 +124,45 @@ var _ = Describe("Istio Install and Encryption E2E", func() {
...
+				Kube:           util.GetKubeClient(),
+				SecretClient:   secretClient,
+				IstioNamespace: installNamespace,
+			}
+			syncSnapshot := getTranslatorSnapshot(mesh, secret)
+			err = meshSyncer.Sync(context.TODO(), syncSnapshot)
+			Expect(err).NotTo(HaveOccurred())
 
-	It("Can install istio without mtls enabled", func() {
-		snap := getSnapshot(false, nil)
```
rickducott: This test was removed because it is redundant with the install/uninstall tests in the *_installer_test

rickducott:
Ran the following tests to verify
* consul_installer_test
* istio_installer_test
* linkerd2_installer_test
* consul_e2e_test
* istio_e2e_test

pkg/install/consul/consul_installer_test.go
```diff
@@ -74,25 +72,34 @@ var _ = Describe("Consul Installer", func() {
...
 		util.TerminateNamespaceBlocking(installNamespace)
-		meshClient.Delete(superglooNamespace, meshName, clients.DeleteOpts{})
-		util.TerminateNamespaceBlocking("supergloo-system")
 	})
 
-	It("Can install consul with mtls enabled", func() {
+	It("Can install and uninstall consul", func() {
 		snap := getSnapshot(true)
 		err := syncer.Sync(context.TODO(), snap)
 		Expect(err).NotTo(HaveOccurred())
```
rickducott: Check that there's at least one pod (or all of them)

pkg/install/install_syncer.go
```diff
@@ -31,15 +33,16 @@ type InstallSyncer struct {
 	Kube           *kubernetes.Clientset
 	MeshClient     v1.MeshClient
 	SecurityClient *security.Clientset
+	CrdCLient      *crdClient.ApiextensionsClient
 }
 
 type MeshInstaller interface {
 	GetDefaultNamespace() string
+	UseHardcodedNamespace() bool
```
rickducott: Add a TODO to remove when #71 is fixed

pkg/install/istio/istio_installer.go
rickducott: We should get rid of deleting Istio CRDs here when the routing syncer merges in, and / or when we separate installing istio helm chart from installing istio CRDs

pkg/install/linkerd2/linkerd2_installer_test.go
```diff
@@ -75,15 +76,31 @@ var _ = Describe("Linkerd2 Installer", func() {
...
+		util.TerminateNamespaceBlocking("linkerd") //hard-coded in chart
 	})
 
-	It("Can install linkerd2", func() {
+	It("Can install and uninstall linkerd2", func() {
 		snap := getSnapshot(true)
 		err := syncer.Sync(context.TODO(), snap)
 		Expect(err).NotTo(HaveOccurred())
-		util.WaitForAvailablePods("linkerd") //hard-coded in chart
+		util.WaitForAvailablePods("linkerd")
```
rickducott: have this take length

test/util/test_helpers.go
```diff
@@ -66,6 +69,17 @@ func GetKubeClient() *kubernetes.Clientset {
 	return client
 }
 
+func GetCrdClient() *client.ApiextensionsClient {
```
rickducott: API extensions client

pkg/install/linkerd2/linkerd2_installer_test.go
rickducott: it was easier to have it return the number of pods, which we can wrap in an expect

rickducott:
Updated based on in-person PR feedback from @ilackarms, re-ran tests. Merging in when CI check passes.

PR #87, opened 2018-11-27
Cli uninstall

EItanya:


hack/install/supergloo.yaml
```diff
@@ -42,7 +42,7 @@ spec:
       containers:
       - name: supergloo
         image: soloio/supergloo:dev
-        imagePullPolicy: Always
```
rickducott: Oops, this shouldn't be committed...

rickducott:
Tested after rebasing and this works. Rough commands to make this work:
```
hack/install/recompile.sh
update supergloo.yaml so that image pull policy for supergloo is "IfNotPresent"
kubectl apply -f hack/install/supergloo.yaml
make install-cli
supergloo install -s -m istio -n istio-system
supergloo uninstall
kubectl logs supergloo-f6698dc58-chncd -n supergloo-system
supergloo install -s -m istio -n istio-system
```

note that recompile then applying the yaml to update the image pull policy, then deploying, is critical. supergloo init uses a different yaml file, with the image pull policy to `Always`. This means the result of recompile gets overwritten, and we're testing old code. Once the new code is in place, the uninstall command works as expected and install afterwards works too.

test/e2e/istio_e2e_test.go
```diff
@@ -199,7 +199,7 @@ var _ = Describe("Istio Install and Encryption E2E", func() {
 			return bookinfons
 		}
 
-		It("Should install istio and enable policy", func() {
```
rickducott: Oops

Issue #83, opened 2018-11-26
Istio install requires istio-system namespace


rickducott:
Repro: run istio_installer_test with a different namespace than istio-system, get the following error:

```
<*errors.withStack | 0xc00297a440>: {
          error: {
              cause: {
                  Code: 2,
                  Message: "release istio failed: namespaces \"istio-system\" not found",
                  Details: nil,
                  XXX_NoUnkeyedLiteral: {},
                  XXX_unrecognized: nil,
                  XXX_sizecache: 0,
              },
              msg: "Error installing helm chart",
          },
          stack: [0x22b2a99, 0x22b2482, 0x22b2283, 0x22b5532, 0x1fc855c, 0x1fc81cc, 0x1fc752f, 0x1fcedb6, 0x1fce5c7, 0x1fd41df, 0x1fd3ee8, 0x1fd350f, 0x1fd59bf, 0x1fd7c17, 0x1fd79a9, 0x22b1d53, 0x10fad7f, 0x105e051],
      }
      Error installing helm chart: rpc error: code = Unknown desc = release istio failed: namespaces "istio-system" not found
  not to have occurred
```

Issue #57, opened 2018-11-21
Istio encryption syncer needs to ensure default root cert is removed


rickducott:
Currently we write to cacerts, but need to make sure the default cert is being removed and cert rotation occurs. This should also be added to istio_e2e_test.go

rickducott:
Blocked on #100, should be quick fix once that is resolved.

PR #92, opened 2018-11-27
istio routing working e2e

ilackarms:
Signed-off-by: Scott Weiss <sdw35@cornell.edu>


test/e2e/istio_routing_e2e_test.go
```diff
@@ -1,91 +1,284 @@
...
 	"github.com/solo-io/supergloo/pkg/api/v1"
 	"github.com/solo-io/supergloo/pkg/setup"
+	"github.com/solo-io/supergloo/test/utils"
 )
 
-var _ = PDescribe("istio routing E2e", func() {
+var _ = Describe("istio routing E2e", func() {
+	var namespace, releaseName string
+	path := os.Getenv("HELM_CHART_PATH_ISTIO")
+	if path == "" {
```
rickducott: Can we just use the default (s3 chart) here?

rickducott:
updated the branch addressing my feedback, also removed the uninstall hook since it isn't used anymore. ran all the install tests and they are fine, and the CLI install/uninstall also works. need to get some PRs in other repos merged in before routing e2e works.

rickducott:
This PR should be good to merge in. The routing E2E passes consistently. It has been pulled into it's own suite because it has some assumptions that break other tests in the e2e suite. The installer tests all pass. The other e2e's are mostly good, though we'll continue to work on them as we fix known issues #97 #57 #83 etc... 

The e2e's utilizing discovery require access to a recent revision of a private solo repo

PR #96, opened 2018-11-28
Cli toggle mtls2

mitchdraft:
cherry picked https://github.com/solo-io/supergloo/pull/80
also addressed comments and replaced strings with constants

### DIFF
--- a/cli/pkg/cmd/install/common.go
+++ b/cli/pkg/cmd/install/common.go
@@ -15,7 +15,7 @@ import (
 func installationSummaryMessage(opts *options.Options) {
 	fmt.Printf("Installing %v in namespace %v.\n", opts.Install.MeshType, opts.Install.Namespace)
 	if opts.Install.Mtls {
-		fmt.Printf("MTLS active.\n")
+		fmt.Printf("mTLS active.\n")
 	}
 	return
 }
@@ -56,10 +56,10 @@ func qualifyFlags(opts *options.Options) error {
 		}
 		//if iop.Mtls {
 		//	if iop.SecretRef.Name == "" {
-		//		return fmt.Errorf("please specify a secret name to use MTLS")
+		//		return fmt.Errorf("please specify a secret name to use mTLS")
 		//	}
 		//	if iop.SecretRef.Namespace == "" {
-		//		return fmt.Errorf("please specify a secret namespace to use MTLS")
+		//		return fmt.Errorf("please specify a secret namespace to use mTLS")
 		//	}
 		//}
 		return nil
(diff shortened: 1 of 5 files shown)


mitchdraft:
this now allows you to use the default secret during install w/ mTLS
fyi @rickducott there's a bit of extra diffing since I replaced a local definition of ResourceRef with the real one

cli/pkg/cmd/install/common.go
```diff
@@ -28,7 +29,8 @@ func getNewInstallName(opts *options.Options) string {
 
 func getMetadataFromOpts(opts *options.Options) core.Metadata {
 	return core.Metadata{
-		Name:      getNewInstallName(opts),
+		Name: getNewInstallName(opts),
```
rickducott: nit: The spacing here suggests the code should be formatted, which I have warnings about as a pre-commit hook. You could set this up by running `make init`

cli/pkg/cmd/meshtoolbox/mtls/root.go
```diff
@@ -0,0 +1,221 @@
...
+	superglooV1 "github.com/solo-io/supergloo/pkg/api/v1"
+	"github.com/spf13/cobra"
+)
+
+// we have not yet implemented mtls updates
+// if you are working on mTLS updates, set DEV_MTLS to true
+// when you implement mTLS updates, remove the placeholder
+// comment: consider passing this setting as a dev flag
+// pros: easier to configure
+// cons: may confuse users when shown in help msg, might not add a lot of value
```
rickducott: my 2c, I like this as is. It's very easy to dev on the CLI by flipping this flag and running `make install`, agree with cons

cli/pkg/cmd/meshtoolbox/mtls/root.go
```diff
@@ -0,0 +1,221 @@
...
+		} else {
+			mesh.Encryption.TlsEnabled = !mesh.Encryption.TlsEnabled
+
+		}
+	default:
+		panic(fmt.Errorf("Operation %v not recognized", operation))
+	}
+
+	// 4. write the changes
+	writtenMesh, err := (*meshClient).Write(mesh, clients.WriteOpts{OverwriteExisting: true})
```
rickducott: Some work to be figured out on the backend whether this is updated in the mesh or install crd. I think it's right to update the mesh crd, but changing this from false to true could require a helm redeploy, which only really is invoked during the install syncer currently. i think we leave as-is until this is sorted out in the backend

mitchdraft:
I just threw in two small tweaks, that should be all for now

PR #98, opened 2018-11-29
Fixes to istio security 

rickducott:
Still not getting non-self-signed certs to work


rickducott:
This PR contains a few incremental improvements:
* Fixes helm overrides for istio installation, so that security / mtls is actually enabled when flagged on at install time. 
* Adds a test for deploying istio with mtls but uses the default certs, to validate that the default mesh policy and destination rule is set up for mTLS. 
* Extracts the secret syncing for Istio (where the cacerts secret is configured) so that it can be used both during install, and post-install. 
* Makes the test for installing istio with custom certs pending, until the citadel startup issue is debugged

rickducott:
A change was reverted on this branch that will be necessary in solo-kit to get custom certs with empty cert-chains to work. The basic gist: soloo-kit omits default / empty strings when serializing to kubernetes, and citadel's pod will fail to start because it won't be able to mount the cert chain file. Since custom certs are not working yet, this PR is going to merge (with the vendor override reverted) and the solo-kit change will not be made until we get custom certs working.

PR #105, opened 2018-11-30
Install test refactor / consolidation

### LINKED ISSUE #104: Consolidate install logic
* CLI / installers should know about default charts, which should be a const in the code somewhere.
* Install e2e tests should be moved to the same suite, to reduce duplicated code

rickducott:
Fixes #104


rickducott:
Now all install / uninstall e2es are run from install_suite_test.go, ran it to confirm tests all pass.

PR #107, opened 2018-11-30
Appmesh support

ilackarms:


cli/pkg/cmd/create/aws_secret.go
```diff
@@ -0,0 +1,138 @@
...
+	sOpts := &(opts.Create).AwsSecret
+
+	secret := &gloov1.Secret{
+		Metadata: core.Metadata{
+			Name:      sOpts.Name,
+			Namespace: sOpts.Namespace,
+		},
+		Kind: &gloov1.Secret_Aws{
+			Aws: &gloov1.AwsSecret{
+				// these can be read in from ~/.aws/credentials by default (if user does not provide)
```
rickducott: I added this comment but never implemented this feature. We should implement!

test/e2e/istio_routing_e2e_test.go
```diff
@@ -0,0 +1,341 @@
+package e2e
```
rickducott: I think this got added due to a bad merge, the real file should be `test/e2e/routing/istio_routing_e2e_test.go`, which is also on the branch. The diff between the two is small.

PR #109, opened 2018-11-30
Helm upgrade

rickducott:
The Istio chart has been updated to rerun the security update job after helm upgrades. The chart lives in the supergloo-helm repo and has been updated on s3. 

This PR makes a few contributions: 
* The install loop calls helm update now with the latest overrides when there is an install CRD that changes and already has a mesh CRD. 
* Wires toggling mTLS for istio into the CLI by updating the install CRD
* E2E tests that show plain text curl commands failing when strict mTLS is toggled on. 

As a follow up task, we could wire up this change into the mesh loop, and implement this without helm. The effect of this helm upgrade is simply to toggle `mode: PERMISSIVE` off and on in the default meshpolicy resource.


rickducott:
The helm charts are now moved to a separate repo (supergloo-helm) and so removed from this PR.

rickducott:
Updated

rickducott:
Besides e2e tests, confirmed that the CLI updates the mtls value and triggers the update code path by looking at the CRD and supergloo logs.

PR #118, opened 2018-12-03
Fail fast if we cannot connect to the kubernetes cluster

marcogschmidt:
Since we currently do not set any timeouts on our kubernetes clients, it takes the CLI a long time before returning an error in case the kubeconfig is not pointing to a running cluster. This is frustrating for the user.

I added a connection attempt with a 1s timeout before we start our usual init steps to avoid this.

### DIFF
--- a/cli/pkg/cmd/uninstall/common.go
+++ b/cli/pkg/cmd/uninstall/common.go
@@ -2,12 +2,13 @@ package uninstall
 
 import (
 	"fmt"
+	"strings"
+
 	"github.com/gogo/protobuf/types"
 	"github.com/solo-io/solo-kit/pkg/api/v1/clients"
 	"github.com/solo-io/supergloo/cli/pkg/cmd/options"
 	"github.com/solo-io/supergloo/pkg/api/v1"
 	"github.com/solo-io/supergloo/pkg/constants"
-	"strings"
 )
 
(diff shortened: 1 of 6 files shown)


marcogschmidt:
@rickducott You're right. It seems a `dep` process was hanging locally. I ran it again and the dependencies are fixed. Updated the PR accordingly.

rickducott:
Should have guessed, option 3 -- dep being dep ;)

cli/pkg/cmd/uninstall/common.go
```diff
@@ -123,21 +121,21 @@ func uninstallMeshes(meshList []string, installClient *v1.InstallClient) error {
 	return err
 }
 
-func updateMeshInstall(installCrd *v1.Install, installClient *v1.InstallClient) error  {
-	installCrd.Enabled = &types.BoolValue{Value:false}
-	_, err := (*installClient).Write(installCrd, clients.WriteOpts{OverwriteExisting:true})
+func updateMeshInstall(installCrd *v1.Install, installClient *v1.InstallClient) error {
```
rickducott: This function name should be more like `uninstallMesh` or `disableMeshInstall`

marcogschmidt:
Changed the method name (and also a bug in the error logging in the same file)

PR #121, opened 2018-12-04
language fix

EItanya:


cli/pkg/cmd/install/specific.go
```diff
@@ -87,18 +87,15 @@ func generateAppMeshInstallSpecFromOpts(opts *options.Options) *v1.Mesh {
...
 
 	prompt := &survey.MultiSelect{
-		Message: "Which namespaces should this mesh watch:",
+		Message: fmt.Sprintf("Which namespace(s) would you like the new %s mesh to have access to: ", meshName),
 		Options: opts.Cache.Namespaces,
 	}
 
 	chosenNamespaces := []string{}
-	// survey.AskOne(prompt, &chosenNamespaces, nil)
 	if err := survey.AskOne(prompt, &chosenNamespaces, survey.Required); err != nil {
```
rickducott: If we're leaving this code here, I don't think `survey.Required` is correct. We either want it to be an optional (and make sure it's clear that not selecting namespaces -> cluster-wide), or we want to ask a bool question first whether we want to restrict the mesh to only watch certain namespaces for CRD changes. The latter may be a bit more clear from a user perspective

EItanya: good point, fixed

PR #142, opened 2018-12-17
version check and version def script

mitchdraft:
this pr:
- updates the checkpoint lib to do version checks
- adds a script to generates a version spec in such a way that it can be imported

question:
- do we want to give an error if someone builds/publishes a non standard version (anything other than XX.YY.ZZ)?


hack/version/main.go
```diff
@@ -0,0 +1,83 @@
...
+	if err := generateVersionFile(tag, relativeFilename); err != nil {
+		log.Fatal(err)
+	}
+
+	fmt.Printf(tag)
+
+}
+
+func getGitTag() (string, error) {
+	cmd := exec.Command("git", "describe", "--tags")
```
rickducott: This will fail if there are no tags, but that's fine

pkg/version/generated.go
```diff
@@ -0,0 +1,5 @@
+package version
+
+// Code generated by go run hack/version/main.go. DO NOT EDIT.
+
+const Version = "v0.1.0-45-gf4ae4c8"
```
rickducott: I think this is out of date -- there's a 0.1.1 release (and we've decided to switch over to using tags without the "v") so locally I have the version as 0.1.1-2-gc6300b5

I'm kinda worried that this version is going to be out of date every time we do a release, is there a chicken and egg problem where you need to regenerate after tagging the release, but regenerating leads to a code change that increments the version...? 

This seems like an incremental improvement but we might want to instead leave Version undefined in (non-generated) code and then override the variable with a flag provided by the make file, along the lines of:
https://www.reddit.com/r/golang/comments/4cpi2y/question_where_to_keep_the_version_number_of_a_go/
https://stackoverflow.com/questions/11354518/golang-application-auto-build-versioning

PR #150, opened 2019-01-22
Fix formatting in CLI reference

christian-posta:
Formatting on the CLI page for the `flags` is all messed up. Seems the mkdocs site generator is pretty picky about the `md` formatting. I've fixed it in this PR

### DIFF
--- a/doc/docs/cli.md
+++ b/doc/docs/cli.md
@@ -66,6 +66,7 @@ supergloo uninstall [flags]
 ```
 
 Flags:
+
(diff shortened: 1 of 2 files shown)


ilackarms:
ah! thank you @christian-posta 
mkdocs can be annoying, we originally considered using [Sphinx](http://www.sphinx-doc.org/en/master/) but their markdown support and was lacking. Envoy use Sphinx + `.rst` for their docs

Issue #156, opened 2019-01-28
Make Supergloo Image Configurable


danehans:
The Supergloo image is not configurable, making it difficult for developers outside Solo.io to build and test.

PR #157, opened 2019-01-28
Refactors cli init command to support configurable superglooo image

danehans:
Previously, the Supergloo image was hard-coded in the installation manifest.

- Updates Makefile to remove hard-coded image.
- Refactors init cli command to create install resources using client-go instead of os.Exec kubectl commands.
- Removes the install manifest since resources are being created directly.
- Provides image name customization by exposing a cli flag named "-image".

Fixes issue #156


danehans:
@ilackarms I see you made changes to the init manifest, so I rebased and updated the PR to support init sg and disc image flags.

cli/pkg/cmd/initsupergloo/initsupergloo.go
```diff
@@ -15,26 +31,187 @@ func Cmd(opts *options.Options) *cobra.Command {
...
+func initSupergloo(opts *options.Options) {
+	initHelm(opts)
+	createKubeNamespace(opts, constants.GlooNamespace)
+	createKubeNamespace(opts, constants.SuperglooNamespace)
+	createKubeClusterRoleBindings(opts, constants.SuperglooClusterRoleBindings)
+	createKubeDeployment(opts, "supergloo")
+	createKubeDeployment(opts, "discovery")
 
-	kubectl := "kubectl"
-	args := []string{"apply", "-f", "https://raw.githubusercontent.com/solo-io/supergloo/master/hack/install/supergloo.yaml"}
```
ilackarms: imo what would be preferable here is not to deprecate the old behavior. at the very least, i'd be in favor of adding a flag to the opts like `--custom-install` that calls a `initSuperglooCustom` function here, and default to using the old method

keep in mind that we'll be moving to a helm chart (or some other source of truth) at some point, which we can use to support these install customizations.

cli/pkg/cmd/top.go
```diff
@@ -72,11 +72,5 @@ func App(version string) *cobra.Command {
 		os.Exit(1)
 	}
 
-	err = setup.InitSupergloo(&opts)
-	if err != nil {
-		fmt.Println(errors.Wrap(err, "Error during initialization!"))
-		os.Exit(1)
-	}
-
```
ilackarms: thanks for this! we definitely don't want this here

ilackarms:
@danehans overall looks good, but i'd like to preserve the original behavior as the default (`kubectl apply -f <manifest url>) and put the new custom behavior behind a flag

thanks for removing that forced init in `top.go`.. it definitely shouldn't be there

ilackarms:
@danehans closing this PR for now

the supergloo rewrite has been merged to master and we're currently using a helm chart for installation. images can be overridden using custom `values.yaml` file, which can be an arg for the cli or a `helm install`

PR #158, opened 2019-01-29
fix case where labels are empty, bump discovery version

ilackarms:
currently discovery is misconfigured and set to an old version in the install yaml

this fixes discovery to deploy properly and scan all namespaces
also fixes an edge case where a service has no labels and it causes the routing syncer to write invalid destination rules


pkg/translator/istio/mesh_routing_syncer.go
```diff
@@ -183,6 +183,9 @@ func destinationRulesForUpstreams(rules v1.RoutingRuleList, meshes v1.MeshList,
 		for host, labelSets := range labelsByHost {
 			var subsets []*v1alpha3.Subset
 			for _, labels := range labelSets {
+				if len(labels) == 0 {
```
rickducott: We can unit test this

PR #165, opened 2019-02-06
install without tiller

ilackarms:
this PR starts us on the path to being able to install a helm chart without tiller

implementing 3 functions:
- render manifests from helm templates
- create kube resources from rendered manifests
- delete kube resources from rendered templates

tested with istio helm chart


pkg/install/helm/helm_chart_installer_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/solo-io/supergloo/pkg/install/helm"
+)
+
+var istioCrd = apiextensions.CustomResourceDefinition{}
+
+var _ = Describe("HelmChartInstaller", func() {
+	It("installs from a helm chart", func() {
```
rickducott: Nit: there is a file based (instead of url) code path that isn’t getting covered here

ilackarms: i just didn't want to commit a hard-coded file. i can amend this test to `wget` that file and then use file uri

PR #166, opened 2019-02-07
add istio install helper, improve crd wait

ilackarms:
this PR does two things:
- improve the helm chart installer; make CRD waits more authoritative (do an actual apiserver request to verify the crd exists)
- create a helper method to install versioned manifests for istio (tested with 1.0.3 and 1.0.5) with relevant helm value overrides

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -5,13 +5,12 @@ steps:
   env: ['PROJECT_ROOT=github.com/solo-io/supergloo']
   id: 'dep'
 
-# Disabled until new supergloo artifacts are due for release
-#- name: 'gcr.io/cloud-builders/go'
-#  args: ['build', '-o', 'supergloo', 'github.com/solo-io/supergloo/cmd']
-#  env:
-#    - 'PROJECT_ROOT=github.com/solo-io/supergloo'
-#    - 'GOPATH=/workspace/gopath'
-#  id: 'compile'
+- name: 'gcr.io/cloud-builders/go'
+  args: ['build', '-o', 'supergloo', 'github.com/solo-io/supergloo/cmd']
+  env:
+    - 'PROJECT_ROOT=github.com/solo-io/supergloo'
+    - 'GOPATH=/workspace/gopath'
+  id: 'compile'
 
 - name: gcr.io/cloud-builders/gcloud
   args:
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -5,12 +5,13 @@ steps:
   env: ['PROJECT_ROOT=github.com/solo-io/supergloo']
   id: 'dep'
 
-- name: 'gcr.io/cloud-builders/go'
-  args: ['build', '-o', 'supergloo', 'github.com/solo-io/supergloo/cmd']
-  env:
-    - 'PROJECT_ROOT=github.com/solo-io/supergloo'
-    - 'GOPATH=/workspace/gopath'
-  id: 'compile'
+# Disabled until new supergloo artifacts are due for release
+#- name: 'gcr.io/cloud-builders/go'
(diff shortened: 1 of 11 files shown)


pkg/install/helm/helm_chart_installer.go
```diff
@@ -96,24 +102,53 @@ func ApplyManifests(ctx context.Context, namespace string, manifests []manifest.
...
+	for _, crd := range crds {
+		crdName := crd.Name
+		err = retry.Do(func() error {
+			crd, err := crdClientset.ApiextensionsV1beta1().CustomResourceDefinitions().Get(crdName, v1.GetOptions{})
+			if err != nil {
+				return errors.Wrapf(err, "validating crd %v registration failed", crdName)
+			}
 
+			contextutils.LoggerFrom(ctx).Info("registered crd %v", crd.ObjectMeta)
+			return nil
```
yuval-k: is think you should also test the status, right?
You can have a situation where a crd exists, but not yet ready and custom resources cannot be created

pkg/install/istio/istio_installer_test.go
```diff
@@ -24,13 +24,13 @@ import (
 
 var T *testing.T
 
-func TestIstioInstaller(t *testing.T) {
+func XTestIstioInstaller(t *testing.T) {
```
rickducott: Why X?

ilackarms: i thought it would be a problem to have 2 test suites in the same package, but i guess not

cloudbuild.yaml
```diff
@@ -5,12 +5,13 @@ steps:
   env: ['PROJECT_ROOT=github.com/solo-io/supergloo']
   id: 'dep'
 
-- name: 'gcr.io/cloud-builders/go'
-  args: ['build', '-o', 'supergloo', 'github.com/solo-io/supergloo/cmd']
-  env:
-    - 'PROJECT_ROOT=github.com/solo-io/supergloo'
-    - 'GOPATH=/workspace/gopath'
-  id: 'compile'
+# Disabled until new supergloo artifacts are due for release
```
rickducott: Can we uncomment this now that the PR is ready?

ilackarms: i'd prefer not to as it adds 5min to every CI run

would you be opposed to leaving it disabled until we're ready to start shipping supergloo binaries again?

i'd also be open to investigating boiling down the time this step takes (5 min seems excessive for a compile)

rickducott: I'm opposed to turning it off. I looked at the CI runs and I'm not sure what problem you're trying to solve by turning it off -- compile runs in parallel to the tests, which take longer than the time to compile. 

[For instance, a build with it turned off](https://console.cloud.google.com/cloud-build/builds/4e3942de-ef7f-40d4-9bd6-1c0d53dd38dd?project=solo-public&angularJsUrl=%2Fcloud-build%2Fbuilds%2F4e3942de-ef7f-40d4-9bd6-1c0d53dd38dd%3Fproject%3D907026730415&authuser=1)

[A build with it turned on](https://console.cloud.google.com/cloud-build/builds/b4812613-fd74-46eb-9699-de0f7789a7a1?project=solo-public&angularJsUrl=%2Fcloud-build%2Fbuilds%2Fb4812613-fd74-46eb-9699-de0f7789a7a1%3Fproject%3D907026730415&authuser=1)

The build where it was turned off was actually the slower of the two, because the test task took longer. 

There are a few things that are easy that would speed up the build here, which you can see how we do in github.com/solo-io/gloo:
* Cache dep
* In parallel to dep, run a task that downloads the large containers (go-make, e2e-ginkgo)

ilackarms: i wasn't aware they were running in parallel. i'll reenable it

PR #169, opened 2019-02-13
start api refactor

ilackarms:
this pr does the following:

- start with a clean skeleton for the api
- implement a stub for the new `RoutingRule`, `Mesh` and `MeshGroup` resources
- implement a go "script" for vendoring/importing external protos from github for use with solo-kit
- use this script to vendor in istio networking protos


api/external/import_protos.go
```diff
@@ -0,0 +1,151 @@
...
+		modifiedProto = injectSoloKit(modifiedProto, skt.messageName, skt.shortName, skt.pluralName)
+	}
+	_, err = fmt.Fprint(out, modifiedProto)
+	return err
+}
+
+var goPackageStatementRegex = regexp.MustCompile(`option go_package.*=.*;`)
+
+func replaceGoPackage(in, importPath string) string {
+	in = strings.Replace(in, `import "gogoproto/gogo.proto";`, "", -1)
```
yuval-k: it seems that you are removing this import here, but adding it back 2 lines below?

ilackarms: in case it's already there, dumb way of preventing redeclare

api/v1/solo-kit.json
```diff
@@ -0,0 +1,38 @@
+{
+  "title": "SupaGloo",
```
rickducott: Supergloo

api/external/import_protos.go
```diff
@@ -0,0 +1,151 @@
...
+	"regexp"
+	"strings"
+
+	"github.com/solo-io/go-utils/errors"
+	"github.com/solo-io/solo-kit/pkg/utils/log"
+)
+
+// all you should need to do is append to this!
+var protosToImport = []importedProto{
+	{
```
rickducott: I think we should reconsider this approach. We are going to have a general solution for vendoring in protos soon, because we need that in order to be able to do solo kit generation without hacky gopath manipulation. We blocked documenting deving against gloo until that exists, so I model it as relatively high priority and expect we'll solve it within the next few weeks. 

If we already had a way to vendor in protos, then all this is doing is reading them in and rewriting them with additional metadata so we can do solo kit generation without requiring any changes to solo kit. The only changes we make to the protos seems to be: 1) defining the solo kit message name, short name, and plural; and 2) adding status and metadata fields. Seems like we could define (1) in a different file, and update solo kit to be able to handle external protos without explicitly requiring (2). That way, we can vendor in proto files and leave them untouched, and we just have to maintain a very small file providing the solo kit options.

PR #170, opened 2019-02-17
V2 translator

ilackarms:
this PR implements the base of the translator for istio. the top-level logic for constructing a set of istio config from supergloo api objects is implemented here.

separation between meshes, encrpytion translation, and individual plugins have not yet been implemented


api/v1/mesh.proto
```diff
@@ -24,7 +24,11 @@ message Mesh {
     // Metadata contains the object metadata for this resource
     core.solo.io.Metadata metadata = 101 [(gogoproto.nullable) = false];
 
+    message Istio {string b = 1;}
 
+    oneof mesh_type {
+        Istio istio = 1;
+    }
```
ilackarms: this piece isn't in use yet; will become the spot where mesh-specific config gets loaded

api/v1/routing.proto
```diff
@@ -25,4 +34,113 @@ message RoutingRule {
...
+        // https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS
+        // for further details about cross origin resource sharing.
+        istio.networking.v1alpha3.CorsPolicy cors_policy = 10;
+
+        // Mirror HTTP traffic to a another destination. Traffic will still be sent
+        // to its original destination as normal.
+        gloo.solo.io.Destination mirror = 9;
+
+        // manipulate request and response headers
+        HeaderManipulation header_manipulaition = 12;
```
mitchdraft: spelling here: header_manipulation

pkg/translator/istio/translator_encryption.go
```diff
@@ -0,0 +1,88 @@
...
+
+func initDestinationRule(writeNamespace, host string, labelSets []map[string]string) *v1alpha3.DestinationRule {
+
+	var subsets []*v1alpha3.Subset
+	for _, labels := range labelSets {
+		if len(labels) == 0 {
+			continue
+		}
+		subsets = append(subsets, &v1alpha3.Subset{
+			Name:   subsetName(labels),
```
mitchdraft: I worry about name collisions whenever we sanitize strings, could you add a check for uniqueness post-sanitization?

test/inputs/input_routing_rules.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"time"
+
+	"github.com/gogo/protobuf/types"
+	"github.com/solo-io/solo-kit/pkg/api/v1/resources/core"
+	v1 "github.com/solo-io/supergloo/pkg/api/v1"
+)
+
+func BookInfoRoutingRules(namespace string) v1.RoutingRuleList {
```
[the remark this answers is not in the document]
ilackarms: i will add more rules to this and make the function take more parameters so it's useful across multiple tests. right now it's just the minimum required for the one test that's using it

pkg/translator/istio/translator_encryption.go
```diff
@@ -0,0 +1,88 @@
...
+	keys, values := stringutils.KeysAndValues(labels)
+	name := ""
+	for i := range keys {
+		name += keys[i] + "-" + values[i] + "-"
+	}
+	name = strings.TrimSuffix(name, "-")
+	return sanitizeName(name)
+}
+
+func sanitizeName(name string) string {
```
yuval-k: does sanitize name makes sure its a valid dns name?
if so, do you need to worry about length?

pkg/translator/istio/destination_rules.go
```diff
@@ -0,0 +1,71 @@
...
+// creates a destination rule for every host
+// then a subset for every unique set of labels therein
+func destinationRulesFromUpstreams(writeNamespace string, upstreams gloov1.UpstreamList) (v1alpha3.DestinationRuleList, error) {
+	var destinationRules v1alpha3.DestinationRuleList
+	labelsByHost := make(map[string][]map[string]string)
+	for _, us := range upstreams {
+		labels := utils.GetLabelsForUpstream(us)
+		host, err := utils.GetHostForUpstream(us)
+		if err != nil {
+			return nil, errors.Wrapf(err, "getting host for upstream")
```
yuval-k: are these upstreams guaranteed to have a host?

pkg/translator/istio/translator.go
```diff
@@ -0,0 +1,162 @@
...
+	return meshConfig, resourceErrs, nil
+}
+
+func labelsAndPortsByHost(upstreams gloov1.UpstreamList) (map[string][]labelsPortTuple, error) {
+	labelsByHost := make(map[string][]labelsPortTuple)
+	for _, us := range upstreams {
+		labels := utils.GetLabelsForUpstream(us)
+		host, err := utils.GetHostForUpstream(us)
+		if err != nil {
+			return nil, errors.Wrapf(err, "getting host for upstream")
```
yuval-k: are these upstreams guaranteed to have a host?

pkg/translator/istio/translator.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	meshConfig.Sort()
+
+	return meshConfig, resourceErrs, nil
+}
+
+func labelsAndPortsByHost(upstreams gloov1.UpstreamList) (map[string][]labelsPortTuple, error) {
+	labelsByHost := make(map[string][]labelsPortTuple)
+	for _, us := range upstreams {
+		labels := utils.GetLabelsForUpstream(us)
```
yuval-k: are these upstreams created by discovery?
if so, do we need to take to account that upstreams are created with different labels? or is that the assumption here?

api/v1/routing.proto
```diff
@@ -25,4 +34,113 @@ message RoutingRule {
 
     // target where we apply this rule. this can be a mesh group or an individual mesh
     core.solo.io.ResourceRef target_mesh = 1;
+
+    // requests originating from these pods will have the rule applied
+    // leave empty to have all pods in the mesh apply these rules
+    PodSelector source_selector = 2;
```
yuval-k: is a pod selector a good fit for supergloo abstractions? 
what are we planning to do for vm based meshes like consul connect?

pkg/translator/istio/translator.go
ilackarms: kube and static will

there's more work to support static upstreams with istio (need to create a service entry for them)

it's possible to log the error here rather than returning on it (basically an error indicates it's an unsupported upstream type). i chose to return since it's something i think we should implement; i.e. if a user gives us an upstream that causes this func to error, we should extend support for that type

pkg/translator/istio/translator.go
ilackarms: that is the assumption

api/v1/routing.proto
ilackarms: we can revert to `UpstreamSelector` but at least in the case of kube this makes the workflow easier for users.

iiuc we're only focusing on kubernetes right now (even in the case of connect).

pkg/translator/utils/inspect_upstream.go
```diff
@@ -26,6 +35,7 @@ func GetHostForUpstream(us *gloov1.Upstream) (string, error) {
 	}
 	return hosts[0], nil
```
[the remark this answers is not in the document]
ilackarms: the first host (for the kube upstream type) includes the namespace, which is the one we want

pkg/translator/utils/inspect_upstream.go
```diff
@@ -16,6 +16,15 @@ func GetLabelsForUpstream(us *gloov1.Upstream) map[string]string {
 	return us.Metadata.Labels
 }
 
+func GetNamespaceForUpstream(us *gloov1.Upstream) string {
+	switch specType := us.UpstreamSpec.UpstreamType.(type) {
+	case *gloov1.UpstreamSpec_Kube:
+		return specType.Kube.ServiceNamespace
```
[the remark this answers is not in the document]
ilackarms: selecting pods by their namespace, so we want the service namespace

PR #174, opened 2019-02-21
v2 istio install

ilackarms:
adds a crd for performing installs
implements a function which performs installs from the crd
adds a functionality to perform updates to existing installs
- store the applied manifest as a gzipped string inside the crd
- use the previously applied configuration to perform the diff helm requires


ilackarms:
this pr is deceptively large: everything in `utils/helm` was moved from `pkg2` to `pkg`

PR #175, opened 2019-02-22
update solo kit, update apis, regenerate

ilackarms:
update apis


api/external/istio/authorization/v1alpha1/solo-kit.json
```diff
@@ -0,0 +1,10 @@
+{
+    "title": "",
+    "description": "",
+    "name": "istio.authentication.v1alpha1",
+    "version": "v1alpha1",
+    "docs_dir": "",
```
rickducott: I think this should be "docs" consistent with other repos

generate.go
```diff
@@ -2,14 +2,18 @@ package main
...
 
 //go:generate go run generate.go
 
 func main() {
 	log.Printf("starting generate")
-	if err := cmd.Run(".", true, true, []string{"../gloo"}, []string{"pkg2", "api2"}); err != nil {
+	docsOpts := &cmd.DocsOptions{
+		Output: options.Hugo,
+	}
+	if err := cmd.Run(".", true, docsOpts, []string{"../gloo"}, []string{"pkg2", "api2"}); err != nil {
```
rickducott: I don't see the generated docs in this PR

PR #176, opened 2019-02-22
add main func, update dockerfile

ilackarms:
adds a main, base setup.Main, along with a test

### DIFF
--- a/pkg/setup/setup_test.go
+++ b/pkg/setup/setup_test.go
@@ -16,18 +16,22 @@ import (
 var _ = Describe("Setup", func() {
 	var (
 		namespace string
+		ctx       context.Context
+		cancel    func()
 	)
 	BeforeEach(func() {
 		namespace = "a" + testutils.RandString(6)
 		err := setup.SetupKubeForTest(namespace)
 		Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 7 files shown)


pkg/setup/setup_test.go
```diff
@@ -0,0 +1,64 @@
...
+		Expect(err).NotTo(HaveOccurred())
+		errHandler := func(err error) {
+			Expect(err).NotTo(HaveOccurred())
+		}
+
+		mockSyncer := &mockInstallSyncer{}
+
+		go func() {
+			defer GinkgoRecover()
+			err = runInstallEventLoop(context.TODO(), errHandler, cs, v1.InstallSyncers{mockSyncer})
```
rickducott: I think this will run in the background until the process exits? Can we avoid that

pkg/setup/setup_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+		mockSyncer := &mockInstallSyncer{}
+
+		go func() {
+			defer GinkgoRecover()
+			err = runInstallEventLoop(ctx, errHandler, cs, v1.InstallSyncers{mockSyncer})
+			Expect(err).NotTo(HaveOccurred())
+		}()
+
+		// create an
```
rickducott: nit: fix or remove comment

PR #177, opened 2019-02-22
add cli with install

ilackarms:
adds a base for a cli with a command for install based on istio

includes a unit test that covers a range of inputs and versions


ilackarms:
also moves current `pkg` to `pkg2`

PR #178, opened 2019-02-25
add istio install syncer

ilackarms:
a bunch of files were moved from `pkg2` which were approved in a previous PR: specifically the Helm Install utils and the Istio install logic which wraps around them

this pr refactors those functions and integrates them into an `istioInstallSyncer` which implements a `v1.InstallSyncer`

the syncer simply wraps the install functions and ensures that the set of installs present in the `snapshot` is valid. it also calls a reporter with validation errors


pkg/install/istio/syncer.go
```diff
@@ -0,0 +1,130 @@
...
+
+			logger.Infof("overwriting previous mesh %v", existingMesh.Metadata)
+			createdMesh.Metadata.ResourceVersion = existingMesh.Metadata.ResourceVersion
+
+		}
+
+		logger.Infof("writing installed mesh %v", createdMesh.Metadata)
+		if _, err := s.meshClient.Write(createdMesh,
+			clients.WriteOpts{Ctx: ctx, OverwriteExisting: true}); err != nil {
+
```
rickducott: What to do in this case?

pkg/install/istio/syncer_test.go
```diff
@@ -0,0 +1,107 @@
...
+	v1 "github.com/solo-io/supergloo/pkg/api/v1"
+	. "github.com/solo-io/supergloo/pkg/install/istio"
+	"github.com/solo-io/supergloo/test/inputs"
+)
+
+type mockIstioInstaller struct {
+	enabledInstalls, disabledInstalls v1.InstallList
+}
+
+func (i *mockIstioInstaller) EnsureIstioInstall(ctx context.Context, install *v1.Install) (*v1.Mesh, error) {
```
rickducott: Would be good to try with a mock installer that throws errors, not just one that succeeds.

pkg/install/istio/syncer.go
ilackarms: whoops! good catch. i'll add more unit testing here

PR #180, opened 2019-02-26
add uninstall command, improve install command

ilackarms:
changes the current install command:
will enable a disabled install if that install crd already exists and is disabled
will error if the install exists and is enabled

adds an uninstall command
which sets the `disabled` field of an install crd to true


cli/pkg/cmd/install/install_test.go
```diff
@@ -65,5 +66,37 @@ var _ = Describe("Install", func() {
...
+			namespace := "ns"
+			inst := inputs.IstioInstall(name, namespace, "any", "1.0.5", false)
+			ic := helpers.MustInstallClient()
+			_, err := ic.Write(inst, clients.WriteOpts{})
+			Expect(err).NotTo(HaveOccurred())
+
+			err = utils.Supergloo("install istio " +
+				fmt.Sprintf("--name=%v ", name) +
+				fmt.Sprintf("--namespace=%v ", namespace))
+			Expect(err).To(HaveOccurred())
```
rickducott: Should set expects on the error message to protect against bad user experience

rickducott:
Looks good, one comment

PR #184, opened 2019-02-27
Typo

jpetazzo:


rickducott:
Thanks @jpetazzo. We have actually migrated supergloo docs into the solo-io/solo-docs repository. I migrated this change over to there: https://github.com/solo-io/solo-docs/pull/88

PR #189, opened 2019-03-01
Unit test installer

ilackarms:
switch the tests in install/istio from integration tests (which are really just testing the underyling helm installer) to use a mock helm installer, instead ensuring the behavior specific to the istio installer is behaviorally tested

### DIFF
--- a/pkg/install/istio/install_istio_test.go
+++ b/pkg/install/istio/install_istio_test.go
@@ -2,122 +2,109 @@ package istio
 
 import (
 	"context"
-	"time"
 
 	. "github.com/onsi/ginkgo"
-	"github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
-	"github.com/solo-io/go-utils/kubeutils"
-	"github.com/solo-io/go-utils/testutils"
-	"github.com/solo-io/solo-kit/test/helpers"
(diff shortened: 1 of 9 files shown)


pkg/install/istio/install_istio_test.go
```diff
@@ -2,122 +2,109 @@ package istio
...
+			Expect(err).NotTo(HaveOccurred())
+			Expect(actualManifests).NotTo(BeEmpty())
+
+			t.opts.previousInstall = actualManifests
+
+			t.opts.Observability.EnableJaeger = false
+			_, err = installer.installIstio(context.TODO(), t.opts)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(updateManifests).NotTo(BeEmpty())
+			Expect(manifestsHaveKey(updateManifests, "istio/charts/tracing/templates/service-jaeger.yaml")).To(BeFalse())
```
rickducott: I expected this test to check for an update manifest, not to check for the absence of an update manifest. It's not clear to me how this checks what the "It" suggests -- than an update is actually called

ilackarms: this check is asserting that setting `EnableJaeger` to be false produced manifests not containing jaeger yamls

ilackarms: `updateManifests` are also set when the mock `update` function is called

PR #196, opened 2019-03-03
traffic shifting plugin for istio

ilackarms:
depends on #194


pkg/translator/istio/plugins/istio_http.go
```diff
@@ -0,0 +1,91 @@
...
+				Host:   host,
+				Subset: utils.SubsetName(labels),
+				Port:   port,
+			},
+			Weight: weight,
+		})
+	}
+	// adjust weight in case rounding error occurred
+	if weightNeeded := 100 - totalIstioWeights; weightNeeded != 0 {
+		shiftedDestinations[0].Weight += weightNeeded
```
yuval-k: i noticed that this one gets the rounded error weight - but what if its original weight is 0?
not it will get traffic it shouldnt get.
I'm not sure if that's a valid case or can be ignored, but i noticed that `TrafficShiftingRuleSpec` generates an upstream with zero weight

pkg/translator/istio/plugins/istio_http.go
```diff
@@ -0,0 +1,91 @@
...
+		host, err := utils.GetHostForUpstream(upstream)
+		if err != nil {
+			return errors.Wrapf(err, "could not find host for upstream")
+		}
+		var port *v1alpha3.PortSelector
+		intPort, err := utils.GetPortForUpstream(upstream)
+		if err != nil {
+			return errors.Wrapf(err, "could not find port for upstream")
+		}
+		if intPort > 0 {
```
yuval-k: can the port be zero?
is that an error?

pkg/translator/istio/plugins/istio_http.go
```diff
@@ -0,0 +1,91 @@
...
+	return nil
+}
+
+func processTrafficShiftingRule(upstreams gloov1.UpstreamList, rule *v1.TrafficShifting, out *v1alpha3.HTTPRoute) error {
+	if rule.Destinations == nil || len(rule.Destinations.Destinations) == 0 {
+		return errors.Errorf("traffic shifting destinations cannot be missing or empty")
+	}
+	var shiftedDestinations []*v1alpha3.HTTPRouteDestination
+	var totalWeights uint32
+	var totalIstioWeights int32
```
yuval-k: this variable can move down near where it is used.

pkg/translator/istio/plugins/istio_http.go
```diff
@@ -0,0 +1,91 @@
...
+	return nil
+}
+
+func (*istioHttpPlugin) ProcessRoute(params Params, in v1.RoutingRuleSpec, out *v1alpha3.HTTPRoute) error {
+	switch rule := in.RuleType.(type) {
+	case *v1.RoutingRuleSpec_TrafficShifting:
+		if err := processTrafficShiftingRule(params.Upstreams, rule.TrafficShifting, out); err != nil {
+			return errors.Wrapf(err, "processing traffic shifting rule")
+		}
+	case *v1.RoutingRuleSpec_FaultInjection:
```
yuval-k: since these are not implemented, should we error?

pkg/translator/istio/plugins/istio_http.go
ilackarms: ah, i should fix that. thanks, 0 weight should be an error case

ilackarms: and i need to fix `TrafficShiftingRuleSpec` as well

pkg/translator/istio/plugins/istio_http.go
ilackarms: that seems fair for now

pkg/translator/istio/plugins/istio_http.go
ilackarms: you're correct on this, removing the check

pkg/translator/istio/plugins/istio_http_test.go
```diff
@@ -0,0 +1,135 @@
...
+					core.ResourceRef{Name: "default-reviews-v2-9080", Namespace: "default"},
+					core.ResourceRef{Name: "default-reviews-v3-9080", Namespace: "default"},
+				)
+				out := &v1alpha3.HTTPRoute{}
+
+				err := NewIstioHttpPlugin().ProcessRoute(
+					params, *in, out)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(out.Route).To(HaveLen(3))
```
rickducott: Construct an expected object and then have an `Expect(out).isEquivalentTo(expectedOut)

ilackarms:
/kick
there appears to be a flake with `pkg/install/utils/helm/helm_chart_installer_test.go:43`

i can look into this in a separate pr

pkg/translator/istio/plugins/istio_http_test.go
```diff
@@ -0,0 +1,159 @@
...
+					core.ResourceRef{Name: "default-details-v1-9080", Namespace: "default"},
+				)
+				out := &v1alpha3.HTTPRoute{}
+
+				err := NewIstioHttpPlugin().ProcessRoute(
+					params, *in, out)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(out.Route).To(HaveLen(4))
+				Expect(out.Route[0].Destination.Host).To(Equal("reviews.default.svc.cluster.local"))
```
rickducott: Construct the expected object and do the comparison against that

PR #202, opened 2019-03-05
Fix cli install

ilackarms:
fixes installation via the cli, adds install.sh script

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,9 +114,6 @@ SOURCES=$(shell find . -name "*.go" | grep -v test | grep -v mock)
 install-cli:
 	cd cli/cmd && go build -ldflags=$(LDFLAGS) -o $(GOPATH)/bin/supergloo
 
-.PHONY: supergloo-cli
-supergloo-cli: $(OUTPUT_DIR)/supergloo-cli-linux-amd64
-
 $(OUTPUT_DIR)/supergloo-cli-linux-amd64: $(SOURCES)
 	CGO_ENABLED=0 GOARCH=amd64 GOOS=linux go build -ldflags=$(LDFLAGS) -o $@ cli/cmd/main.go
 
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -0,0 +1,55 @@
+#! /bin/sh
+
+set -eu
+
+
+SUPERGLOO_VERSION=$(curl -H"Accept: application/vnd.github.v3+json" https://api.github.com/repos/solo-io/supergloo/releases/latest| python -c "import sys, json; print(json.load(sys.stdin)['tag_name'])" )
```
rickducott: This will fail if the latest release is a prerelease

In the gloo script we handle this

install.sh
```diff
@@ -0,0 +1,55 @@
...
+(
+  cd "$tmp"
+
+  echo "Downloading from ${url}..."
+
+  SHA=$(curl -sL "${url}.sha256" | awk '{ print $1 }')
+  curl -LO "${url}"
+  echo ""
+  echo "Download complete!, validating checksum..."
+  checksum=$(openssl dgst -sha256 "${filename}" | awk '{ print $2 }')
```
rickducott: i've been comparing to `shasum -a 256`, hopefully the same thing?

Makefile
```diff
@@ -114,6 +114,9 @@ SOURCES=$(shell find . -name "*.go" | grep -v test | grep -v mock)
 install-cli:
 	cd cli/cmd && go build -ldflags=$(LDFLAGS) -o $(GOPATH)/bin/supergloo
 
+.PHONY: supergloo-cli
+supergloo-cli: $(OUTPUT_DIR)/supergloo-cli-linux-amd64
```
rickducott: This shouldn't be hardcoded to linux, that won't work for most of us. Instead make this an alias for the correct GOOs. Here's what we do in gloo:

```
$(OUTPUT_DIR)/glooctl: $(SOURCES)
	go build -ldflags=$(LDFLAGS) -gcflags=$(GCFLAGS) -o $@ $(CLI_DIR)/cmd/main.go
```

install.sh
ilackarms: it's copy-pasted from gloo, so, yes?

PR #205, opened 2019-03-06
add --update flag for cli install

ilackarms:
to allow installs to be updated from the command line


cli/pkg/flagutils/install.go
```diff
@@ -8,7 +8,16 @@ import (
 	"github.com/spf13/pflag"
 )
 
-func AddIstioInstallFlags(set *pflag.FlagSet, in *options.InputInstall) {
+func AddInstallFlags(set *pflag.FlagSet, in *options.Install) {
+
+	set.BoolVar(&in.Update,
```
rickducott: Should this be update or upgrade? Looking at brew / apt-get, it seems like upgrade is used to refer to installing a newer version of a package, update is used to install a newer version of the tool / directory.

cli/pkg/cmd/install/istio_test.go
```diff
@@ -99,5 +100,28 @@ var _ = Describe("Install", func() {
 			Expect(err).To(HaveOccurred())
 			Expect(err.Error()).To(ContainSubstring("already installed and enabled"))
 		})
+		It("should update existing enabled install if --update set to true", func() {
+			name := "input"
+			namespace := "ns"
+			inst := inputs.IstioInstall(name, namespace, "any", "1.0.5", false)
+			inst.InstalledManifest = "a previously insatlled manifest"
```
rickducott: installed

cli/pkg/cmd/install/istio_test.go
```diff
@@ -99,5 +100,28 @@ var _ = Describe("Install", func() {
...
+			inst.InstalledMesh = &core.ResourceRef{"installed", "mesh"}
+			ic := helpers.MustInstallClient()
+			_, err := ic.Write(inst, clients.WriteOpts{})
+			Expect(err).NotTo(HaveOccurred())
+
+			err = utils.Supergloo("install istio " +
+				fmt.Sprintf("--name=%v ", name) +
+				fmt.Sprintf("--namespace=%v ", namespace) +
```
rickducott: Confused about install namespace in here.

cli/pkg/cmd/install/istio_test.go
```diff
@@ -99,5 +100,28 @@ var _ = Describe("Install", func() {
...
+				"--mtls=true " +
+				"--update=true ")
+			Expect(err).NotTo(HaveOccurred())
+
+			updatedInstall, err := ic.Read(namespace, name, clients.ReadOpts{})
```
rickducott: This test would be better if you just verify that updatedInstall is exactly the same as inst, maybe minus some resource version. It's not clear to me why the test passes due to the install namespace thing, and as a result I don't think the checks here are strong enough

cli/pkg/flagutils/install.go
ilackarms: it's not necessarily newer. it's used to modify config values. one of those is version, but it can even be a downgrade. you're just updating the existing install object

cli/pkg/cmd/install/istio_test.go
ilackarms: it defaults to `istio-system`. i added a test for this

PR #218, opened 2019-03-11
create config syncer which calls translator

ilackarms:
implement the config syncer and event loop setup


pkg/config/istio/syncer_reconcilers.go
```diff
@@ -0,0 +1,163 @@
...
+
+	"github.com/solo-io/supergloo/pkg/translator/istio"
+
+	"github.com/solo-io/go-utils/contextutils"
+	"github.com/solo-io/go-utils/errors"
+	"github.com/solo-io/solo-kit/pkg/api/v1/clients"
+	"github.com/solo-io/solo-kit/pkg/api/v1/resources"
+	"github.com/solo-io/solo-kit/pkg/api/v1/resources/core"
+	policyv1alpha1 "github.com/solo-io/supergloo/pkg/api/external/istio/authorization/v1alpha1"
+	"github.com/solo-io/supergloo/pkg/api/external/istio/networking/v1alpha3"
```
marcogschmidt: Nit: use alias like the other 2 `external` imports (for consistency)

pkg/config/setup/setup.go
```diff
@@ -0,0 +1,121 @@
...
+
+	if enableIstio {
+		istioSyncer, err := createIstioConfigSyncer(ctx, cs)
+		if err != nil {
+			return nil, err
+		}
+		syncers = append(syncers, istioSyncer)
+	}
+
+	return syncers, nil
```
marcogschmidt: If `syncers` is `nil` (when `enableIstio` is false), this would cause a nil pointer dereference in the [event loop](https://github.com/solo-io/supergloo/blob/dcd288d15fcf58e5dd1543148d5c604082b3869e/pkg/api/v1/config_event_loop.sk.go#L80), or am I missing something?

pkg/config/setup/setup.go
```diff
@@ -0,0 +1,121 @@
...
+	policyv1alpha1 "github.com/solo-io/supergloo/pkg/api/external/istio/authorization/v1alpha1"
+	"github.com/solo-io/supergloo/pkg/api/external/istio/networking/v1alpha3"
+	rbacv1alpha1 "github.com/solo-io/supergloo/pkg/api/external/istio/rbac/v1alpha1"
+	v1 "github.com/solo-io/supergloo/pkg/api/v1"
+	"github.com/solo-io/supergloo/pkg/config/istio"
+	istiotranslator "github.com/solo-io/supergloo/pkg/translator/istio"
+	"github.com/solo-io/supergloo/pkg/translator/istio/plugins"
+)
+
+func RunConfigEventLoop(ctx context.Context, cs *clientset.Clientset, customErrHandler func(error), enableIstio bool) error {
```
marcogschmidt: FWICT this function is not used anywhere. Will this happen in another PR?

pkg/install/utils/helm/helm_chart_installer_test.go
```diff
@@ -39,6 +39,7 @@ var _ = Describe("HelmChartInstaller", func() {
 	AfterEach(func() {
 		testutils.TeardownKube(ns)
 		testutils2.TeardownIstio(kubeClient)
+		testutils2.WaitForIstioTeardown(ns)
```
marcogschmidt: Nit: better name for `testutils2`

pkg/config/setup/setup.go
ilackarms: yes

pkg/config/setup/setup.go
ilackarms: there will be no `nil pointer` as `ConfigSyncers.Sync` is implemented even if `ConfigSyncers` is empty/nil:
https://github.com/solo-io/supergloo/blob/master/pkg/api/v1/config_event_loop.sk.go#L24

marcogschmidt: You're right, my bad

PR #239, opened 2019-03-13
add target-mesh to cli input routing rule

ilackarms:
cli command for creating routing rule was missing a flag `--target-mesh`, this adds it


cli/pkg/cmd/initialize/init_test.go
```diff
@@ -1,19 +1,27 @@
...
 	. "github.com/onsi/gomega"
+	"github.com/solo-io/go-utils/errors"
 	"github.com/solo-io/supergloo/cli/test/utils"
+	superglootest "github.com/solo-io/supergloo/test/e2e/utils"
 	"github.com/solo-io/supergloo/test/testutils"
 	v12 "k8s.io/api/core/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 var _ = Describe("init", func() {
```
EItanya: Lock this suite

cli/pkg/cmd/initialize/initialize_suite_test.go
```diff
@@ -11,3 +13,19 @@ func TestInitialize(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "Initialize Suite")
 }
+
+var (
+	lock *testutils.TestClusterLocker
+	err  error
+)
+
+var _ = BeforeSuite(func() {
```
EItanya: In my latest commit I removed the TestClusterLocker from supergloo testutils and put it in go-utils clusterlocker
You should use that one instead as my commit deletes the version in testutils.

ilackarms:
/kick

tests still have a race issue; lock release does not wait for namespaces to be completely torn down. need to address this in a new PR

PR #240, opened 2019-03-13
E2e test lock

EItanya:


test/testutils/cluster_lock.go
```diff
@@ -0,0 +1,106 @@
...
+		if e != nil {
+			if IsLockIsUseError(e) {
+				return true
+			}
+		}
+		return false
+	}),
+}
+
+var lockInUseError = fmt.Errorf("lock is current in use")
```
ilackarms: nit: currently

test/testutils/cluster_lock_test.go
```diff
@@ -0,0 +1,48 @@
...
+	It("can handle concurrent requests", func() {
+		for idx := 0; idx < 5; idx++ {
+			idx := idx
+			go func() {
+				lock, err := testutils.NewTestClusterLocker(clientset, "default", strconv.Itoa(idx))
+				Expect(err).NotTo(HaveOccurred())
+				Expect(lock.AcquireLock()).NotTo(HaveOccurred())
+				Expect(lock.ReleaseLock()).NotTo(HaveOccurred())
+			}()
+		}
```
ilackarms: i'd like to a see a test ensuring that the lock is acquired by only one goroutine at a time and all the others wait for it.

maybe create some kind of shared resource with an `Expect` that it's only ever accessed by one goroutine at a time

PR #242, opened 2019-03-13
Quay

EItanya:


cloudbuild.yaml
```diff
@@ -90,8 +90,8 @@ steps:
   # 3) Publish helm chart
   - name: 'gcr.io/cloud-builders/docker'
     entrypoint: 'bash'
-    args: ['-c', 'docker login --username soloiobot --password $$DOCKER_HUB_PASSWORD']
-    secretEnv: ['DOCKER_HUB_PASSWORD']
+    args: ['-c', 'docker login quay.io --username "solo-io+solobot" --password QUAY_IO_PASSWORD']
```
rickducott: do we need $$?

EItanya: yup, double click accidentally deleted it

install/helm/supergloo/generate-values.go
```diff
@@ -17,11 +20,17 @@ const (
 	chartTemplate = "install/helm/supergloo/Chart-template.yaml"
 	chartOutput   = "install/helm/supergloo/Chart.yaml"
 
+	glooPkg = "github.com/solo-io/gloo"
```
rickducott: Use `versionutils.GlooPkg` instead

PR #255, opened 2019-03-14
ingress install api

EItanya:


api/v1/install.proto
```diff
@@ -32,56 +32,70 @@ message Install {
...
 
-        // install grafana with istio
-        bool install_grafana = 6;
+// Installation options for Istio
+message Istio {
 
-        // install prometheus with istio
-        bool install_prometheus = 7;
+    // which namespace to install to
+    string installation_namespace = 1;
```
ilackarms: we can probably extract `installation_namespace` to the outer proto

api/v1/install.proto
```diff
@@ -98,12 +112,52 @@ message Install {
...
+
+        // install a custom helm chart
+        // an associated ingress object
+        // will not be created by supergloo
+        // create the mesh object manually
+        // in a future release, supergloo
+        // will feature discovery
+        // which can discover mesh objects
+        // automatically
+        Custom custom = 2;
```
ilackarms: i think we should move towards removing `Custom` rather than duplicating it. there's no implementation for it planned for the MVP, i'm not a huge fan of placeholder APIs. might as well remove it as part of this PR

api/v1/install.proto
```diff
@@ -98,12 +112,52 @@ message Install {
...
+    // install this helm chart.
+    // format should be a URI accessible to supergloo
+    string custom_helm_chart = 9;
+
+    // optional. if specified,
+    // these values will be used
+    // when rendering the helm chart
+    string custom_helm_values = 10;
+
+}
```
ilackarms: let's just remove this `Custom` for now, it's not going to be implemented for a long time

api/v1/install.proto
```diff
@@ -98,12 +112,52 @@ message Install {
...
+
+message GlooIngress {
+
+    // which namespace to install to
+    string installation_namespace = 1;
+
+    // which version of the gloo helm chart to install
+    // ignored if using custom helm chart
+    string gloo_version = 2;
+
```
ilackarms: the install for the ingress should contain a (nullable) reference to a mesh called `ingress_mesh`. if `ingress_mesh` is set, it tells supergloo to do the piping to ensure that outbound tls is configured properly on the ingress

ilackarms:
overall i like the direction, basically my two thoughts are:
- provide a way to link an ingress to a mesh
- remove the `Custom` message

api/v1/install.proto
```diff
@@ -32,78 +32,92 @@ message Install {
...
+message IngressInstall {
+
+    oneof install_type {
+        GlooIngress gloo = 1;
+
+    }
+    
+    // reference to the Mesh crd which this ingress is pointing
+    // read-only.
```
ilackarms: the comment here doesn't make sense
should be something like "if specified, this ingress will share policy, routing, and mtls configuration with the target mesh. this is required to connect to mtls-secured services via the ingress"

PR #265, opened 2019-03-16
set root cert on mesh crd via cli

ilackarms:


ilackarms:
based on #261

cli/pkg/surveyutils/root_cert.go
```diff
@@ -0,0 +1,96 @@
...
+	return nil
+}
+
+func SurveyMesh(prompt string, ctx context.Context) (core.ResourceRef, error) {
+	meshes, err := helpers.MustMeshClient().List("", clients.ListOpts{Ctx: ctx})
+	if err != nil {
+		return core.ResourceRef{}, err
+	}
+
+	byKey := make(map[string]core.ResourceRef)
```
EItanya: Could this logic possibly be split out because it's the exact same between this and the next function

PR #272, opened 2019-03-18
ensure supergloo-system is completely torn down before releasing lock

### LINKED ISSUE #270: fix CI race with namespace teardown
while we have a lock that prevents cluster-scope tests from running concurrently, these tests don't check that teardown has finished before releasing their lock

we need to make sure that e2e tests wait for all cluster-scope resources to be cleared before releasing the lock

ilackarms:
fixes #270

### DIFF
--- a/cli/pkg/cmd/initialize/initialize_suite_test.go
+++ b/cli/pkg/cmd/initialize/initialize_suite_test.go
@@ -27,7 +27,7 @@ var _ = BeforeSuite(func() {
 })
 
 var _ = AfterSuite(func() {
+	defer lock.ReleaseLock()
 	testutils.TeardownSuperGloo(testutils.MustKubeClient())
 	testutils.WaitForNamespaceTeardown("supergloo-system")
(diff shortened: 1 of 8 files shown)


cli/pkg/cmd/initialize/initialize_suite_test.go
```diff
@@ -27,5 +27,7 @@ var _ = BeforeSuite(func() {
 })
 
 var _ = AfterSuite(func() {
+	testutils.TeardownSuperGloo(testutils.MustKubeClient())
+	testutils.WaitForNamespaceTeardown("supergloo-system")
```
rickducott: I think if either of these functions fail, then an expect will fail, meaning the After function will stop executing, and the lock won't be released. The idiom I have been using is:
```
defer lock.ReleaseLock()
```
At the beginning of the After function. If it fails... it's not like we can do anything about it 🤷‍♂️

ilackarms: good point

PR #274, opened 2019-03-19
istio secret deleter for cleaing out citadel secrets

ilackarms:
generated istio secrets should be deleted when the user switches to using a custom root cert: see https://istio.io/docs/tasks/security/plugin-ca-cert/#plugging-in-the-existing-certificate-and-key for reference


test/inputs/input_istio_mesh.go
```diff
@@ -6,14 +6,18 @@ import (
 )
 
 func IstioMesh(namespace string, secretRef *core.ResourceRef) *v1.Mesh {
+	return IstioMeshWithInstallNs(namespace, "istio-was-installed-herr", secretRef)
```
marcogschmidt:
```suggestion
	return IstioMeshWithInstallNs(namespace, "istio-was-installed-here", secretRef)
```

PR #280, opened 2019-03-20
Prometheus config resource

ilackarms:
creates a solo-kit resource from a prometheus config, stored as a kubernetes configmap. contains utility functions for converting between an untyped solo-kit resource (wrapping a proto.Struct) and a typed Prometheus config


pkg/api/custom/clients/prometheus/prometheus_test.go
```diff
@@ -0,0 +1,103 @@
...
+				testPrometheusSerializer(namespace, kube, client)
+			})
+		})
+	}
+})
+
+func testConvert(namespace string, kube kubernetes.Interface, client PrometheusConfigClient) {
+	name := "prometheus-config"
+	err := CreatePrometheusConfigmap(namespace, name, kube)
+	Expect(err).NotTo(HaveOccurred())
```
marcogschmidt: We should change all the `Expect(...)` functions in these helper function to `ExpectWithOffset(1, ...)` in order to correctly log the line where the assertion failed.

pkg/api/custom/clients/prometheus/prometheus_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+var _ = Describe("Prometheus Config Conversion", func() {
+	var (
+		namespace string
+	)
+	for _, test := range []typed.ResourceClientTester{
+		&typed.KubeConfigMapRcTester{},
```
marcogschmidt: Are we expecting to add more `ResourceClientTester`s here? If not, we can remove the loop.

PR #282, opened 2019-03-20
Cli create AWS secret

marcogschmidt:
Added `supergloo create secret aws` command to CLI. We need this in order for supergloo to be able to manage AWS App Mesh resources.

**NOTE**: check the first commit for the actual changes.


cli/pkg/options/options.go
```diff
@@ -81,6 +82,13 @@ type CreateTlsSecret struct {
 	CaCertFilename     string
 }
 
+type AwsSecret struct {
```
ilackarms: nit: CreateAwsSecret

marcogschmidt: Renamed

PR #283, opened 2019-03-20
added other namespace to delete

EItanya:


pkg/install/gloo/syncer.go
```diff
@@ -91,7 +91,7 @@ func (s *installSyncer) handleDisabledInstalls(ctx context.Context,
 				resourceErrs.AddError(in, errors.Wrapf(err, "uninstall failed"))
 			} else {
 				resourceErrs.Accept(in)
-				if err := s.meshClient.Delete(installedIngress.Namespace,
+				if err := s.ingressClient.Delete(installedIngress.Namespace,
```
ilackarms: i think we should add a test for this

PR #285, opened 2019-03-21
prometheus scrape config syncer

ilackarms:
adds 2 packages:
- `stats/prometheus`: includes a `PrometheusConfigUpdater` which ensures a set of prometheus scrape configs are present in a Prometheus configmap (read/write to storage)
- `stats/istio`: includes a utility function for rendering the scrape configs required for istio.

the plan is to integrate these two packages into an istio Prometheus syncer which will watch for mesh crds and update a `targetPrometheus` (which will be a new field on mesh proto) with the correct scrape configs for that mesh / install namespace


test/inputs/input_prometheus_configmap.go
```diff
@@ -1,5 +1,25 @@
...
+	v1 "github.com/solo-io/supergloo/pkg/api/external/prometheus/v1"
+	"gopkg.in/yaml.v2"
+)
+
+func InputPrometheusConfig(name, namespace string) *v1.PrometheusConfig {
+	return &v1.PrometheusConfig{
+		Metadata: core.Metadata{
+			Name:        name,
+			Namespace:   namespace,
+			Annotations: map[string]string{"resource_kind": reflect.TypeOf(&v1.PrometheusConfig{}).String()},
```
EItanya: `resource.kind`

PR #286, opened 2019-03-22
Gloo istio mtls

EItanya:


api/v1/install.proto
```diff
@@ -130,5 +130,8 @@ message GlooInstall {
     // ignored if using custom helm chart
     string gloo_version = 2;
 
+    // which mesh for gloo to target
+    repeated string meshes = 3;
```
ilackarms: why a string and not a resource ref?

api/v1/solo-kit.json
```diff
@@ -46,6 +46,10 @@
       {
         "name": "Mesh",
         "package": "supergloo.solo.io"
+      },
+      {
+        "name": "MeshIngress",
+        "package": "supergloo.solo.io"
```
ilackarms: was this missing before? i'm surprised to see it on this pr

pkg/config/setup/setup.go
```diff
@@ -87,7 +92,7 @@ func createIstioConfigSyncer(ctx context.Context, cs *clientset.Clientset) (v1.C
 }
 
 // start the istio config event loop
-func runConfigEventLoop(ctx context.Context, clientset *clientset.Clientset, errHandler func(err error), syncers v1.ConfigSyncer) error {
+func runConfigEventLoop(ctx context.Context, clientset *clientset.Clientset, errHandler func(err error), syncers v1.ConfigSyncers) error {
```
ilackarms: `ConfigSyncers` is a `ConfigSyncer`, no need for this change (i prefer keeping it more open-ended as an interface)

cli/pkg/cmd/install/gloo.go
```diff
@@ -56,6 +56,7 @@ func installGlooFromOpts(opts *options.Options) (*v1.Install, error) {
 				InstallType: &v1.MeshIngressInstall_Gloo{
 					Gloo: &v1.GlooInstall{
 						GlooVersion: opts.Install.GlooIngressInstall.GlooVersion,
+						Meshes:      []string{"istio"},
```
ilackarms: do we want this as a hard-coded config option? should there not be a flag+survey util to set this option?

pkg/install/gloo/installer.go
```diff
@@ -78,12 +87,35 @@ func (installer *defaultInstaller) EnsureGlooInstall(ctx context.Context, instal
 		return nil, errors.Wrapf(err, "converting installed manifests to gzipped string")
 	}
 
-	meshIngress := &v1.MeshIngress{
-		Metadata: core.Metadata{
-			Namespace: install.Metadata.Namespace,
-			Name:      install.Metadata.Name,
-		},
-		MeshIngressType: &v1.MeshIngress_Gloo{},
+	meshRefs := make([]*core.ResourceRef, 0)
```
ilackarms: the convention we've been going with so far to initialize empty slices is `var sliceName []sliceType`

`append` will still work on nil arrays: https://tour.golang.org/moretypes/15

pkg/install/gloo/installer.go
```diff
@@ -78,12 +87,35 @@ func (installer *defaultInstaller) EnsureGlooInstall(ctx context.Context, instal
...
+	for _, mesh := range meshes {
+		if contains(glooInstall.Gloo.Meshes, mesh.Metadata.Name) {
```
ilackarms: this can cause an issue with name collisions across namespaces. i think the `meshes` field should be a slice of mesh resource refs to be consistent with the rest of the API

cli/pkg/cmd/install/gloo.go
EItanya: Haven't gotten to the CLI part yet

api/v1/solo-kit.json
EItanya: Yes it was

api/v1/install.proto
EItanya: No good reason, I can switch it

api/v1/install.proto
```diff
@@ -130,5 +130,8 @@ message GlooInstall {
     // ignored if using custom helm chart
     string gloo_version = 2;
 
+    // reference to the Mesh(s) that this ingress is acting upon
```
ilackarms: i would refine this comment a bit further to clarify what this feature means: enable the ingress to route to services within these mTLS-enabled meshes

pkg/install/gloo/installer.go
```diff
@@ -64,6 +64,15 @@ func (installer *defaultInstaller) EnsureGlooInstall(ctx context.Context, instal
...
 	}
 
+	var meshIngress *v1.MeshIngress
+	if installIngress.Ingress.InstalledIngress != nil {
+		var err error
+		meshIngress, err = meshIngresses.Find(installIngress.Ingress.InstalledIngress.Strings())
+		if err != nil {
+			return nil, errors.Wrapf(err, "installed ingress not found")
+		}
+	}
```
ilackarms: nit: move `meshIngress` closer to where it is used

pkg/install/gloo/installer.go
```diff
@@ -78,12 +87,38 @@ func (installer *defaultInstaller) EnsureGlooInstall(ctx context.Context, instal
...
+			if glooMesh.Namespace == mesh.Metadata.Namespace &&
+				glooMesh.Name == mesh.Metadata.Name {
+				ref := mesh.Metadata.Ref()
+				meshRefs = append(meshRefs, &ref)
+			}
+		}
+	}
+
+	if meshIngress != nil {
+		meshIngress.Meshes = meshRefs
```
ilackarms: i don't understand the purpose of creating and passing `meshRefs`. why not simply copy `glooInstall.Gloo.Meshes` here?

pkg/install/gloo/installer.go
```diff
@@ -78,12 +87,38 @@ func (installer *defaultInstaller) EnsureGlooInstall(ctx context.Context, instal
...
+			Metadata: core.Metadata{
+				Namespace: install.Metadata.Namespace,
+				Name:      install.Metadata.Name,
+			},
+			MeshIngressType: &v1.MeshIngress_Gloo{
+				Gloo: &v1.GlooMeshIngress{
+					InstallationNamespace: install.Metadata.Namespace,
+				},
+			},
+			Meshes: meshRefs,
```
ilackarms: same as above (meshRefs)

pkg/install/gloo/installer.go
```diff
@@ -78,12 +87,38 @@ func (installer *defaultInstaller) EnsureGlooInstall(ctx context.Context, instal
...
-		Metadata: core.Metadata{
-			Namespace: install.Metadata.Namespace,
-			Name:      install.Metadata.Name,
-		},
-		MeshIngressType: &v1.MeshIngress_Gloo{},
+	var meshRefs []*core.ResourceRef
+	for _, mesh := range meshes {
+		for _, glooMesh := range glooInstall.Gloo.Meshes {
+			if glooMesh.Namespace == mesh.Metadata.Namespace &&
+				glooMesh.Name == mesh.Metadata.Name {
```
ilackarms: the convention here is to check
```
mesh, err := meshes.Find(ref)
if err != nil { // not found
...
```
rather than ranging over the slice

pkg/registration/gloo/syncer.go
```diff
@@ -0,0 +1,123 @@
...
+	"k8s.io/api/extensions/v1beta1"
+
+	"github.com/solo-io/go-utils/contextutils"
+	"github.com/solo-io/go-utils/errors"
+	"github.com/solo-io/supergloo/pkg/api/clientset"
+	v1 "github.com/solo-io/supergloo/pkg/api/v1"
+	kubev1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type glooConfigSyncer struct {
```
ilackarms: rename
maybe something like `glooMtlsSyncer`

pkg/registration/gloo/syncer.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (s *glooConfigSyncer) Sync(ctx context.Context, snap *v1.RegistrationSnapshot) error {
+	ctx = contextutils.WithLogger(ctx, fmt.Sprintf("gloo-config-sync-%v", snap.Hash()))
+	logger := contextutils.LoggerFrom(ctx)
+	logger.Infof("begin sync %v: %v", snap.Hash(), snap.Stringer())
+	defer logger.Infof("end sync %v", snap.Hash())
+	logger.Debugf("full snapshot: %v", snap)
+
+	glooMeshIngresses := make(v1.MeshIngressList, 0)
```
ilackarms: nit: `var glooMeshIngresses v1.MeshIngressList`

pkg/registration/gloo/syncer.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (s *glooConfigSyncer) handleGlooMeshIngressConfig(ingress *v1.MeshIngress, meshes v1.MeshList) error {
+
+	glooMeshIngress, isGloo := ingress.MeshIngressType.(*v1.MeshIngress_Gloo)
+	if !isGloo {
+		return errors.Errorf("only gloo mesh ingress currently supported")
+	}
+	targetMeshes := ingress.Meshes
+
```
ilackarms: there should be a log line here indicating we're about to modify a kube deployment

pkg/registration/gloo/syncer.go
```diff
@@ -0,0 +1,123 @@
...
+	deployment, err := s.cs.Kube.ExtensionsV1beta1().Deployments(glooMeshIngress.Gloo.InstallationNamespace).Get("gateway-proxy", kubev1.GetOptions{})
```
ilackarms: can we make `"gateway-proxy"` a configuration option rather than a hard-coded string?

pkg/registration/gloo/syncer.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "unable to find deployemt for gateway-proxy in %s", glooMeshIngress.Gloo.InstallationNamespace)
+	}
+
+	update, err := ShouldUpdateDeployment(deployment, targetMeshes, meshes)
+	if err != nil {
+		return err
+	}
+
+	if update {
```
ilackarms: nit: 
```
if !update {
 return nil
}
// un-nested logic
```

pkg/registration/gloo/syncer.go
```diff
@@ -0,0 +1,123 @@
...
+		_, err := s.cs.Kube.ExtensionsV1beta1().Deployments(glooMeshIngress.Gloo.InstallationNamespace).Update(deployment)
+		if err != nil {
+			return errors.Wrapf(err, "unable to rewrite deployment after update")
+		}
+	}
+
+	return nil
+}
+
+func ShouldUpdateDeployment(deployment *v1beta1.Deployment, targetMeshes []*core.ResourceRef, meshes v1.MeshList) (bool, error) {
```
ilackarms: should this function be private?

pkg/install/gloo/installer.go
EItanya: Just wanted to double check that they exist I guess

pkg/registration/gloo/volume_info.go
```diff
@@ -0,0 +1,143 @@
...
+	optional          = true
+)
+
+type VolumeList []corev1.Volume
+
+func (s VolumeList) Remove(i int) VolumeList {
+	tmp := make(VolumeList, len(s))
+	copy(tmp, s)
+	tmp[i] = tmp[len(tmp)-1]
+	return tmp[:len(tmp)-1]
```
ilackarms: consider preserve order here? e.g. with 
```
return apppend(tmp[:i], tmp[i+1:]...)
```

pkg/registration/gloo/volume_info.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+type VolumeMountList []corev1.VolumeMount
+
+func (s VolumeMountList) Remove(i int) VolumeMountList {
+	tmp := make(VolumeMountList, len(s))
+	copy(tmp, s)
+	tmp[i] = tmp[len(tmp)-1]
+	return tmp[:len(tmp)-1]
```
ilackarms: same as above

pkg/registration/gloo/volume_info.go
```diff
@@ -0,0 +1,143 @@
...
+import (
+	"path/filepath"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/solo-io/solo-kit/pkg/api/v1/resources/core"
+	v1 "github.com/solo-io/supergloo/pkg/api/v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
```
ilackarms: can we make all the functions and types in this file private?

pkg/registration/gloo/syncer.go
```diff
@@ -0,0 +1,123 @@
...
+	var volumes VolumeList = deployment.Spec.Template.Spec.Volumes
+	gatewayProxyContainer := deployment.Spec.Template.Spec.Containers[0]
+	var mounts VolumeMountList = gatewayProxyContainer.VolumeMounts
+
+	newDeploymentVolumes, err := ResourcesToDeploymentInfo(targetMeshes, meshes)
```
ilackarms: suggest renaming `newDeploymentVolumes` and `ResourcesToDeploymentInfo` to

```
secretVolumeInfo, err := makeSecretVolumesForMeshes(targetMeshes, meshes)
```

pkg/install/gloo/installer.go
ilackarms: if you want to validate, i would suggest using `if _, err := meshes.Find(ref); err != nil {...`

pkg/registration/gloo/syncer.go
EItanya: I agree with this in theory, but we don't even allow modification of this value from our helm chart so it doesn't seem worth it

ilackarms: i was thinking that at the very least the user can indicate `ingress-proxy` or `clusteringress-proxy`

pkg/registration/gloo/syncer.go
```diff
@@ -0,0 +1,128 @@
...
+
+	return nil
+}
+
+func shouldUpdateDeployment(deployment *v1beta1.Deployment, targetMeshes []*core.ResourceRef, meshes v1.MeshList) (bool, error) {
+	var volumes VolumeList = deployment.Spec.Template.Spec.Volumes
+	gatewayProxyContainer := deployment.Spec.Template.Spec.Containers[0]
+	var mounts VolumeMountList = gatewayProxyContainer.VolumeMounts
+
+	newDeploymentVolumes, err := resourcesToDeploymentInfo(targetMeshes, meshes)
```
ilackarms: suggestion: rename function to something like `secretVolumesForMesh`

both `resources` and `deploymentInfo` sound very generic to me

PR #290, opened 2019-03-24
adding and removing jobs from a prometheus config

ilackarms:
includes api for setting the target prometheus instance on a mesh

this is mostly functions and refactoring to make it easier to work with the prometheus config object, convert to/from the solo kit resource (which is a string wrapper)


changelog/v0.3.5/add-scrape-configs.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
```
EItanya: nit: this is user facing because it's an addition to the API

PR #295, opened 2019-03-25
Cli set mesh stats

ilackarms:
set prometheus stats propagation for a mesh via the cli


cli/pkg/cmd/set/stats.go
```diff
@@ -0,0 +1,78 @@
...
+		},
+	}
+
+	flagutils.AddSetStatsFlags(cmd.PersistentFlags(), &opts.SetStats)
+
+	return cmd
+}
+
+func setStats(opts *options.Options) error {
+	meshRef := opts.SetStats.TargetMesh
```
EItanya: possibly split into separate validation function for separation of concerns.
Can even possibly call from PreRunE

PR #298, opened 2019-03-26
Appmesh injector webhook

marcogschmidt:
Added a `MutationAdmissionWebhook` server that injects to-be-created-pods with AWS App Mesh sidecar proxies. Check out the README under `cmd/admissionwebhook` for a detailed description of the webhook logic.

This PR includes only unit tests. There will be a follow up with e2e tests that deploy the webhook to EKS. These will most likely be included with the App Mesh registration functionality.

**NOTE**: Added the WIP label as I am cleaning up the tests a bit.


api/v1/mesh.proto
```diff
@@ -41,12 +43,42 @@ message Mesh {
...
 }
 
+// Mesh object representing AWS App Mesh
+message AwsAppMesh {
+
+    // Reference to the secret that holds the AWS credentials that will be used to access the AWS App Mesh service.
+    core.solo.io.ResourceRef aws_secret = 1;
+
+    // The AWS region the AWS App Mesh control plane resources (Virtual Nodes, Virtual Routers, etc.) will be created in.
+    string region = 2;
```
ilackarms: suggestion: make this an `enum` instead, with supported options available. so we don't have to do any validation ourselves

ilackarms: though i'm not sure how proto supports using strings as an input, so maybe its not worth it

pkg/webhook/plugins/appmesh_plugin.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	for _, mesh := range meshes {
+		appMesh, isAppMesh := mesh.MeshType.(*v1.Mesh_AwsAppMesh)
+		if !isAppMesh {
+			continue
+		}
+		if appMesh.AwsAppMesh == nil {
+			contextutils.LoggerFrom(ctx).Warnf("unexpected nil value for AwsAppMesh in mesh %s.%s", mesh.Metadata.Namespace, mesh.Metadata.Name)
+			continue
```
ilackarms:
```
appMesh := mesh.GetAppMesh()
if appMesh == nil // oneof was of a different type
```

marcogschmidt: Cool, didn't know about that. Updated it.

PR #302, opened 2019-03-26
gloo mtls e2e

EItanya:


cli/pkg/cmd/install/gloo.go
```diff
@@ -95,6 +95,7 @@ func validateGlooInstall(opts *options.Options) error {
 			if _, err := meshes.Find(v.Namespace, v.Name); err != nil {
 				return errors.Wrapf(err, "mesh resource %s.%s is not a valid mesh", v.Namespace, v.Name)
```
ilackarms: replace `%s.%s` with `%s` and `v.Namespace, v.Name` with `v.Key()`

also suggest renaming `v` to `meshRef`

install/helm/supergloo/generate/generate.go
```diff
@@ -0,0 +1,137 @@
...
+var (
+	osGlooVersion string
+	rootPrefix    string
+)
+
+func Run(version, pullPolicy, prefix string) error {
+	rootPrefix = prefix
+	if glooVersion, err := getOsGlooVersion(); err != nil {
+		return err
+	} else {
```
ilackarms: no need for `else` here

PR #317, opened 2019-03-31
gloo proxy url

EItanya:


api/v1/ingress.proto
```diff
@@ -31,13 +31,13 @@ message MeshIngress {
     oneof mesh_ingress_type {
         GlooMeshIngress gloo = 1;
     }
+    // where Gloo has been installed
```
ilackarms: where the ingress has been installed

PR #322, opened 2019-04-02
Check containerPorts on pods that need to be injected with sidecar

marcogschmidt:


pkg/webhook/plugins/appmesh_plugin.go
```diff
@@ -172,9 +172,21 @@ func getTemplateData(pod *corev1.Pod, mesh *v1.Mesh) (templateData, error) {
 		return templateData{}, errors.Errorf("pod is missing required virtual node label %v", vnLabel)
 	}
 
+	if len(pod.Spec.Containers) != 1 {
```
EItanya: support multiple containers

PR #325, opened 2019-04-02
New install utils

ilackarms:


pkg/install/utils/kubeinstall/kube_installer.go
```diff
@@ -0,0 +1,435 @@
...
+		group := groupedResourcesToDelete[i-1]
+		g := errgroup.Group{}
+		for _, res := range group.Resources {
+			res := res
+			g.Go(func() error {
+				if err := r.preDelete(res); err != nil {
+					return err
+				}
+				resKey := fmt.Sprintf("%v %v.%v", res.GroupVersionKind().Kind, res.GetNamespace(), res.GetName())
+				contextutils.LoggerFrom(ctx).Infof("deleting resource %v", resKey)
```
EItanya: nit: define this logger up top and log more information.
There is a lot going on here that would be useful to expose, especially at a debug level

pkg/install/utils/kuberesource/get_cluster_resources_test.go
```diff
@@ -0,0 +1,64 @@
...
+	. "github.com/onsi/gomega"
+	"github.com/solo-io/go-utils/kubeutils"
+	"github.com/solo-io/go-utils/testutils"
+	"github.com/solo-io/supergloo/test/e2e/utils"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var _ = Describe("GetClusterResources", func() {
```
EItanya: Can we figure our a way to run this in parallel so it is non-blocking?

pkg/install/utils/kuberesource/sort.go
```diff
@@ -0,0 +1,113 @@
...
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/helm/pkg/tiller"
+)
+
+// set the install order based on
+// Helm's list
+var installOrder = func() map[string]int {
+	installOrderMapping := make(map[string]int)
+	for installValue, kind := range tiller.InstallOrder {
```
EItanya: Why use tiller.InstallOrder if you have a local InstallOrder as well below which seems identical?

ilackarms: something i forgot to clean up

pkg/install/utils/kuberesource/get_cluster_resources_test.go
ilackarms: seems like we should address that in an overall parallelization strategy
but yes, this test should not affect any other test

ilackarms:
/kick
oom

PR #330, opened 2019-04-03
Cli register appmesh

marcogschmidt:
BOT NOTES: 
resolves https://github.com/solo-io/supergloo/issues/214


cli/pkg/cmd/register/app_mesh.go
```diff
@@ -0,0 +1,131 @@
...
+						Namespace: opts.RegisterAppMesh.ConfigMap.Namespace,
+						Name:      opts.RegisterAppMesh.ConfigMap.Name,
+					}
+				}
+			}
+
+			_, err := clients.MustMeshClient().Write(mesh, skclients.WriteOpts{})
+			if err != nil {
+				return errors.Wrapf(err, "failed to create mesh %v", mesh.Metadata.String())
+			}
```
EItanya: The convention that we're following in other commands which create resources is to print the resource once it has been completed check install for example

cli/pkg/cmd/register/app_mesh_test.go
```diff
@@ -0,0 +1,226 @@
...
+	})
+
+	Describe("flag validation works as expected", func() {
+
+		It("fails if no name was provided", func() {
+			err := utils.Supergloo(fmt.Sprintf(
+				"register appmesh --region %s --secret %s --auto-inject %s",
+				"us-east-1",
+				"DEFAULT.SECRET",
+				"false",
```
EItanya: Did you figure out the weird behavior with bool flags

cli/pkg/options/values.go
```diff
@@ -29,7 +29,7 @@ func (v *ResourceRefsValue) String() string {
 
 func (v *ResourceRefsValue) Set(s string) error {
 	split := strings.SplitN(s, ".", 2)
-	if len(split) != 2 {
+	if len(split) != 2 || strings.Trim(split[0], " ") == "" || strings.Trim(split[1], " ") == "" {
```
EItanya: good call

cli/pkg/surveyutils/register_appmesh.go
```diff
@@ -0,0 +1,191 @@
...
+	default:
+		return errors.Errorf("%v is not a known selector type", injectionSelectorType)
+	}
+
+	return nil
+}
+
+func surveyConfigMap(opts *options.Options) error {
+	var useCustomConfigMap string
+	if err := cliutil.ChooseFromList(
```
EItanya: There's boilerplate code for selecting resources, the function is called `surveyResources` in that package

cli/pkg/cmd/register/app_mesh_test.go
marcogschmidt: Yeah, it does not work with the other commands either, it's a bug. I filed [this issue](https://github.com/solo-io/supergloo/issues/323) to track it.

cli/pkg/cmd/register/app_mesh.go
marcogschmidt: Right, I forgot. Thank for catching that!

cli/pkg/options/values.go
marcogschmidt: 👍

cli/pkg/cmd/register/app_mesh_test.go
EItanya: okay cool

cli/pkg/surveyutils/register_appmesh.go
marcogschmidt: Unfortunately I can't use it here as it accepts only our custom resources.

PR #338, opened 2019-04-05
Refactor installer

ilackarms:
Installers are now stateless using new install utils


pkg/install/setup/setup.go
```diff
@@ -29,7 +31,22 @@ func RunInstallEventLoop(ctx context.Context, cs *clientset.Clientset, customErr
 		}
 	}
 
-	installSyncers := createInstallSyncers(cs, errHandler)
+	installCache := kubeinstall.NewCache()
+	go func() {
```
EItanya: Shouldn't this be blocking? the install won't work without it?

test/e2e/istio/e2e_test.go
```diff
@@ -37,43 +36,47 @@ import (
...
-
-		// TODO (ilackarms): add a flag to switch between starting supergloo locally and deploying via cli
-		sgtestutils.DeleteSuperglooPods(kube, superglooNamespace)
-		istioName := "my-istio"
-		glooName := "gloo"
+var _ = Describe("istio e2e", func() {
+	istioName := "my-istio"
+	glooName := "gloo"
 
+	It("it installs istio", func() {
```
EItanya: Not sure that this works. does ginkgo assure that these run serially? I know that "It" statements are meant to be able to to tun independently

PR #347, opened 2019-04-10
Mesh discovery

EItanya:
BOT NOTES: 
resolves https://github.com/solo-io/supergloo/issues/213


api/v1/mesh.proto
```diff
@@ -41,12 +41,34 @@ message Mesh {
...
+
+    DiscoveryMetadata discovery_metadata = 5;
+}
+
+
+// Generic discovery data shared between different meshes
+message DiscoveryMetadata {
+
+    // list of namespaces which we know are being injected by a given mesh
+    string injected_namespace_label = 1;
```
ilackarms: this should probably be a key-value pair

also, note that injection can be overridden for specific pods in a labeled namespace: https://istio.io/docs/setup/kubernetes/additional-setup/sidecar-injection/#policy

api/v1/mesh.proto
```diff
@@ -85,6 +107,11 @@ message MtlsConfig {
     // whether or not mutual TLS should be enabled between pods in this mesh
     bool mtls_enabled = 1;
 
+    //
+    // if mtlsEnabled is false, this field is ignored
+    // indicated whether or not self-signed certs are enabled for a given mesh
+    bool self_signed_cert = 2;
```
ilackarms: this field is istio specific - it indicates whether Citadel should generate a self-signed root cert or expect one to be provided by the user

pkg/meshdiscovery/pkg/discovery/setup_syncer.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func configurePlugins() []MeshDiscovery {
+	plugins := []MeshDiscovery{
+		istio.NewIstioMeshDiscovery(),
+	}
+	return plugins
+}
+
+// start the install event loop
```
ilackarms: fix comment

pkg/meshdiscovery/pkg/discovery/syncer.go
```diff
@@ -0,0 +1,57 @@
...
+	for _, meshDiscoveryPlugin := range s.plugins {
+		meshes, err := meshDiscoveryPlugin.DiscoverMeshes(ctx, snap)
+		if err != nil {
+			multierr = multierror.Append(multierr, err)
+			logger.Errorf(err.Error())
+		}
+		discoveredMeshes = append(discoveredMeshes, meshes...)
+	}
+
+	err := s.meshReconciler.Reconcile("", discoveredMeshes, nil, clients.ListOpts{})
```
ilackarms: do we want to make sure the reconciler uses some labels to indicate these were discovered meshes? perhaps not since we may want to update a user-created mesh

api/v1/mesh.proto
EItanya: Talked about

PR #351, opened 2019-04-12
Linkerd install cli

ilackarms:
BOT NOTES: 
resolves https://github.com/solo-io/supergloo/issues/350


cli/pkg/cmd/install/linkerd.go
```diff
@@ -0,0 +1,80 @@
...
+					return err
+				}
+			}
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return createLinkerdInstall(opts)
+		},
+	}
+	flagutils.AddMetadataFlags(cmd.PersistentFlags(), &opts.Metadata)
```
EItanya: We can add the first three as persistent flags on install rather than repeating them for each install type. That's the point of persistent flags vs normal flags

PR #352, opened 2019-04-14
Mesh crd from discovery

EItanya:


pkg/install/common/mesh_install_syncer.go
```diff
@@ -133,3 +148,30 @@ func (s *MeshInstallSyncer) handleActiveInstalls(ctx context.Context,
...
+							Namespace: mesh.Metadata.Namespace,
+						}
+					}
+				case *v1.Mesh_LinkerdMesh:
+					if meshType.LinkerdMesh.InstallationNamespace == in.InstallationNamespace &&
+						mesh.Metadata.Name == in.Metadata.Name {
+						meshInstall.InstalledMesh = &core.ResourceRef{
+							Name:      mesh.Metadata.Name,
+							Namespace: mesh.Metadata.Namespace,
+						}
```
ilackarms: reduce the code duplication here: 
make the switch statement retrieve the `installNampespace` of the install, then do the loop over the meshes in one code block

pkg/install/common/mesh_install_syncer.go
```diff
@@ -133,3 +148,30 @@ func (s *MeshInstallSyncer) handleActiveInstalls(ctx context.Context,
 	}
 	return nil, nil
 }
+
+func addMeshToInstall(in *v1.Install, meshes v1.MeshList) {
```
ilackarms: in a follow up pr, we should remove this func as well as the `installedMesh` field from install crd

pkg/install/istio/istio_installer.go
```diff
@@ -76,37 +75,7 @@ func (i *defaultIstioInstaller) EnsureIstioInstall(ctx context.Context, install
...
-				InstallationNamespace: installNamespace,
-				IstioVersion:          istio.IstioVersion,
-			},
-		},
-		MtlsConfig: &v1.MtlsConfig{
-			MtlsEnabled:     istio.EnableMtls,
-			RootCertificate: istio.CustomRootCert,
-		},
-	}
+	return nil, nil
```
ilackarms: if this func always `nil`, why do we have `*v1.Mesh` as a return type?

EItanya: I tried to remove it, but it won't be possible until linkerd is discovered as well

pkg/meshdiscovery/pkg/discovery/istio/istio.go
```diff
@@ -38,9 +38,24 @@ func filterIstioMeshes(meshes v1.MeshList) v1.MeshList {
 	return result
 }
 
+func filterIstioInstalls(installs v1.InstallList) v1.InstallList {
```
ilackarms: nit: rename to `getIstioInstalls` (filter implies removing)

pkg/meshdiscovery/pkg/discovery/istio/istio.go
```diff
@@ -104,9 +119,11 @@ func mergeMeshes(discoveredMeshes, existingMeshes v1.MeshList) (v1.MeshList, err
 			},
 		}
 		discoveredMesh.MtlsConfig = discoveredMesh.DiscoveryMetadata.MtlsConfig
-		discoveredMesh.Metadata = core.Metadata{
-			Namespace: getWriteNamespace(),
-			Name:      fmt.Sprintf("istio-%s", discoveredMesh.DiscoveryMetadata.InstallationNamespace),
+		if discoveredMesh.Metadata.Name == "" || discoveredMesh.Metadata.Namespace == "" {
+			discoveredMesh.Metadata = core.Metadata{
+				Namespace: getWriteNamespace(),
+				Name:      fmt.Sprintf("istio-%s", discoveredMesh.DiscoveryMetadata.InstallationNamespace),
```
ilackarms: note that this naming convention will need to be updated when we move to multi cluster

pkg/meshdiscovery/pkg/discovery/istio/istio.go
```diff
@@ -131,6 +148,31 @@ func constructDiscoveryData(ctx context.Context, istioPilotPod *v1.Pod) (*v1.Mes
 		InjectedNamespaceLabel: injectionLabel,
 	}
 	mesh.DiscoveryMetadata = discoveryData
+
+	// If install crd exists, overwrite discovery data
+	for _, install := range existingInstalls {
```
ilackarms: this loop should probably be made into a utility function that we can use in the follow up PR to remove `installedMesh`

EItanya: As a part of my new PR I made all of this code into utilities

pkg/meshdiscovery/pkg/discovery/istio/istio.go
EItanya: Agreed, a lot of changes will probably have to be made

pkg/install/istio/istio_installer.go
ilackarms: the interface is `EnsureIstioInstall(ctx context.Context, install *v1.Install, meshes v1.MeshList) (*v1.Mesh, error)`

ilackarms: not shared

PR #378, opened 2019-04-21
Linkerd prometheus syncer + e2e

ilackarms:
BOT NOTES: 
resolves https://github.com/solo-io/supergloo/issues/377


ilackarms:
/kick
just want to check if we have a flake

PR #384, opened 2019-04-22
Update AWS App Mesh translator

marcogschmidt:
Updated translator logic and added more tests.


pkg/translator/appmesh/configuration.go
```diff
@@ -33,232 +37,345 @@ type AwsAppMeshConfiguration interface {
...
-	return vrListeners
+	return awsMatchers, nil
 }
 
-func podMatchesRule(vnode *appmesh.VirtualNodeData, selector *v1.PodSelector, upstreams gloov1.UpstreamList) (bool, error) {
-	selectorType := selector.GetSelectorType()
-	if selectorType == nil {
-		return false, errors.Errorf("pod selector type cannot be nil")
+// Fails if multiple rules target the given host on different ports.
+func validate(host string, routeMap routesByPort) (port uint32, routes []*appmesh.HttpRoute, err error) {
```
EItanya: We don't really use named return values anywhere because they are fairly difficult to read, could you possibly switch these over to explicit returns

pkg/translator/appmesh/initialize.go
```diff
@@ -16,7 +15,36 @@ import (
 	kubev1 "k8s.io/api/core/v1"
 )
 
-func getPodInfo(mesh *v1.Mesh, kubePod *kubev1.Pod) (*podInfo, error) {
+func getPodsForMesh(meshName string, pods v1.PodList) (awsAppMeshPodInfo, v1.PodList, error) {
+	var appMeshPodList v1.PodList
+	appMeshPods := make(awsAppMeshPodInfo, 0)
+	for _, pod := range pods {
+		kubePod, err := kubernetes.ToKube(pod)
```
EItanya: No need to make this transformation anymore, `v1.Pod` is essentially a copy of the actual kube pod now, this conversion should no longer be nescessary

pkg/translator/appmesh/initialize.go
```diff
@@ -16,7 +15,36 @@ import (
...
+	sort.SliceStable(appMeshPodList, func(i, j int) bool {
+		iPod := appMeshPodList[i]
+		jPod := appMeshPodList[j]
+		return fmt.Sprintf("%s.%s", iPod.Namespace, iPod.Name) < fmt.Sprintf("%s.%s", jPod.Namespace, jPod.Name)
+	})
+
+	return appMeshPods, appMeshPodList, nil
+}
+
+func getPodInfo(meshName string, kubePod *kubev1.Pod) (*podInfo, error) {
```
EItanya: related to the previous comment, switch `*kubev1.Pod` to `*v1.Pod`

pkg/translator/appmesh/configuration.go
marcogschmidt: Sure

pkg/translator/appmesh/initialize.go
marcogschmidt: Oh, nice. Will change

pkg/translator/appmesh/initialize.go
marcogschmidt: Done

PR #393, opened 2019-04-24
AWS App Mesh API client

marcogschmidt:
Added custom AWS App Mesh client that wraps the actual App Mesh client in the AWS Go SDK and exposes only the methods we need with a simplified interface and wired it into the reconciler (which will be implemented in a follow up PR).


pkg/config/appmesh/syncer.go
```diff
@@ -13,11 +13,16 @@ import (
 
 type appMeshConfigSyncer struct {
 	translator appmesh.Translator
+	reconciler Reconciler
 	reporter   reporter.Reporter
 }
 
-func NewAppMeshConfigSyncer(translator appmesh.Translator, reporter reporter.Reporter) (v1.ConfigSyncer, error) {
-	return &appMeshConfigSyncer{translator: translator, reporter: reporter}, nil
+func NewAppMeshConfigSyncer(translator appmesh.Translator, reconciler Reconciler, reporter reporter.Reporter) (v1.ConfigSyncer, error) {
```
EItanya: nit: why does this function return an error

pkg/translator/appmesh/configuration.go
```diff
@@ -23,17 +23,28 @@ type podInfo struct {
...
+type virtualServiceByHost map[string]*appmeshApi.VirtualServiceData
 
-type routesByPort map[uint32][]*appmesh.HttpRoute
+type routesByPort map[uint32][]*appmeshApi.HttpRoute
 type routesByDestinationAndPort map[string]routesByPort
 
+// Snapshot of App Mesh resources. The maps associate a resource name with its data.
+type ResourceSnapshot struct {
+	MeshName        string
+	VirtualNodes    map[string]*appmeshApi.VirtualNodeData
```
EItanya: What was the motivation for making these maps rather than lists?

pkg/config/appmesh/syncer.go
marcogschmidt: Good catch, I guess it is a leftover. I'll remove it.

pkg/translator/appmesh/configuration.go
marcogschmidt: In the reconciler we will need to look up the resources by their name multiple times, so I put them in maps. This indexing could be done by the reconciler, but since we already need to iterate over the vs/vn maps to convert them to arrays, we save some cycles by doing it here.

EItanya: Sounds good to me

PR #398, opened 2019-04-25
Reduce jitter

EItanya:
BOT NOTES: 
resolves https://github.com/solo-io/supergloo/issues/395


pkg/registration/pubsub.go
```diff
@@ -0,0 +1,145 @@
...
+	"sync"
+	"time"
+
+	"github.com/solo-io/go-utils/contextutils"
+	"github.com/solo-io/solo-kit/pkg/api/v1/clients"
+	"github.com/solo-io/solo-kit/pkg/api/v1/eventloop"
+	"go.uber.org/zap"
+)
+
+type Reciever <-chan EnabledConfigLoops
```
marcogschmidt: Nit: typo

pkg/registration/pubsub.go
```diff
@@ -0,0 +1,145 @@
...
+type PubSub struct {
+	subscriberCache []chan EnabledConfigLoops
+	subscriberLock  sync.RWMutex
+}
+
+func NewPubsub() *PubSub {
+	return &PubSub{}
+}
+
+func (r *PubSub) Subscribe() Reciever {
```
marcogschmidt: Looks like `Subscribe`/`Unsubscribe` are only called in this package. If so, can we make them private?

pkg/registration/pubsub_test.go
```diff
@@ -0,0 +1,248 @@
...
+				watches = append(watches, reciever)
+				go func() {
+					<-reciever
+					recievedUpdates++
+				}()
+			}
+			pubSub.publish(ctx, EnabledConfigLoops{})
+			Eventually(func() int {
+				return recievedUpdates
+			}, time.Second*15, time.Second/2).Should(Equal(4))
```
ilackarms: 15 seconds seems a bit long here

EItanya: yea, but it pings every half of a second so unless something goes really wrong, it should get it on the second one most likely

PR #399, opened 2019-04-25
Appmesh reconciler

marcogschmidt:


pkg/config/appmesh/reconciler_helper.go
```diff
@@ -0,0 +1,222 @@
...
+	client   Client
+	snapshot *translator.ResourceSnapshot
+}
+
+func (h *reconcileHelper) createAll() error {
+	if _, err := h.client.CreateMesh(h.ctx, h.snapshot.MeshName); err != nil {
+		return err
+	}
+
+	for _, vn := range h.snapshot.VirtualNodes {
```
EItanya: Is there a specific order this all needs to be done in, or can you parallelize some of it?

marcogschmidt: Good point, I was just thinking about it. There is an order, but there are parallelizable calls. Will update.

PR #410, opened 2019-04-28
Aws mesh config

EItanya:
BOT NOTES: 
resolves https://github.com/solo-io/supergloo/issues/409


ilackarms:
e2e test?

EItanya:
> e2e test?

@ilackarms good call, I'll write one this morning

pkg/meshdiscovery/config/appmesh/syncer.go
```diff
@@ -0,0 +1,111 @@
...
+		Ctx:      ctx,
+		Selector: appmesh.DiscoverySelector,
+	}
+	return s.reconciler.Reconcile("", updatedMeshes, nil, listOpts)
+}
+
+func updatedMesh(mesh *v1.Mesh, config *appmeshcfg.AwsAppMeshConfigurationImpl) *v1.Mesh {
+	result := mesh
+	awsMesh := mesh.GetAwsAppMesh()
+	if awsMesh == nil {
```
marcogschmidt: Since this should never happen (we pass in a list of app meshes), I would return an error here to notify of the unexpected behavior.

pkg/meshdiscovery/config/appmesh/syncer.go
```diff
@@ -0,0 +1,111 @@
...
+	if result.DiscoveryMetadata == nil {
+		result.DiscoveryMetadata = &v1.DiscoveryMetadata{}
+	}
+
+	var meshUpstreams []*core.ResourceRef
+	for _, upstream := range config.UpstreamList {
+		ref := upstream.Metadata.Ref()
+		meshUpstreams = append(meshUpstreams, &ref)
+	}
+	result.DiscoveryMetadata.Upstreams = meshUpstreams
```
marcogschmidt: `DiscoveryMetadata` contains other fields that we could potentially set here with the information at hand, e.g. `enable_auto_inject` which is a property on the appmesh type. Is there a reason you don't set them here?

pkg/meshdiscovery/config/appmesh/appmesh_test.go
```diff
@@ -0,0 +1,118 @@
...
+		syncer = newAppmeshDiscoveryConfigSyncer(reconciler)
+		input = scenarios.SumAppMeshRelatedResources()
+	})
+	AfterEach(func() {
+		ctrl.Finish()
+	})
+
+	It("returns nil with 0 meshes", func() {
+		snap := &v1.AppmeshDiscoverySnapshot{}
+		reconciler.EXPECT().Reconcile(gomock.Any(), nil, gomock.Any(), gomock.Any()).Times(1)
```
marcogschmidt: You could append a `.Return(nil)` here

marcogschmidt: Actually I am not sure what the mock call returns if you don't specify it, probably it default to nil values for the return types

PR #414, opened 2019-04-29
Use Go build tool

marcogschmidt:
This PR switches supergloo over to using the [Solo go build tool](https://github.com/solo-io/build) as part of the effort to standardize our builds and reduce boilerplate `make` configuration.

With this change, every CI build now produces and publishes the build artifacts:
1. Container images:
    - If the build is a release build, push images to `quay`
    - Else push to `gcr`
2. Helm chart:
    - If the build is a release build, push it to the chart repo at `gs://supergloo-helm/`
    - Else push to `gs://supergloo-helm-test/`

This allows any pods deployed to the test k8s clusters during e2e tests to fetch the images that correspond to the current build.

See [squash](https://github.com/solo-io/) for a usage example.


test/e2e/appmesh/e2e_test.go
```diff
@@ -26,7 +26,7 @@ var _ = Describe("E2e", func() {
 		// start discovery
 		var superglooErr error
 		projectRoot := filepath.Join(os.Getenv("GOPATH"), "src", os.Getenv("PROJECT_ROOT"))
-		err := generate.RunWithGlooVersion("dev", "Always", projectRoot, "0.13.18")
+		err := generate.RunWithGlooVersion("dev", "dev", "Always", projectRoot, "0.13.18")
```
mitchdraft: since you're pushing new containers each time, I think you want to use the newly built/pushed images

you can get those through the build libraries, like this:
https://github.com/solo-io/squash/blob/master/test/testutils/utils.go#L38
`buildRun, err := ingest.InitializeBuildRun()`

same for the other locations

cloudbuild.yaml
```diff
@@ -49,6 +49,7 @@ steps:
       - 'PROJECT_ROOT=github.com/solo-io/supergloo'
       - 'GOPATH=/workspace/gopath'
       - 'TAGGED_VERSION=$TAG_NAME'
+      - 'BUILD_ID=$BUILD_ID'
```
mitchdraft: push before you test, and include BUILD_ID and TAGGED_VERSION in your ginkgo-e2e steps and you can use the images you just built

Gopkg.toml
```diff
@@ -48,6 +44,10 @@
   name = "github.com/solo-io/solo-kit"
   version = "0.7.2"
 
+[[constraint]]
+  name = "github.com/solo-io/build"
+  version = "0.0.1"
```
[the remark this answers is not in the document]
marcogschmidt: Sure, I can wait. I have marked this as WIP as I want to double-check what happens to the helm repo with the new behavior.

cloudbuild.yaml
marcogschmidt: Good catch! Will fix, thank you

test/e2e/appmesh/e2e_test.go
marcogschmidt: Yeah, I was meaning to do that in a follow-up PR. This change was simply to make the tests compile. These tests currently use `RunWithGlooVersion` to just deploy discovery and run supergloo locally (notice the `DeleteSuperglooPods` a few lines below`), so the version you pass in there is disregarded. I suspect it was done that way also to get around the issue I am addressing here.

I didn't know about those utility functions though! I'll think about including the change in this PR.

solo-project.yaml
```diff
@@ -0,0 +1,9 @@
+config:
+  release_container_registry:
+    quay:
+      base_url: quay.io
+      organization: solo-io
+  test_container_registry:
+    gcr:
+      base_url: gcr.io
+      project_name: solo-public
```
mitchdraft: when you bump to build v0.0.2, `project_name` becomes `project_id`

PR #441, opened 2019-05-13
Upgrade command

ilackarms:
BOT NOTES: 
resolves https://github.com/solo-io/supergloo/issues/444


cli/pkg/cmd/upgrade/cmd.go
```diff
@@ -0,0 +1,107 @@
+package upgrade
+
```
rickducott: This doesn’t interact well with brew, I thought we were moving away

PR #443, opened 2019-05-14
Filter discovery

rickducott:
BOT NOTES: 
resolves https://github.com/solo-io/supergloo/pull/440

### DIFF
--- a/pkg/meshdiscovery/mesh/syncer.go
+++ b/pkg/meshdiscovery/mesh/syncer.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/solo-io/go-utils/cliutils"
+	"github.com/solo-io/go-utils/stringutils"
 
 	"github.com/hashicorp/go-multierror"
 	"github.com/solo-io/go-utils/contextutils"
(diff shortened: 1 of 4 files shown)


pkg/meshdiscovery/mesh/syncer.go
```diff
@@ -4,12 +4,16 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/solo-io/go-utils/cliutils"
```
EItanya: do you need to use cliutils, I just checked the cliutils lib and it pulls in cobra which is unecessary for the mesh discovery bin

pkg/meshdiscovery/mesh/syncer.go
```diff
@@ -63,3 +68,32 @@ func (s *meshDiscoverySyncer) Sync(ctx context.Context, snap *v1.DiscoverySnapsh
 
 	return multierr.ErrorOrNil()
 }
+
+func FilterOutNamespacesWithInstalls(snap *v1.DiscoverySnapshot) *v1.DiscoverySnapshot {
+	var namespaces []string
+	for _, install := range snap.Installs {
```
EItanya: active installs need to remain in the list

PR #457, opened 2019-05-20
enable smi translation

ilackarms:
this pr contains a large number of features, i plan on breaking it down into smaller PRs as makes sense
BOT NOTES: 
resolves https://github.com/solo-io/supergloo/issues/456


pkg/config/setup/listener.go
```diff
@@ -21,18 +21,26 @@ type superglooConfigLoop struct {
...
 }
 
 func (scl *superglooConfigLoop) Start(ctx context.Context, enabled registration.EnabledConfigLoops) (eventloop.EventLoop, error) {
 	var syncers v1.ConfigSyncers
 
 	if enabled.Istio {
-		istioSyncer, err := createIstioConfigSyncer(ctx, scl.cs)
-		if err != nil {
-			return nil, err
+		if enabled.IstioSmi {
```
EItanya: Going to approve but I Don't love this nested conditional logic for the configuration

EItanya: (from Marco): since `create*ConfigSyncer` have the same signature, we could do something like:
```go
syncerFunc := createIstioConfigSyncer
if enabled.IstioSmi {
    syncerFunc = createSmiConfigSyncer
}
syncer, err := syncerFunc(ctx, scl.cs)
if err != nil {
    return nil, err
}
syncers = append(syncers, istioSyncer)
```

api/v1/selector.proto
```diff
@@ -23,6 +23,12 @@ message PodSelector {
         repeated core.solo.io.ResourceRef upstreams = 1 [(gogoproto.nullable) = false];
     }
 
+    // select pods based on the services they back
+    message ServiceSelector {
```
EItanya: Marco: love this new selector!

pkg/translator/smi/translator_routing.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+}
+
+func trafficSplitsForRule(rule *v1.RoutingRule, upstreams gloov1.UpstreamList, services kubernetes.ServiceList) (v1alpha1.TrafficSplitList, error) {
+	var trafficSplits v1alpha1.TrafficSplitList
+	if rule.Spec == nil {
+		return nil, errors.Errorf("rule spec cannot be nil")
+	}
+	switch ruleType := rule.Spec.RuleType.(type) {
+	case *v1.RoutingRuleSpec_TrafficShifting:
```
EItanya: Marco: We should error here if the traffic shifting rule is not applicable

PR #466, opened 2019-05-29
Update e2e tests

marcogschmidt:
Update end-to-end tests to use containers and chart built during CI build instead of previous releases.

### DIFF
--- a/install/helm/supergloo/generate/generate.go
+++ b/install/helm/supergloo/generate/generate.go
@@ -26,24 +26,15 @@ const (
 	constraint = "constraint"
 )
 
-var (
-	osGlooVersion string
-	rootPrefix    string
-)
+var rootPrefix = ""
 
 func Run(version, imageTag, imageRepoPrefix, pullPolicy string) error {
-	glooVersion, err := getOsGlooVersion()
+	glooVersion, err := getOsGlooVersion(rootPrefix)
 	if err != nil {
 		return err
 	}
-	return RunWithGlooVersion(version, imageTag, imageRepoPrefix, pullPolicy, "", glooVersion)
-}
 
-func RunWithGlooVersion(version, imageTag, imageRepoPrefix, pullPolicy, prefix, glooVersion string) error {
-	rootPrefix = prefix
-	osGlooVersion = glooVersion
-
-	if err := generateValuesYaml(imageTag, pullPolicy, imageRepoPrefix); err != nil {
+	if err := generateValuesYaml(imageTag, pullPolicy, imageRepoPrefix, glooVersion); err != nil {
 		return fmt.Errorf("generating values.yaml failed: %v", err)
 	}
 	if err := generateChartYaml(version); err != nil {
@@ -52,8 +43,8 @@ func RunWithGlooVersion(version, imageTag, imageRepoPrefix, pullPolicy, prefix, 
 	return nil
 }
 
-func getOsGlooVersion() (string, error) {
-	tomlTree, err := parseToml(rootPrefix)
+func getOsGlooVersion(prefix string) (string, error) {
+	tomlTree, err := parseToml(prefix)
 	if err != nil {
 		return "", err
 	}
@@ -66,8 +57,8 @@ func getOsGlooVersion() (string, error) {
 }
 
 func parseToml(prefix string) ([]*toml.Tree, error) {
-	path := filepath.Join(prefix, gopkgToml)
-	config, err := toml.LoadFile(path)
+	tomlPath := filepath.Join(prefix, gopkgToml)
+	config, err := toml.LoadFile(tomlPath)
 	if err != nil {
 		return nil, err
 	}
@@ -116,7 +107,7 @@ func writeYaml(obj interface{}, path string) error {
 	return nil
 }
 
-func generateValuesYaml(imageTag, pullPolicy, imageRepoPrefix string) error {
+func generateValuesYaml(imageTag, pullPolicy, imageRepoPrefix, glooVersion string) error {
 	config, err := readConfig(filepath.Join(rootPrefix, DefaultValues))
 	if err != nil {
 		return err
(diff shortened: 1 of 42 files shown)


install/helm/supergloo/generate/generate.go
```diff
@@ -116,20 +118,29 @@ func writeYaml(obj interface{}, path string) error {
...
+	// Sidecar injector image gets pushed with the same tag as supergloo
+	config.Supergloo.SidecarInjectorImage.Tag = imageTag
+
 	config.Discovery.Deployment.Image.Tag = osGlooVersion
 	config.Discovery.Deployment.Image.PullPolicy = pullPolicy
 
 	config.MeshDiscovery.Deployment.Image.Tag = imageTag
 	config.MeshDiscovery.Deployment.Image.PullPolicy = pullPolicy
 
+	if imageRepoPrefix != "" {
```
mitchdraft: this should never be false, right? I think that will always have a string value, if not, should error, right?

install/helm/supergloo/generate/generate.go
```diff
@@ -143,3 +154,10 @@ func generateChartYaml(version string) error {
 
 	return writeYaml(&chart, filepath.Join(rootPrefix, ChartOutput))
 }
+
+// We want to turn "quay.io/solo-io/gloo" into "<newPrefix>/gloo".
+func replaceRepositoryPrefix(repository, newPrefix string) string {
```
mitchdraft: this is just getting the values: supergloo, discovery, and mesh-discovery
instead of reading them from the helm values, could we get them from a go `const`?

install/helm/supergloo/templates/3-deployment.yaml
```diff
@@ -29,6 +29,8 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            - name: SIDECAR_INJECTOR_IMAGE
+              value: "{{ .Values.supergloo.sidecarInjectorImage.repository }}:{{ .Values.supergloo.sidecarInjectorImage.tag }}"
```
mitchdraft: what about this format instead?
```suggestion
              value: "{{ .Values.supergloo.sidecarInjectorImage.imageRpository }}/sidecar-injector:{{ .Values.supergloo.sidecarInjectorImage.tag }}"
```

install/helm/supergloo/values-defaults.yaml
```diff
@@ -16,6 +16,9 @@ supergloo:
       repository: quay.io/solo-io/supergloo
       pullPolicy: Always
     stats: true
+  sidecarInjectorImage:
+    repository: quay.io/solo-io/sidecar-injector
```
mitchdraft: what about this? (as in above suggestion)
```suggestion
    imageRepository: quay.io/solo-io
```

install/helm/supergloo/values-defaults.yaml
```diff
@@ -16,6 +16,9 @@ supergloo:
...
+    pullPolicy: IfNotPresent
```
mitchdraft: should this be `Always`, as above?

pkg/translator/appmesh/README.md
```diff
@@ -1,7 +1,7 @@
 # AWS App Mesh translator
 This package contains the logic for translating a set of Supergloo Routing Rules to a set of AWS App Mesh resources.
 
-Following are is a high-level description of the steps that are executed during translation for each mesh. The 
+Following is a high-level description of the steps that are executed during translation for each mesh. The
```
mitchdraft:
```suggestion
The following is a high-level description of the steps that are executed during translation for each mesh. The 
```

test/e2e/appmesh/e2e_suite_test.go
```diff
@@ -64,21 +81,23 @@ var _ = BeforeSuite(func() {
...
-	// create sg ns
 	_, err = kube.CoreV1().Namespaces().Create(&kubev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{Name: superglooNamespace},
 	})
 	Expect(err).NotTo(HaveOccurred())
 
-	// start supergloo
+	// start supergloo (requires setting the two envs if running locally)
+	Expect(os.Setenv(registrationsetup.PodNamespaceEnvName, superglooNamespace)).NotTo(HaveOccurred())
+	image := fmt.Sprintf("%s/sidecar-injector:%s", imageRepoPrefix, buildVersion)
```
mitchdraft: sidecar-injector should probably be imported from some `const`

install/helm/supergloo/generate/generate.go
marcogschmidt: Good point, now that the build tool always provides the repo prefix we can get rid of this conditional logic.

install/helm/supergloo/templates/3-deployment.yaml
EItanya: The underlying type has the field `repository` not `imageRepository`
It's a common field among all of our generate scripts

install/helm/supergloo/values-defaults.yaml
EItanya: same as above

install/helm/supergloo/templates/3-deployment.yaml
mitchdraft: Yes, Marco and I discussed changing the underlying spec. Since `sidecar-injector` is a "constant" it should not be changeable through the helm values.
I think Marco is planning on defining a new struct to use for this

EItanya: But then that would be true for every single one of our containers. Currently we use this pattern across all of our projects

install/helm/supergloo/generate/generate.go
marcogschmidt: Yeah, I created a new package in `pkg/constants` with string values that are used in several places, e.g. environment variable names and image names.

install/helm/supergloo/templates/3-deployment.yaml
marcogschmidt: I'd rather have the image name in a  go constant and use that to build the image name in the chart generation script.

marcogschmidt: Oops, did not refresh the page and did not see the new comments. I thought more about it and I'd rather keep the whole image name (e.g.`quay.io/solo-io/supergloo`) as a helm value and concatenate the prefix and the image name in our chart generation script. I don't see the value of splitting the name in 2 in the values file (pun not intended).

test/e2e/appmesh/e2e_suite_test.go
marcogschmidt: Defined it as a constant in the new `constants` package.

pkg/translator/appmesh/README.md
marcogschmidt: Done.

install/helm/supergloo/values-defaults.yaml
marcogschmidt: Based on our conversation yesterday I switched all of them to `IfNotPresent`.

PR #470, opened 2019-06-07
Refactor meshdiscovery

ilackarms:
this pr for the purpose of review only, will be split into smaller chunks


pkg/meshdiscovery/istio/istio_discovery_syncer.go
```diff
@@ -0,0 +1,235 @@
...
+	var istioMeshes v1.MeshList
+	for _, pilot := range pilots {
+		var autoInjectionEnabled bool
+		sidecarInjector, err := snap.Deployments.Find(pilot.namespace, "istio-sidecar-injector")
+		if err == nil && (sidecarInjector.Spec.Replicas == nil || *sidecarInjector.Spec.Replicas > 0) {
+			autoInjectionEnabled = true
+		}
+
+		var smiEnabled bool
+		smiAdapter, err := snap.Deployments.Find(pilot.namespace, "smi-adapter-istio")
```
EItanya: Is it a requirement that they're in the same namespace?

pkg/meshdiscovery/istio/istio_discovery_syncer.go
```diff
@@ -0,0 +1,235 @@
...
+			root := customRootCa.Metadata.Ref()
+			rootCa = &root
+		}
+
+		mtlsConfig := &v1.MtlsConfig{
+			MtlsEnabled:     globalMtlsEnabled,
+			RootCertificate: rootCa,
+		}
+
+		meshUpstreams := func() []*core.ResourceRef {
```
EItanya: Why limit this to the pilot namespace?

pkg/meshdiscovery/istio/istio_discovery_syncer.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+	}
+	return injectedPods, nil
+}
+
+// after the first write, i.e. on any update
+// we are only interested in updating discovery metadata at this point
+// do not want to overwrite user-modified config settings
+// smi is also included here as it's not intended to be user-writeable
+func onlyUpdateDiscoveryMetadata(original, desired *v1.Mesh) (b bool, e error) {
```
rickducott: 1) this function doesn't need the second return argument
2) it's hard to tell what you're trying to accomplish with this function; the name describes the return value but not the side effect.

rickducott: After looking at how it's used, we can ignore (1) (since we're adhering to the reconciler spec). 

I think we should rename to `reconcileMesh` (because this is a very strange looking function without the context that it is getting passed into the reconciler), and then update the comment to indicate what true/false mean in addition to what the intended side effects are. I think false means we don't need to do any reconciling and we're in the desired state. True means we need to do reconciling, but we want to preserve stuff from the original, so we update the desired object and return true.

pkg/meshdiscovery/istio/istio_discovery_syncer.go
rickducott: Reading the tests, it seems to work fine this way already -- trying to figure out why

pkg/meshdiscovery/istio/istio_discovery_syncer.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+func detectPilotDeployments(deployments kubernetes.DeploymentList) ([]pilotDeployment, error) {
+	var pilots []pilotDeployment
+	for _, deployment := range deployments {
+		for _, container := range deployment.Spec.Template.Spec.Containers {
+			if strings.Contains(container.Image, "istio/pilot") {
+				split := strings.Split(container.Image, ":")
+				if len(split) != 2 {
+					return nil, errors.Errorf("invalid or unexpected image format for pilot: %v", container.Image)
```
rickducott: should not fail all discovery due to this error.

pkg/meshdiscovery/istio/istio_discovery_syncer_test.go
```diff
@@ -0,0 +1,373 @@
...
+	"github.com/solo-io/supergloo/test/inputs"
+	appsv1 "k8s.io/api/apps/v1"
+	kubev1 "k8s.io/api/core/v1"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var _ = Describe("IstioDiscoverySyncer", func() {
```
rickducott: We need to add tests for all the error cases (if we do intend to propagate them as errors; less important for things we log). I think the errors should be well-defined so we can match on them specifically, along the lines of: https://github.com/solo-io/sm-marketplace/blob/d670b7cda227429f09228156fa57fa4024278df8/services/apiserver/server/errors.go (the best way we know how to do it so far)

Error cases include: 
1) error detecting mesh policy
2) error detecting pilot
3) error detecting upstreams for pilot due to invalid sidecar config

pkg/meshdiscovery/istio/istio_discovery_syncer.go
ilackarms: injectedPods is a map<string, PodList> where the string is the *namespace of the pilot instance* configuring the pod list sidecars

check out https://github.com/solo-io/supergloo/blob/1d3e101700eebcb2ac08bdcc8b750aacb450cfa0/pkg/meshdiscovery/istio/istio_discovery_syncer.go#L199 to see how it's constructed

pkg/meshdiscovery/istio/istio_discovery_syncer_test.go
```diff
@@ -0,0 +1,373 @@
...
+	Context("pilot present, istio crds registered", func() {
+		var snap *v1.DiscoverySnapshot
+		BeforeEach(func() {
+			crdGetter.shouldSucceed = true
+			snap = &v1.DiscoverySnapshot{
+				Deployments: []*kubernetes.Deployment{istioDeployment("istio-system", "1234")},
+			}
+		})
+		Context("no meshpolicy, no adapter, no smi, no root cert, no injected pods", func() {
+			It("determines the correct namespace and version of the mesh", func() {
```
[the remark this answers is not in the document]
ilackarms: i just wanted to be clear about the base expectation for this (the minimal) case. i feel it's better to have some information than none

cmd/meshdiscovery/README.md
```diff
@@ -0,0 +1,2 @@
+# MeshDiscovery Event Loops
+
```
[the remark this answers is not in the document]
ilackarms: todo for now

rickducott:
Ok, I'm happy with this. Going to remove the "Request Changes", but marked as "Work in Progress" since we want do additional work before merging into master. For additional work, we're going to open PRs against this branch instead of master, to review incremental progress, and then we'll merge this PR into master once we're done.

pkg/meshdiscovery/istio/istio_discovery_syncer.go
```diff
@@ -0,0 +1,307 @@
...
+
+var discoveryLabels = map[string]string{
+	"discovered_by": "istio-mesh-discovery",
+}
+
+func (i *istioDiscoverySyncer) ShouldSync(old, new *v1.DiscoverySnapshot) bool {
+	if old == nil {
+		return true
+	}
+	desired1, err1 := i.desiredMeshes(context.TODO(), old)
```
EItanya: nit: cache response from old

pkg/meshdiscovery/istio/syncer_error_types.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+func newErrorDetectingMeshPolicy(err error) *SyncerError {
+	return &SyncerError{
+		Type:    ErrorType_DetectingMeshPolicy,
+		Message: fmt.Sprintf("detecting default MeshPolicy: %v", err),
+	}
+}
+
+func IsErrorDetectingMeshPolicy(err error) bool {
```
EItanya: If you use go-utils errors package here you can do more powerful error type checking

pkg/meshdiscovery/istio/istio_discovery_syncer.go
ilackarms: what do you mean cache response?

EItanya: `desired1, err1 := i.desiredMeshes(context.TODO(), old)`
This should be the same as
`desired2, err2 := i.desiredMeshes(context.TODO(), new)`
from the previous run.
Not super important, just a minor caching oppurtunity.

ilackarms: a valid suggestion, but in this case i don't think desiredMeshes should incur enough overhead to warrant the extra code complexity

joekelley:
Testing in Service Mesh Hub against containers pushed from this branch and seeing the following error on the mesh discovery pod:
```
E0613 15:14:24.847841       1 reflector.go:134] github.com/solo-io/solo-kit/pkg/api/v1/clients/kube/controller/controller.go:71: Failed to list *v1.Deployment: deployments.apps is forbidden: User "system:serviceaccount:sm-marketplace:mesh-discovery" cannot list resource "deployments" in API group "apps" at the cluster scope
```

Looks like we need to grant get permissions on deployments in the mesh-discovery cluster role

pkg/meshdiscovery/istio/istio_discovery_syncer.go
```diff
@@ -0,0 +1,264 @@
...
+
+type pilotDeployment struct {
+	version, namespace string
+}
+
+func detectPilotDeployments(ctx context.Context, deployments kubernetes.DeploymentList) []pilotDeployment {
+	var pilots []pilotDeployment
+	for _, deployment := range deployments {
+		for _, container := range deployment.Spec.Template.Spec.Containers {
+			if strings.Contains(container.Image, "istio/pilot") {
```
joekelley: I'm running a release of istio with pilot image `gcr.io/istio-release/pilot:release-1.1-latest-daily` that this check doesn't catch

pkg/meshdiscovery/istio/istio_discovery_syncer.go
ilackarms: after all, we do need caching

ilackarms:
/kick

Step #22 - "istio-10x-test":               s: "lock is currently in use",

PR #575, opened 2020-04-14
Migrate skv2

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/574


custom_client.gotmpl
```diff
@@ -0,0 +1,174 @@
+package {{.Version}}
```
EItanya, line 1: Can we put these in a subfolder somewhere, perhaps `api/templates` or just `templates`

generate.go
```diff
@@ -204,7 +233,55 @@ func main() {
...
+					{
+						Kind: "Node",
+					},
+				},
+				CustomTypesImportPath: "k8s.io/api/core/v1",
+				ApiRoot:               "pkg/api/kubernetes",
+				CustomTemplates: map[string]string{
+					"custom_clients/custom_clients.go":          customClientTemplate,
+					"custom_clients/custom_client_providers.go": customClientProviders,
+				},
```
EItanya, line 284: Can we combine these 2 and just render the full set of custotm clients for both, it feels confusing to have them seperate.

custom_client.gotmpl
```diff
@@ -0,0 +1,174 @@
...
+    Create{{ $resource.Kind }}(ctx context.Context, obj *{{ $resource.Kind }}, opts ...client.CreateOption) error
+
+    // Delete deletes the {{ $resource.Kind }} object.
+    Delete{{ $resource.Kind }}(ctx context.Context, key client.ObjectKey, opts ...client.DeleteOption) error
+
+    // Update updates the given {{ $resource.Kind }} object.
+    Update{{ $resource.Kind }}(ctx context.Context, obj *{{ $resource.Kind }}, opts ...client.UpdateOption) error
+
+    // If the {{ $resource.Kind }} object exists, update its spec. Otherwise, create the {{ $resource.Kind }} object.
+    Upsert{{ $resource.Kind }}Spec(ctx context.Context, obj *{{ $resource.Kind }}, opts ...client.UpdateOption) error
```
EItanya, line 80: Is there anything different about this file other than the Upsert function. If so I think we should try and get this into `skv2` rather than maintaining a template which only differs by 1 function

harveyxia: No differences except the Upsert function—one complication with including it in the standard client template is that not all k8s kinds have a `Spec` field, but if we can work around that, I agree that it would be a commonly useful generated method.

generate.go
harveyxia: The reason they are separate is because `Secret`, `ConfigMap`, and `ServiceAccount` do not have spec fields, and so the generated `Upsert` method will have compile errors. I'm also not happy with having two declarations for the same group/version, open to alternatives.

One possibility is to modify the client template to not render the Upsert method for kinds without a Spec field

harveyxia: turns out we don't use any upsert methods other than SecretClient.UpsertData, which I've moved into an isolated function in the one place that it's used, so this isn't an issue

PR #597, opened 2020-04-20
represent AWS credentials as k8s Secret

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/596


pkg/k8s_secrets/aws_creds/secret_aws_creds.go
```diff
@@ -0,0 +1,79 @@
...
+	secretData[AWSAccessKeyID] = []byte(creds.AccessKeyID)
+	secretData[AWSSecretAccessKey] = []byte(creds.SecretAccessKey)
+	return &k8s_core_types.Secret{
+		ObjectMeta: k8s_meta_types.ObjectMeta{
+			Labels:    map[string]string{AWSCredsSecretLabel: "true"},
+			Name:      secretName,
+			Namespace: secretNamespace,
+		},
+		Type: k8s_core_types.SecretTypeOpaque,
+		Data: secretData,
```
yuval-k, line 62: i believe data needs to be base64 encoded?
you can use StringData instead

pkg/k8s_secrets/aws_creds/secret_aws_creds.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/rotisserie/eris"
+	k8s_core_types "k8s.io/api/core/v1"
+	k8s_meta_types "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// TODO unify Secret type naming
+	AWSSecretType = "solo.io/aws-credentials"
+	// Uniquely identifies the AWS account
+	AWSCredsSecretLabel = "solo.io/kubeconfig"
```
EItanya, line 14: Kubeconfig?

harveyxia: copy-pasta mistake, removing label usage given our consensus on using Secret type

pkg/k8s_secrets/aws_creds/secret_aws_creds.go
harveyxia: after brief discussion with @GrahamGoudeau, to clarify, writing to either `Data` or `StringData` does not affect base64 encoding, it just affects whether the data is represented in Golang as a `map[string][]byte` or as a `map[string]string`. The k8s clients should do the base64 encoding upon write

PR #616, opened 2020-04-22
Clean up additional resources during meshctl uninstall

GrahamGoudeau:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/593
resolves https://github.com/solo-io/service-mesh-hub/issues/603


cli/pkg/tree/cluster/deregister/deregistration.go
```diff
@@ -78,14 +116,101 @@ func (c *clusterDeregistrationClient) Run(ctx context.Context, kubeCluster *zeph
...
+		return FailedToCleanUpKubeConfigCrd(err, kubeCluster.GetName())
+	}
+
+	err = c.cleanUpServiceAccounts(ctx, clientForCluster, kubeCluster)
 	if err != nil {
+		return FailedToCleanUpServiceAccount(err, kubeCluster.GetName())
+	}
+
+	// the CSR agent installs only CRDs from the security group. Remove only those
+	_, err = c.crdRemover.RemoveCrdGroup(ctx, kubeCluster.GetName(), config.RestConfig, zephyr_security_scheme.SchemeGroupVersion)
```
harveyxia, line 145: To future proof this should we just uninstall all `zephyr_*` CRDs?

cli/pkg/tree/cluster/deregister/deregistration.go
```diff
@@ -78,14 +116,101 @@ func (c *clusterDeregistrationClient) Run(ctx context.Context, kubeCluster *zeph
...
+	var intermediateCerts []*k8s_core_v1.Secret
+	for _, secretIter := range allSecrets.Items {
+		secret := secretIter
+
+		if secret.Type == cert_secrets.IntermediateCertSecretType {
+			intermediateCerts = append(intermediateCerts, &secret)
+		}
+	}
+
+	for _, tlsSecretIter := range intermediateCerts {
```
harveyxia, line 186: can the logic in this loop be merged with the loop above?

codegen/charts/service_mesh_hub/generate_service_mesh_hub.go
```diff
@@ -15,34 +15,10 @@ import (
...
-// Istio clients
-//go:generate mockgen -package mock_istio_security_clients -destination ./test/mocks/clients/istio/security/v1alpha3/clients.go github.com/solo-io/service-mesh-hub/pkg/api/istio/security/v1beta1 AuthorizationPolicyClient
-//go:generate mockgen -package mock_istio_networking_clients -destination ./test/mocks/clients/istio/networking/v1beta1/clients.go github.com/solo-io/service-mesh-hub/pkg/api/istio/networking/v1alpha3 DestinationRuleClient,EnvoyFilterClient,GatewayClient,ServiceEntryClient,VirtualServiceClient
-// Linkerd clients
-//go:generate mockgen -package mock_linkerd_clients -destination ./test/mocks/clients/linkerd/v1alpha2/clients.go github.com/solo-io/service-mesh-hub/pkg/api/linkerd/v1alpha2 ServiceProfileClient
-// SMI clients
-//go:generate mockgen -package mock_smi_clients -destination ./test/mocks/clients/smi/split/v1alpha1/clients.go github.com/solo-io/service-mesh-hub/pkg/api/smi/split/v1alpha1 TrafficSplitClient
 func main() {
```
harveyxia, line 20: While we're here cleaning up codegen, should we also break out the docs generation into a separate file in case we want to just regenerate docs?

cli/pkg/tree/cluster/deregister/deregistration.go
GrahamGoudeau: This is definitely not as future-proof as it could be.

However, removing all zephyr_* CRDs is one of the bugs this addresses; when the management plane cluster IS one of the registered clusters, you end up removing all your kubeconfig resources before you're done with them and stuff gets out of whack

PR #619, opened 2020-04-23
Handle mesh control plane deletions and reconcile Mesh state at pod startup

GrahamGoudeau:
End-to-end tested by scaling mesh-discovery down to zero, un-installing Istio, then scaling discovery back up. The mesh went away.

This also contains that cleanup we've been meaning to do, with no longer attaching the cluster name to deployments in-memory, instead passing them through the interface


services/mesh-discovery/pkg/discovery/mesh/mesh_finder.go
```diff
@@ -70,8 +92,70 @@ func (m *meshFinder) GenericDeployment(deployment *apps_v1.Deployment) error {
 	return nil
 }
 
+// When the pod starts up, we reconcile the existing state of discovered resources with a newly-computed set of discovered resources.
+// If the newly-computed set is missing entries from the current state, we must have missed an event, and we must reconcile the two.
+func (m *meshFinder) reconcileExistingState() error {
```
harveyxia, line 97: I wonder if we should be performing a reconcile for create/update/delete of all discovered resources, not just on initialization. It would be more compute intensive, but it should decrease the likelihood that SMH from gets into inconsistent states. We've been throwing this idea around when discussing event handling in the past

PR #632, opened 2020-04-28
Garbage collect workloads when they are orphaned

GrahamGoudeau:


services/mesh-discovery/pkg/discovery/mesh-workload/mesh_workload_finder.go
```diff
@@ -65,35 +67,169 @@ type meshWorkloadFinder struct {
...
+		// then do nothing since we have not missed any event
+		if notDeleted && existingMeshWorkload.Spec.Equal(discoveredWorkload.Spec) {
+			continue
+		}
+
+		if discoveredWorkload != nil && !existingMeshWorkload.Spec.Equal(discoveredWorkload.Spec) {
+
+			// we missed an update event, make sure the state of the cluster matches what we discovered from the pod
+			existingMeshWorkload.Spec = discoveredWorkload.Spec
+			err := m.localMeshWorkloadClient.UpdateMeshWorkload(m.ctx, &existingMeshWorkload)
```
harveyxia, line 172:
```suggestion
			err := m.localMeshWorkloadClient.UpsertMeshWorkloadSpec(m.ctx, &existingMeshWorkload)
```

services/mesh-discovery/pkg/discovery/mesh-workload/mesh_workload_finder.go
```diff
@@ -65,35 +67,169 @@ type meshWorkloadFinder struct {
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+func (m *meshWorkloadFinder) handleMeshDelete(meshBeingDeleted *zephyr_discovery.Mesh) error {
```
harveyxia, line 188: nit:

```suggestion
func (m *meshWorkloadFinder) handleMeshDelete(deletedMesh *zephyr_discovery.Mesh) error {
```

services/mesh-discovery/pkg/discovery/mesh-workload/mesh_workload_finder.go
```diff
@@ -65,35 +67,169 @@ type meshWorkloadFinder struct {
...
+
+	// for each mesh that we have rediscovered, ensure that we can match it to a mesh that currently exists in the cluster
+	for _, meshWorkloadIter := range existingMeshWorkloads.Items {
+		existingMeshWorkload := meshWorkloadIter
+
+		discoveredWorkload, notDeleted := rediscoveredWorkloadsByName[existingMeshWorkload.GetName()]
+
+		// if we both 1. rediscovered the workload from a pod, and 2. its spec is still the same,
+		// then do nothing since we have not missed any event
+		if notDeleted && existingMeshWorkload.Spec.Equal(discoveredWorkload.Spec) {
```
harveyxia, line 164: The conditionals here might be clearer if expressed as the following:

```
if notDeleted {
  if existingMeshWorkload.Spec.Equal(discoveredWorkload.Spec) {
    // do nothing
  } else {
    // update MeshWorkload
  }
} else {
  // delete existing
}
```

services/mesh-discovery/pkg/discovery/mesh-workload/mesh_workload_finder.go
```diff
@@ -102,53 +238,72 @@ func (d *meshWorkloadFinder) handleMeshCreate(mesh *zephyr_discovery.Mesh) error
...
 
-func (d *meshWorkloadFinder) handlePodCreate(pod *k8s_core_types.Pod) error {
-	pod.SetClusterName(d.clusterName)
-	logger := logging.BuildEventLogger(d.ctx, logging.CreateEvent, pod)
+func (m *meshWorkloadFinder) handlePodDelete(pod *k8s_core_types.Pod) error {
+	logger := logging.BuildEventLogger(m.ctx, logging.DeleteEvent, pod)
+
+	logger.Debugf("Handling delete for pod %s.%s", pod.GetName(), pod.GetNamespace())
+
+	discoveredMeshWorkload, err := m.discoverMeshWorkload(pod)
```
harveyxia, line 282: I'm tempted to want to add a new method `BuildObjectKey` to the `MeshWorkloadScanner` interface, since [we only need the ObjectMeta here](https://github.com/solo-io/service-mesh-hub/blob/20b4dd01bd1f3afaeb9fcfc9d375c6b10533a3cd/services/mesh-discovery/pkg/discovery/mesh-workload/istio/istio_mesh_workload_scanner.go#L43-L47):

```
               &zephyr_core_types.ResourceRef{
			Name:      deployment.Name,
			Namespace: deployment.Namespace,
			Cluster:   pod.ClusterName,
		}
```

but that might be overkill since it's recomputing the MeshWorklaod doesn't seem particularly resource heavy, so I'll leave it up to you

services/mesh-discovery/pkg/discovery/mesh-workload/mesh_workload_finder.go
```diff
@@ -239,25 +396,32 @@ func (d *meshWorkloadFinder) discoverMeshWorkload(pod *k8s_core_types.Pod) (*zep
...
+	m.discoveredMeshTypesMutex.RUnlock()
+
+	for _, discoveredMeshType := range discoveredMeshTypes {
+		meshWorkloadScanner, ok := m.meshWorkloadScannerImplementations[zephyr_core_types.MeshType(discoveredMeshType)]
 		if !ok {
 			return nil, eris.Errorf("Missing mesh workload scanner implementation for mesh type %d", discoveredMeshType)
 		}
 
-		discoveredControllerRef, discoveredMeshWorkloadObjectMeta, err := meshWorkloadScanner.ScanPod(d.ctx, pod)
+		// some of the workload scanners depend on this being set in memory
```
harveyxia, line 409: i've made that change here :) 

https://github.com/solo-io/service-mesh-hub/pull/631/files#diff-d51ecc1aa5081938a7089545c3d95f72R33

services/mesh-discovery/pkg/discovery/mesh-workload/mesh_workload_finder.go
GrahamGoudeau: I have a slight preference for the phrasing as-is because of the reduced indentation/scoping, but let me know if you feel strongly

services/mesh-discovery/pkg/discovery/mesh-workload/mesh_workload_finder.go
GrahamGoudeau: Yeah I think I'll leave it as-is, would prefer to avoid recomputing the mesh workload. Again let me know if you want to talk more

PR #637, opened 2020-04-28
initial pass at a troubleshooting doc

christian-posta:


docs/content/guides/troubleshooting.md
```diff
@@ -0,0 +1,104 @@
...
+NAME                   TYPE           CLUSTER-IP    EXTERNAL-IP      PORT(S)                                                                                                                                      AGE
+istio-ingressgateway   LoadBalancer   10.8.49.163   35.233.234.111   15020:32347/TCP,80:32414/TCP,443:32713/TCP,15029:31362/TCP,15030:32242/TCP,15031:31899/TCP,15032:32471/TCP,31400:31570/TCP,15443:30416/TCP   4d2h
+istio-pilot            ClusterIP      10.8.54.191   <none>           15010/TCP,15011/TCP,15012/TCP,8080/TCP,15014/TCP,443/TCP                                                                                     4d2h
+istiod                 ClusterIP      10.8.51.94    <none>           15012/TCP,443/TCP      
+
+```
+
+Note, the external-ip. Any `ServiceEntry` created for cross-cluster service discovery will use this external ip. The logic for getting the `ingressgateway` IP can be found in the [federation code base](https://github.com/solo-io/service-mesh-hub/blob/master/services/mesh-networking/pkg/federation/dns/external_access_point_getter.go#L85) of the `mesh-networking` service.
+
+##### What Istio versions are supported?
```
GrahamGoudeau: might be worth flagging the 1.4 -> 1.5 upgrade issues as well

docs/content/guides/troubleshooting.md
```diff
@@ -0,0 +1,104 @@
...
+
+The last bullet in the list, checking federation status, is likely the most helpful especially after you've tried to apply a `VirtualMesh`. Often it's best to check the `VirtualMesh` CR `status` field to make sure it doesn't see any issues:
+
+```shell
+kubectl get virtualmesh -n service-mesh-hub -o jsonpath='{.items[].status}'
+```
+
+NOTE: all of the Service Mesh Hub CRs have a `status` field that give you an indication of what has happened with its processing. Service Mesh Hub, like most operator-style controllers, implements a state machine and helps resources transition between states. All of the current state transitions are available on this `status` field of any of the Service Mesh Hub CRs.
+
+## Logging
```
GrahamGoudeau: You should be able to set `"DEBUG_MODE"` to any nonempty string in any of our pod env vars to turn on way more verbose logging, may also be worth mentioning. Similarly there's a `"LOG_LEVEL"` env var

docs/content/guides/troubleshooting.md
```diff
@@ -0,0 +1,104 @@
...
+
+Knowing how to get logs from the Service Mesh Hub components is crucial for getting feedback about what's happening. Service Mesh Hub has three core components:
+
+* `csr-agent` - responsible for creating certificate/key pairs and issuing a Certificate Signing Request for a particular mesh -- this is used in Mesh identity federation
```
GrahamGoudeau: I guess we could consider the csr-agent to be a core component- but it won't be present in the management plane if they haven't registered that cluster. May be worth a quick parenthetical about that

docs/content/guides/troubleshooting.md
```diff
@@ -0,0 +1,104 @@
...
+
+When troubleshooting various parts of the Service Mesh Hub functionality, you will likely want to see the logs for some of these components. For example, when creating a new `VirtualMesh` and you see something like `PROCESSING_ERROR` in the `federation` status, check the logs for the `mesh-networking` component like this:
+
+```shell
+kubectl logs -f deploy/mesh-networking -n service-mesh-hub
+```
+
+This will give you valuable insight into the control loop that manages the state of a `VirtualMesh`
+
+## Common issues
```
GrahamGoudeau: I think we should probably mention that if someone sees a resource version conflict in the logs, they should file an issue on Github under the "Stability" label. I expect more people may start running into that as we get more people with their hands on the app

docs/content/guides/troubleshooting.md
christian-posta: NICE! Okay will add that

PR #640, opened 2020-04-29
added kube dns docs

christian-posta:


docs/content/operations/customize_dns.md
```diff
@@ -0,0 +1,271 @@
...
+
+```shell
+kubectl get cm/coredns -n istio-system
+NAME      DATA   AGE
+coredns   1      179m
+```
+
+Let's add [Server entries](https://coredns.io/manual/plugins/) with plugins to the `istio-coredns-plugin` to stub out domains like the following:
+
+{{< highlight yaml "hl_lines=11-16" >}}
```
GrahamGoudeau, line 50: I think you may want one more line highlighted here

Screen Shot 2020-04-29 at 6 19 53 PM

docs/content/operations/customize_dns.md
```diff
@@ -0,0 +1,271 @@
...
+        cache 30
+        loop
+        reload
+        loadbalance
+    }
+EOF
+{{< /tab >}}
+
+
+{{< tab name="coredns (==1.4.0)" codelang="shell" >}}
```
GrahamGoudeau, line 224: you have two tabs here that say ==1.4.0

PR #647, opened 2020-04-30
Appmesh multiple clusters

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/630


ci/setup-kind-single.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
+
+#####################################
+#
+# Set up a single Kind cluster with SMH installed
```
harveyxia, line 5: This saves some time testing e2e if you don't need the full multicluster/multi-Istio/multi-bookinfo shebang

services/mesh-discovery/pkg/discovery/cluster-tenancy/k8s/cluster_tenancy_finder.go
```diff
@@ -0,0 +1,152 @@
...
+	return nil
+}
+
+// Recompute this cluster's tenancy for the given Mesh
+func (c *clusterTenancyFinder) reconcileTenancyForMesh(
+	ctx context.Context,
+	mesh *zephyr_discovery.Mesh,
+) error {
+	// Currently multicluster tenancy is only support for AppMesh
+	if mesh.Spec.GetAwsAppMesh() == nil {
```
GrahamGoudeau, line 114: So I'm wondering a bit if these abstraction boundaries are what we want- there are a couple things I'm looking at

1. it would be great if all the app mesh logic was constrained to the app mesh tenancy scanner- do we want to spread knowledge about that mesh to this higher level component too?
2. the app mesh tenancy scanner depends upon the behavior of this finder- see the bail-out on the IsNotFound error in `UpdateMeshTenancy`

wondering if to address each of those: 
1. we process all meshes and let the implementations decide what they want to process
2. we could make the scanner implementations more "Pure"- eg, to sketch it out:

```go
type Scanner interface {
  // has no side effects- run on Reconcile and on Delete
  PodIsMemberOfMesh(pod *Pod, mesh *Mesh) (bool, error)
  
  // the two side-effecting methods - these would update the app mesh cluster membership
  AddMeshToCluster(clusterName string, mesh *Mesh) error
  RemoveMeshFromCluster(clusterName string, mesh *Mesh) error
}
```

maybe scanner isn't 100% the right name anymore, but I think drawing the line a little more concretely around the app mesh stuff, as well as no longer making the Scanner thing dependent on the finder's event-handling behavior, would probably be worthwhile

harveyxia: A+ feedback, thanks

services/mesh-discovery/pkg/discovery/cluster-tenancy/k8s/interfaces.go
```diff
@@ -0,0 +1,29 @@
...
+	StartRegistration(
+		ctx context.Context,
+		podEventWatcher k8s_core_controller.PodEventWatcher,
+		meshEventWatcher discovery_controllers.MeshEventWatcher,
+	) error
+}
+
+type ClusterTenancyRegistrar interface {
+	// Scan the pod for existence of a Mesh, return Mesh if found.
+	MeshForWorkload(ctx context.Context, pod *k8s_core.Pod) (*zephyr_discovery.Mesh, error)
```
GrahamGoudeau, line 24: on an initial glance, I would wonder how this method is different from our normal mesh discovery, as the signature is pretty similar.

I think renaming this to `MeshFromSidecar` or something similar involving the proxy sidecar would probably differentiate it well enough from mesh discovery, as well as illustrate our longish term plans for this method a little better

services/mesh-discovery/pkg/discovery/mesh-workload/k8s/mesh_workload_finder.go
```diff
@@ -431,3 +445,12 @@ func (m *meshWorkloadFinder) createOrUpdateWorkload(discoveredWorkload *zephyr_d
 	mw.Labels = discoveredWorkload.Labels
 	return m.localMeshWorkloadClient.UpdateMeshWorkload(m.ctx, mw)
 }
+
+func (m *meshWorkloadFinder) clusterHostsMesh(mesh *zephyr_discovery.Mesh) bool {
+	// AppMesh's data plane can span multiple clusters
+	if mesh.Spec.GetAwsAppMesh() != nil {
+		return stringutils.ContainsString(m.clusterName, mesh.Spec.GetAwsAppMesh().GetClusters())
+	} else {
+		return mesh.Spec.GetCluster().GetName() == m.clusterName
```
GrahamGoudeau, line 454: This method is the same as `ClusterHostsMesh` now. Can we remove this one?

services/mesh-discovery/pkg/discovery/cluster-tenancy/k8s/utils.go
```diff
@@ -0,0 +1,17 @@
+package k8s_tenancy
+
+import (
+	"github.com/solo-io/go-utils/stringutils"
+	zephyr_discovery "github.com/solo-io/service-mesh-hub/pkg/api/discovery.zephyr.solo.io/v1alpha1"
+)
+
+func ClusterHostsMesh(clusterName string, mesh *zephyr_discovery.Mesh) bool {
```
GrahamGoudeau, line 8: prob could use some documentation on this, noting that it just looks up metadata we've already recorded on the objects rather than going and doing some kind of computation

services/mesh-discovery/pkg/discovery/cluster-tenancy/k8s/interfaces.go
harveyxia: good call

PR #649, opened 2020-05-01
Access control enforcement GC and fail on _suite_test.go files missing

GrahamGoudeau:


services/mesh-networking/pkg/access/access-control-enforcer/enforcer_loop.go
```diff
@@ -66,7 +66,19 @@ func (e *enforcerLoop) Start(ctx context.Context) error {
...
-			logger.Debugf("Ignoring event: %+v", virtualMesh)
+			logger.Debugw("event handler enter",
+				zap.Any("spec", virtualMesh.Spec),
+				zap.Any("status", virtualMesh.Status),
+			)
+
+			// manually set this to false so that things get cleaned up
+			virtualMesh.Spec.EnforceAccessControl = false
+
+			err := e.enforceGlobalAccessControl(ctx, virtualMesh)
```
harveyxia, line 77: If this fails, we won't get another OnDelete event for the virtualMesh, so the global access control enforcement may be permanently set.

I wonder if this is a valid use case for a requeue

ci/assertAllSuitesExist.bash
```diff
@@ -0,0 +1,13 @@
+#########################################################################
```
rickducott, line 1: 👏

services/mesh-networking/pkg/access/access-control-enforcer/enforcer_loop.go
harveyxia: Discussed offline—we'll leave a comment and make an issue for now, and defer settling on an approach until we have more discussion with the team

GrahamGoudeau: https://github.com/solo-io/service-mesh-hub/issues/650

PR #652, opened 2020-05-04
Reduce released CLI binary size by 75%

GrahamGoudeau:
TODO: Bump go-mod-make version number
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/651

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,8 @@ ifeq ($(TAGGED_VERSION),)
 endif
 VERSION ?= $(shell echo $(TAGGED_VERSION) | cut -c 2-)
 
-LDFLAGS := "-X github.com/solo-io/service-mesh-hub/pkg/version.Version=$(VERSION)"
+# "-s -w" strips out the debug symbol tables from the binaries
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -62,6 +63,7 @@ update-deps: mod-download
 	GO111MODULE=off go get -u github.com/google/wire/cmd/wire
 	GO111MODULE=off go get -u github.com/golang/mock/gomock
 	GO111MODULE=off go install github.com/golang/mock/mockgen
+	brew install upx
```
harveyxia, line 66: Does our cloudbuild image have brew installed?

christian-posta:
Let's make this an option to reduce the binary, but should not be the default. Let's make sure our builds include DWARF tables for debugging.

Makefile
GrahamGoudeau: ah, I had checked that our unit test CI build doesn't execute this target, but I just found that the docs and codegen does execute this target. So if we went with this I guess I'd move it out. But it sounds like we're not moving forward on this at all so I guess it's irrelevant

PR #655, opened 2020-05-04
Rename mesh-platform to compute-target, AWS secret handler move

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/653
resolves https://github.com/solo-io/service-mesh-hub/issues/654


services/mesh-discovery/pkg/compute-target/aws/api_handler.go
```diff
@@ -74,24 +74,24 @@ func (a *awsCredsHandler) MeshPlatformAdded(ctx context.Context, secret *k8s_cor
...
 	if !ok {
-		logger.Errorf("Error stopping RestAPIDiscoveryReconciler for meshPlatform %s, entry not found in map.", secret)
+		logger.Errorf("Error stopping RestAPIDiscoveryReconciler for compute target %s, entry not found in map.", secret)
 	}
 	cancelFunc()
 	return nil
 }
 
-func (a *awsCredsHandler) buildContext(parentCtx context.Context, meshPlatformName string) (context.Context, context.CancelFunc) {
+func (a *awsCredsHandler) buildContext(parentCtx context.Context, computeTargetName string) (context.Context, context.CancelFunc) {
```
GrahamGoudeau, line 92: just thinking out loud, no action to take here: as this stuff progresses we'll probably have to unify our own mental model of how to refer to things- I could see "compute target name" being like "ec2", "kubernetes", etc. I wonder if we shouldn't do something like "computeTargetId" and "computeTargetKind"/"computeTargetType" or something. 

idk, just a thing for us to keep in mind

harveyxia: That's a good point, currently we are simply using a string (formerly `clusterName` or `clusterId`, whatever was provided by the user during `meshctl cluster register`) as the compute target identifier.

I see two questions here:

1. For automatically discovered compute targets, what should the identifier be? How do we handle ID conflicts with user supplied ID's for manually registered targets?
2. Will we need to record and utilize any additional metadata? Does SMH need to have an ontology of the different compute target types and cloud providers?

PR #658, opened 2020-05-05
Cluster tenancy pure reconcile

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/657


services/mesh-discovery/pkg/discovery/cluster-tenancy/k8s/appmesh/appmesh_tenancy_scanner.go
```diff
@@ -64,6 +64,10 @@ func (a *appmeshTenancyScanner) MeshFromSidecar(
 	return mesh, nil
 }
 
+func (a *appmeshTenancyScanner) MeshContainsCluster(clusterName string, mesh *zephyr_discovery.Mesh) bool {
```
GrahamGoudeau, line 67: I still think that the owner-relationship is inverted from this: a cluster is the thing housing a mesh, because the mesh runs on compute orchestrated by the cluster. So we talk about a mesh "residing on" a cluster, or similar. Maybe this name should just be flipped.

Or even something like `MeshRegisteredOnCluster`, to thematically keep it consistent with the rest of this interface

services/mesh-discovery/pkg/discovery/cluster-tenancy/k8s/cluster_tenancy_finder.go
```diff
@@ -90,77 +92,88 @@ func (c *clusterTenancyFinder) StartRegistration(
...
+      update mesh with cluster
+   Else if cluster does not host mesh and cluster registered on mesh:
+      update Mesh by removing cluster
+*/
+func (c *clusterTenancyFinder) reconcile(ctx context.Context) error {
+	podList, err := c.podClient.ListPod(ctx)
+	if err != nil {
+		return err
+	}
+	meshesOnCluster := make(map[ClusterTenancyRegistrar][]client.ObjectKey)
```
GrahamGoudeau, line 108: we're not actually using this map as a map, it's just a list of registrar/objectkey pairs. So I think it would be easier to reason about what the data structure is used for if that was its real type: 

```
[]struct {
  registrar ClusterTenancyRegistrar
  meshesOnCluster []client.ObjectKey
}
```

services/mesh-discovery/pkg/discovery/cluster-tenancy/k8s/cluster_tenancy_finder.go
```diff
@@ -90,77 +92,88 @@ func (c *clusterTenancyFinder) StartRegistration(
...
-	if err != nil {
-		return err
-	}
-	for _, pod := range podList.Items {
-		pod := pod
-		err := c.reconcileTenancyForPod(ctx, &pod)
-		if err != nil {
-			return err
+	meshObjKey := clients.ObjectMetaToObjectKey(mesh.ObjectMeta)
+	for registrar, meshOnCluster := range meshesOnCluster {
```
GrahamGoudeau, line 161: probably want the plural here for readability- `meshesOnCluster`. But see my other comment about the type of this data structure

services/mesh-discovery/pkg/discovery/cluster-tenancy/k8s/cluster_tenancy_finder.go
```diff
@@ -90,77 +92,88 @@ func (c *clusterTenancyFinder) StartRegistration(
...
+	}
+	meshList, err := c.localMeshClient.ListMesh(ctx)
 	if err != nil {
 		return err
 	}
-	if mesh == nil {
-		return nil
+	for _, mesh := range meshList.Items {
+		mesh := mesh
+		err := c.reconcileMesh(ctx, meshesOnCluster, &mesh)
```
GrahamGoudeau, line 130: This should be something like `reconcileClusterTenancy`, rather than `reconcileMesh`, right? That's the outcome of this method, the mesh is just a detail

services/mesh-discovery/pkg/discovery/cluster-tenancy/k8s/cluster_tenancy_finder.go
```diff
@@ -90,77 +92,88 @@ func (c *clusterTenancyFinder) StartRegistration(
...
-	}
-	for _, pod := range podList.Items {
-		pod := pod
-		err := c.reconcileTenancyForPod(ctx, &pod)
-		if err != nil {
-			return err
+	meshObjKey := clients.ObjectMetaToObjectKey(mesh.ObjectMeta)
+	for registrar, meshOnCluster := range meshesOnCluster {
+		meshContainsCluster := registrar.MeshContainsCluster(c.clusterName, mesh)
+		if meshContainsCluster && !containsMesh(meshObjKey, meshOnCluster) {
```
GrahamGoudeau, line 163: there's a lot of "contains" going on 🤔 see my comment on the name of the `MeshContainsCluster` method, something like that change may help this

services/mesh-discovery/pkg/discovery/cluster-tenancy/k8s/cluster_tenancy_finder.go
harveyxia: ah right, we've discussed this in the past 👍

services/mesh-discovery/pkg/discovery/cluster-tenancy/k8s/appmesh/appmesh_tenancy_scanner.go
harveyxia: Renamed to "ClusterHostsMesh"

services/mesh-discovery/pkg/discovery/cluster-tenancy/k8s/cluster_tenancy_finder.go
harveyxia: So it actually [is being used](https://github.com/solo-io/service-mesh-hub/pull/658/files#diff-a2a7b8073012c919a7bac9b17bc69bd1R116) as a map to retrieve the meshes that a particular registrar discovered,  (and this is because the reconciliation is scoped by registrar between L111 and L122)

services/mesh-discovery/pkg/discovery/cluster-tenancy/k8s/cluster_tenancy_finder.go
harveyxia: renamed to `clusterHostsMesh`

PR #662, opened 2020-05-07
Traffic Policy IR Design Doc

GrahamGoudeau:


design-docs/traffic-policy-ir.md
```diff
@@ -0,0 +1,84 @@
...
+
+We need to decide how to handle validation failures in a way that does not involve destructively mutating
+generated service mesh configuration when one or more traffic policies gets into an invalid state. We also
+want to prevent an invalid Traffic Policy from invalidating all the other Traffic Policies that impact the same
+translated resource. To state it another way, we want to ensure that if a single Traffic Policy is invalid,
+other (valid) Traffic Policies can continue to be updated and translated successfully. 
+
+## Proposed Design
+
+We propose introducing a *new CRD*: `TrafficPolicyIntermediateRepresentation` (please suggest names!) with the following spec:
```
ilackarms, line 29: `TrafficPolicyRevision` ?

GrahamGoudeau: I'm not sure if "Revision" applies, we're not keeping a versioned history here, nor do I think we really have to

harveyxia: In decreasing order of my preference:

1. `TrafficPolicyAggregate`
2. `TrafficPolicyMerged`
3. `TrafficPolicyCombined`

design-docs/traffic-policy-ir.md
```diff
@@ -0,0 +1,84 @@
...
+*Issue Link:* https://github.com/solo-io/service-mesh-hub/issues/585
+
+## Problem Summary
+
+Traffic Policies go through two rounds of validation:
+
+1. Individual validation
+    - eg ensure no negative retry rates
+1. Merge validation
+    - eg ensure no conflicting retry policies across multiple different Traffic Policies
```
harveyxia, line 16:
```suggestion
    - eg ensure no conflicting retry policies across multiple different Traffic Policies that apply to the same target (Istio) or source (AppMesh)
```

design-docs/traffic-policy-ir.md
```diff
@@ -0,0 +1,84 @@
...
+
+Handling these validation failures is complicated by the fact that multiple Traffic Policies can impact
+the same translated resource, like an Istio Virtual Service.
+
+We need to decide how to handle validation failures in a way that does not involve destructively mutating
+generated service mesh configuration when one or more traffic policies gets into an invalid state. We also
+want to prevent an invalid Traffic Policy from invalidating all the other Traffic Policies that impact the same
+translated resource. To state it another way, we want to ensure that if a single Traffic Policy is invalid,
+other (valid) Traffic Policies can continue to be updated and translated successfully.
```
harveyxia, line 25: Perhaps we should include a simple picture/diagram to illustrate this, something like this: 
Image from iOS

design-docs/traffic-policy-ir.md
```diff
@@ -0,0 +1,84 @@
...
+    message ValidatedTrafficPolicy {
+        string name = 1;
+        string namespace = 2;
+
+        // this essentially records the last known good state of the user-written traffic policy
+        TrafficPolicySpec traffic_policy_spec = 3;
+    }
+
+    // the destination which the merged traffic policies target 
+    ResourceRef destination_mesh_service = 1;
```
harveyxia, line 42: Given that AppMesh express access control by configuring the source, or client (see ["backends"](https://docs.aws.amazon.com/app-mesh/latest/APIReference/API_VirtualNodeSpec.html), perhaps we should name this "translation_target" to be generic?

design-docs/traffic-policy-ir.md
```diff
@@ -0,0 +1,84 @@
...
+1. It gets picked up by the `mesh-networking` pod, at which point:
+    - individual validation on the Traffic Policy in isolation occurs (eg standalone data validation, like no negative retry rates)
+        - if individual validation fails, write that to the object and stop
+        - otherwise, write a "validated" status onto the Traffic Policy
+    - a separate reconciler, watching only for Traffic Policies with a "validated" status, picks up that Traffic Policy
+        - this reconciler determines which mesh service(s) the validated traffic policy applies to, and looks up all
+        the `TrafficPolicyIntermediateRepresentation`s with the appropriate destinations.
+        - for each one of those `TrafficPolicyIntermediateRepresentation`s, it checks whether the validated Traffic Policy
+        can be merged with the other Traffic Policies already recorded on the Intermediate Representation. If the Traffic Policy existed
+        already and is already present in the `validated_traffic_policies` list, then consider the rest of the list together with its new state.
```
harveyxia, line 67: If the TrafficPolicy already exists, it seems important also know whether the trigger TP is older or newer than the TP that already exists in the `TrafficPolicyIntermediateRepresentation`.

If it's older than the one that already exists in the `TrafficPolicyIntermediateRepresentation`, we should ignore it, otherwise stale / requeued TP events may incorrectly apply stale state.

If it's equal to, then we can save on computation.

If it's newer than, then we proceed with processing.

design-docs/traffic-policy-ir.md
EItanya, line 29: What if instead of a new CRD we add this info to the MeshService Status. Essentially this data belongs to one anyway, and that may reduce the cognitive load as other parts of the system function in a similar way, writing to the MeshService as an intermediary.

GrahamGoudeau: Yeah I think I like this

design-docs/traffic-policy-ir.md
harveyxia: Discussed offline, assuming there's only a single controller instance handling TP's, this situation will never arise because for a given object, the read cache will never return an _older_ version of that same object. @ilackarms can you confirm my understanding?

harveyxia:
On the whole, the approach outlined here seems sound. Much of the complexity arises due to our selection API which allows for dynamically selecting targets by label/namespace. This means that a single TP instance does not map onto an edge of the network topology, or even a single node. Instead a TP instance maps onto a set of edges/nodes, which can change as sources/targets come into and out of existence.

That being said, there seems no straightforward way to avoid the many-to-many relationship between TP's and network edges/nodes. If a subset of these edges are conflicting with other TP's, it seems like a good user experience to have that conflict be limited to only the conflicting edges, and for the valid edges referenced by the TP to still be applied. With the proper tooling the debugging experience can be made manageable. We already have `meshctl describe` (built by the author of this proposal himself :) )

PR #666, opened 2020-05-07
Eks cluster discovery

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/665


pkg/clients/cluster_authorization.go
```diff
@@ -0,0 +1,29 @@
+package clients
+
+import (
+	k8s_core "github.com/solo-io/service-mesh-hub/pkg/api/kubernetes/core/v1"
+	"github.com/solo-io/service-mesh-hub/pkg/auth"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+type ClusterAuthClientFromConfigFactory func(remoteAuthConfig *rest.Config) (auth.ClusterAuthorization, error)
```
GrahamGoudeau, line 10: 🔥

pkg/clients/cluster_registration_client.go
```diff
@@ -0,0 +1,372 @@
...
+
+func (c *clusterRegistrationClient) Register(
+	ctx context.Context,
+	remoteConfig clientcmd.ClientConfig,
+	remoteClusterName string,
+	remoteWriteNamespace string,
+	overwrite bool,
+	useDevCsrAgentChart bool,
+	localClusterDomainOverride string,
+	remoteContextName string,
```
GrahamGoudeau, line 84: lots of repeated bools and strings here, may be another good contender for like a `RegisterOpts` struct, so we get names on these things

Edit: Yeah got to the end of the PR, method calls with bare primitives like `""` and `false` sketch me out- if this was a struct we could let those take default values without hurting the readability at those call sites

pkg/clients/cluster_registration_client.go
```diff
@@ -0,0 +1,372 @@
...
+	var remoteRestConfig *rest.Config
+	var serviceAccountBearerToken string
+	var secret *k8s_core_types.Secret
+	if remoteRestConfig, err = remoteConfig.ClientConfig(); err != nil {
+		return err
+	}
+	if err = c.checkClusterExistence(ctx, remoteClusterName, overwrite); err != nil {
+		return err
+	}
+	if err = c.ensureRemoteNamespace(ctx, remoteRestConfig, remoteWriteNamespace); err != nil {
```
GrahamGoudeau, line 97: There are a few stylistic things going on in this method that I feel decently strongly about. The three principles I'm thinking of here are:

* keep `if` conditional expressions side-effect free
    - https://www.teamten.com/lawrence/programming/keep-if-clauses-side-effect-free.html
* keep the scope of method-local variables as small as possible
    - https://softwareengineering.stackexchange.com/questions/113262/why-declare-variables-close-to-where-they-are-used
* reduce as much as possible the amount of time that values stay in a not-fully-initialized state
    - haven't been able to find a decent link for this one, but we have two nil-able references (and four values total) hanging out in an uninitialized state for a large chunk of this method, when we could just set them to non-nil, fully initialized values later on

Not only are the conditional expressions in this method setting variables used elsewhere in the function (not only in the block they're gating), but they have pretty profound business-logic side-effects. I'd much rather see conditionals used just for conditional branching, and side-effecting statements at a top-level in the method.

pkg/clients/cluster_registration_client.go
```diff
@@ -0,0 +1,372 @@
...
+	)
+	if err != nil {
+		if !k8s_errs.IsNotFound(err) {
+			return err
+		}
+	} else {
+		if !overwrite {
+			return ClusterAlreadyRegisteredError(remoteClusterName)
+		}
+	}
```
GrahamGoudeau, line 157: can these be collapsed?

```go
if k8s_errs.IsNotFound(err) {
  return nil
} else if err != nil {
  return err
} else if !overwrite {
  return ClusterAlreadyRegistered
}

return nil
```

similar to how `ensureRemoteNamespace` is written below

pkg/clients/cluster_registration_client.go
```diff
@@ -0,0 +1,372 @@
...
+		return nil, err
+	}
+	if len(config.Contexts) < 1 {
+		return nil, EmptyContextsError
+	}
+	var remoteContext *api.Context
+	if len(config.Contexts) == 1 {
+		for _, context := range config.Contexts {
+			remoteContext = context
+		}
```
GrahamGoudeau, line 249: I'm not sure we should do this- this could result in selecting a different context than what they had specified. I think we should always just do a map lookup based on the string that's passed in here, but also see my next comment...

pkg/clients/cluster_registration_client.go
```diff
@@ -0,0 +1,372 @@
...
+	} else {
+		remoteContext = config.Contexts[remoteContextName]
+	}
+	remoteCluster := config.Clusters[remoteContext.Cluster]
```
GrahamGoudeau, line 253: we need `_, ok` checks on these map accesses, we don't want to panic if someone mistypes a context name

pkg/clients/cluster_registration_client.go
```diff
@@ -0,0 +1,372 @@
...
+	remoteWriteNamespace string,
+	remoteConfig clientcmd.ClientConfig,
+	useDevCsrAgentChart bool,
+) error {
+	restConfig, err := remoteConfig.ClientConfig()
+	if err != nil {
+		return err
+	}
+	kubeClient := kubernetes.NewForConfigOrDie(restConfig)
+	helmInstaller := factories.NewHelmInstallerFactory(kubeClient.CoreV1().Namespaces(), &zaptest.Discarder{})
```
GrahamGoudeau, line 195: is this stuff unit testable? I don't recall seeing any `.CoreV1().Interface()`-style calls in our code before

pkg/clients/cluster_registration_client.go
```diff
@@ -0,0 +1,372 @@
...
+}
+
+func (c *clusterRegistrationClient) upsertSecretData(
+	ctx context.Context,
+	secret *k8s_core_types.Secret,
+) error {
+	existing, err := c.secretClient.GetSecret(ctx, client.ObjectKey{Name: secret.Name, Namespace: secret.Namespace})
+	if err != nil {
+		if k8s_errs.IsNotFound(err) {
+			return c.secretClient.CreateSecret(ctx, secret)
```
GrahamGoudeau, line 335: similar comment here, this error checking can be reduced an indentation level

services/mesh-discovery/pkg/compute-target/aws/api_handler.go
```diff
@@ -95,3 +89,14 @@ func (a *awsCredsHandler) buildContext(parentCtx context.Context, computeTargetN
 		zap.String(constants.COMPUTE_TARGET, computeTargetName),
 	))
 }
+
+func (a *awsCredsHandler) initReconcilers(
```
GrahamGoudeau, line 93: is this doing init'ing? Should it be `runReconcilers` or something?

services/mesh-discovery/pkg/compute-target/aws/api_handler.go
```diff
@@ -46,25 +45,20 @@ func (a *awsCredsHandler) ComputeTargetAdded(ctx context.Context, secret *k8s_co
...
 			select {
 			case <-ticker.C:
-				logger.Debugf("Reconciling AppMesh with secret %s.%s", secret.GetName(), secret.GetNamespace())
-				err := appMeshDiscoveryReconciler.Reconcile(reconcilerCtx)
-				if err != nil {
-					logger.Error(err)
-				}
+				a.initReconcilers(logger, reconcilerCtx, creds)
 			case <-reconcilerCtx.Done():
 				ticker.Stop()
```
GrahamGoudeau, line 63: realize this didn't change in this PR, but `ticker.Stop()` is a pretty classic candidate for a `defer` statement. Don't necessarily have to change right now, just thought I'd mention it

services/mesh-discovery/pkg/discovery/k8s-cluster/rest/eks/reconciler.go
```diff
@@ -0,0 +1,151 @@
...
+) error {
+	cluster, err := eksClient.DescribeCluster(ctx, awsClusterName)
+	if err != nil {
+		return err
+	}
+	config, err := e.eksConfigBuilderFactory(eksClient).ConfigForCluster(ctx, cluster)
+	if err != nil {
+		return err
+	}
+	return e.clusterRegistrationClient.Register(
```
GrahamGoudeau, line 140: yeah see my comment elsewhere about the zero-valued args here. But in general this is 💯 that this is happening from mesh-discovery now

cli/pkg/common/common_interfaces.go
```diff
@@ -59,6 +53,8 @@ type KubeClients struct {
 	AccessControlPolicyClient       zephyr_networking.AccessControlPolicyClient
 	ResourceDescriber               description.ResourceDescriber
 	ResourceSelector                selector.ResourceSelector
+	ClusterRegistrationClients      ClusterRegistrationClients
+	ClusterRegistrationClient       clients.ClusterRegistrationClient
```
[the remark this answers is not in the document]
harveyxia: we don't actually use the CsrAgentFactory directly anymore (it's now a dependency of the ClusterRegistrationClient), so it can be removed

pkg/clients/cluster_registration_client.go
harveyxia: to simplify further, rewrote as:

```
if err != nil && !k8s_errs.IsNotFound(err) {
		return err
	} else if !overwrite {
		return ClusterAlreadyRegisteredError(remoteClusterName)
	}
```

pkg/clients/cluster_registration_client.go
harveyxia: Fortunately it is unit testable, there's no initialization logic happening in these two lines. This is due to the fact that the Helm library in go-utils uses the `v1.NamespaceInterface`. We could add a namespace client factory to eliminate this coupling if you think it necessary

services/mesh-discovery/pkg/compute-target/aws/api_handler.go
harveyxia: renamed

pkg/clients/cluster_registration_client.go
harveyxia: Makes sense, I wrote it this way for the syntactic sugar of inling the `if err != nil { return err}` but keeping conditionals side effect free overrides the syntax considerations

PR #673, opened 2020-05-11
Cluster deregistration cleanup

harveyxia:
1. Move `KubeConfigLookup`, `KubeLoader`, and kube `Converter` to `pkg/kubeconfig` since these are used across the CLI and pods now
2. Consolidate CSR agent uninstaller logic into `CsrAgentInstaller.Uninstall`


pkg/clients/cluster-registration/interfaces.go
```diff
@@ -21,3 +22,9 @@ type ClusterRegistrationClient interface {
 		registerOpts ClusterRegisterOpts,
 	) error
 }
+
+// perform all the steps necessary to de-register this cluster from the SMH installation
+type ClusterDeregistrationClient interface {
+	// results in deleting thepkg/clients/cluster-registration/registration_client.go kubeCluster that's passed in
```
GrahamGoudeau, line 28: Bad find/replace?

PR #679, opened 2020-05-11
Meshworkload discovery pure reconcile

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/674


services/mesh-discovery/pkg/discovery/mesh-workload/k8s/mesh_workload_finder.go
```diff
@@ -61,352 +45,196 @@ type meshWorkloadFinder struct {
...
+			if err != nil {
+				return err
+			}
 		}
-
-		rediscoveredWorkloadsByName[discoveredWorkload.GetName()] = discoveredWorkload
 	}
-
-	// for each mesh that we have rediscovered, ensure that we can match it to a mesh that currently exists in the cluster
-	for _, meshWorkloadIter := range existingMeshWorkloads.Items {
```
harveyxia, line 155: The algorithm in the PR is the same as the approach here except it also creates MeshWorkloads that are newly discovered and hadn't existed before (which doesn't occur in here because we only loop through `existingMeshWorkloads`)

PR #681, opened 2020-05-12
enable multi-cluster subset routing

christian-posta:
Fixes https://github.com/solo-io/service-mesh-hub/issues/671
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/671


services/mesh-networking/pkg/routing/traffic-policy-translator/istio-translator/istio_translator.go
```diff
@@ -415,6 +418,60 @@ func (i *istioTrafficPolicyTranslator) translateSubset(
...
+			// we have the subset already
+			return  nil
+		}
+	}
+
+	defaultLabels := map[string]string {
+		"cluster": destClusterName,
+	}
+	destinationRule.Spec.Subsets = append(destinationRule.Spec.Subsets, &istio_networking_types.Subset{
+		Name:   subsetName,
```
harveyxia, line 463: I believe subset names on a given DestinationRule must be unique. This name should already have been applied from invoking `i.translateSubset(ctx, destination)` in L509.

services/mesh-networking/pkg/routing/traffic-policy-translator/istio-translator/istio_translator.go
```diff
@@ -62,6 +62,9 @@ var (
 	MultiClusterSubsetsNotSupported = func(dest *zephyr_networking_types.TrafficPolicySpec_MultiDestination_WeightedDestination) error {
 		return eris.Errorf("Multi cluster subsets are currently not supported, found one on destination: %+v", dest)
 	}
+	MultiClusterDestinationRuleNotFound = func(resource *zephyr_core_types.ResourceRef) error {
+		return eris.Errorf("Could not find multicluster destination rule for %v in namespace %v", resource.Name, resource.Namespace )
```
GrahamGoudeau, line 66: given that these are both strings, can we do `%s` for these instead? So that the compiler can check it

services/mesh-networking/pkg/routing/traffic-policy-translator/istio-translator/istio_translator.go
```diff
@@ -415,6 +418,60 @@ func (i *istioTrafficPolicyTranslator) translateSubset(
 	return subsetName, nil
 }
 
+// ensure that subsets declared in this TrafficPolicy are reflected in the relevant kube Service's DestinationRules
+// return name of Subset declared in DestinationRule
```
harveyxia, line 422:
```suggestion
// For the local DestinationRule representing the multicluster DNS name for a remote service, declare a new subset entry that references the corresponding endpoint on the corresponding ServiceEntry.
// Reference: https://istio.io/blog/2019/multicluster-version-routing/#create-a-service-entry-and-destination-rule-on-cluster1-for-the-remote-reviews-service
```

services/mesh-networking/pkg/routing/traffic-policy-translator/istio-translator/istio_translator.go
```diff
@@ -449,14 +506,18 @@ func (i *istioTrafficPolicyTranslator) translateDestinationRoutes(
...
 				if err != nil {
 					return nil, err
 				}
+
+				if isMulticluster {
+					err := i.translateSubsetForMulticluster(ctx, hostnameForKubeService, subsetName, meshService, destination)
+					if err != nil {
+						return nil, err
+					}
+				}
```
GrahamGoudeau, line 519: I'm missing some context here, so for background: If a traffic policy is multicluster, we want to do BOTH translation of the single-cluster subset stuff and translation for the multicluster subset stuff?

services/mesh-networking/pkg/routing/traffic-policy-translator/istio-translator/istio_translator.go
```diff
@@ -415,6 +418,60 @@ func (i *istioTrafficPolicyTranslator) translateSubset(
...
+
+	destinationRuleRef := &zephyr_core_types.ResourceRef{
+		Name:      serviceMulticlusterName,
+		Namespace: installationNamespace,
+	}
+	destinationRuleClient := i.destinationRuleClientFactory(dynamicClient)
+	destinationRule, err := destinationRuleClient.GetDestinationRule(ctx, clients.ResourceRefToObjectKey(destinationRuleRef))
+	if destinationRule == nil {
+		return MultiClusterDestinationRuleNotFound(destinationRuleRef)
+	}
```
GrahamGoudeau, line 450: You should actually be checking if the error is a NotFoundError here, ie `errors.IsNotFound`, imported from `"k8s.io/apimachinery/pkg/api/errors"`

services/mesh-networking/pkg/routing/traffic-policy-translator/istio-translator/istio_translator.go
```diff
@@ -415,6 +418,60 @@ func (i *istioTrafficPolicyTranslator) translateSubset(
...
+	destClusterName := destination.GetDestination().GetCluster()
+	mesh, err := i.getMeshForMeshService(ctx, meshService)
+	if err != nil {
+		return err
+	}
+
+	clusterName := mesh.Spec.GetCluster().GetName()
+	installationNamespace := mesh.Spec.GetIstio().GetInstallation().GetInstallationNamespace()
+
+	dynamicClient, err := i.dynamicClientGetter.GetClientForCluster(ctx, clusterName)
```
GrahamGoudeau, line 440: this error is getting dropped

services/mesh-networking/pkg/routing/traffic-policy-translator/istio-translator/istio_translator_test.go
```diff
@@ -886,13 +894,18 @@ var _ = Describe("IstioTranslator", func() {
...
-			destNamespace := "namespace"
-			multiClusterDnsName := "multicluster-dns-name"
-			destCluster := "remote-cluster-1"
+			destName := "remote-dest-name"
+			destNamespace := "remote-dest-namespace"
+			destCluster := "remote-cluster"
+			multiClusterDnsName := fmt.Sprintf("%v.%v.%v", destName, destNamespace, destCluster)
+			//meshServiceCluster := "mesh-svc-cluster"
 			declaredSubset := map[string]string{"env": "dev", "version": "v1"}
+			//clusterSubset := map[string]string{"cluster": "dev", }
```
GrahamGoudeau, line 905: can we remove this commented-out code?

services/mesh-networking/pkg/federation/resolver/meshes/istio/istio_federation_client.go
```diff
@@ -206,7 +206,9 @@ func (i *istioFederationClient) setUpServiceEntry(
 	endpoint := &alpha3.ServiceEntry_Endpoint{
 		Address: eap.Address,
 		Ports:   make(map[string]uint32),
+		Labels:  make(map[string]string),
 	}
+	endpoint.Labels["cluster"] = meshService.Spec.KubeService.Ref.Cluster
```
GrahamGoudeau, line 211: Do we need to put "cluster" in a constant somewhere?

services/mesh-networking/pkg/routing/traffic-policy-translator/istio-translator/istio_translator.go
```diff
@@ -415,6 +418,60 @@ func (i *istioTrafficPolicyTranslator) translateSubset(
...
+
+	destinationRuleRef := &zephyr_core_types.ResourceRef{
+		Name:      serviceMulticlusterName,
+		Namespace: installationNamespace,
+	}
+	destinationRuleClient := i.destinationRuleClientFactory(dynamicClient)
+	destinationRule, err := destinationRuleClient.GetDestinationRule(ctx, clients.ResourceRefToObjectKey(destinationRuleRef))
```
GrahamGoudeau, line 447: so given that this depends on the output of federation, which is run in an entirely different goroutine, I'm worried about a race here. Nothing guarantees that this destination rule we're interested in will be in the controller-runtime cache at this point. I'm not sure what the solution is yet, but I think we may need some different approach here.

services/mesh-networking/pkg/routing/traffic-policy-translator/istio-translator/istio_translator.go
harveyxia: Yup, we need 2 DestinationRules.

1. We need a DestinationRule on the client side, representing the remote service, which is [detailed here](https://istio.io/blog/2019/multicluster-version-routing/#create-a-service-entry-and-destination-rule-on-cluster1-for-the-remote-reviews-service)
2. We also need the DestinationRule on the server side that actually contains the subset labels, [detailed here](https://istio.io/blog/2019/multicluster-version-routing/#create-a-destination-rule-on-both-clusters-for-the-local-reviews-service)

services/mesh-networking/pkg/routing/traffic-policy-translator/istio-translator/istio_translator.go
EItanya: Agreed, originally when I did this I put the logic in the federation resolver, lemme try and dig up that PR because I think that's the better approach.

services/mesh-networking/pkg/routing/traffic-policy-translator/istio-translator/istio_translator.go
EItanya: As I mentioned above, I believe these subsets should be added by the federation resolver as that is the component responsible for it, rather than splitting the responsbility

christian-posta: So I do tend to agree, and that's where I was hoping to put them... but the assumptions already in the code was that the destination rule subsets were not already known until a traffic policy was created. And it's true; there is now way in SMH to specify subsets ahead of time at the moment (ie at federation), unless we just assume the user will do that. If the assumption is that the user will always specify the subsets, then we could easily just copy them over on federation resolution. Before this PR, however, the subsets were automatically created when the Traffic Policy was created, not on federation resolution.

services/mesh-networking/pkg/routing/traffic-policy-translator/istio-translator/istio_translator.go
christian-posta: See previous comment. Let me know when you pull up the previous PR.

services/mesh-networking/pkg/routing/traffic-policy-translator/istio-translator/istio_translator.go
EItanya: So harvery and I had a brief discussion on slack about this. Defining all of the possible subsets ends up being  `O(MxN)` which isn't too bad. Plus, this would keep the well defined separation we already have which I think provides a lot of value in terms of simplicity and resource ownership

PR #683, opened 2020-05-12
update the build guide, new kind make targets

christian-posta:


Makefile
```diff
@@ -211,3 +211,12 @@ endef
 docker-push: docker
 	$(foreach component,$(COMPONENTS),$(call docker_push,$(component)))
 	$(call docker_push,csr-agent)
+
+CLUSTER_NAME := $(if $(CLUSTER_NAME),$(CLUSTER_NAME), $(shell kind get clusters | grep management-plane))
```
GrahamGoudeau, line 215: Nice... would you also mind replacing the relevant section of setup-kind.sh with this? There's a part where we pipe the output of `make docker` through sed and then manually load all the images. Looks like that can be replaced with this

christian-posta: Yah! good call

christian-posta: Okay, added -- and added to the guide

christian-posta:
Ah crap, need to add the `make init` to the build guide... will do this before we go forward with the PR

PR #686, opened 2020-05-13
Aws discovery config

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/685


api/settings/v1alpha1/aws.proto
```diff
@@ -0,0 +1,49 @@
+syntax = "proto3";
+package settings.zephyr.solo.io;
+option go_package = "github.com/solo-io/service-mesh-hub/pkg/api/settings.zephyr.solo.io/v1alpha1/types";
+
+import "gogoproto/gogo.proto";
+option (gogoproto.equal_all) = true;
+
+message AwsSettings {
+    repeated AwsAccountSettings aws_account_settings = 1;
```
harveyxia, line 9: @GrahamGoudeau @ilackarms @yuval-k 

Wanted to get your thoughts once more on this structure. To help visualize the YAML, here are some examples:

```
apiVersion: settings.zephyr.solo.io/v1alpha1
kind: Settings
metadata:
  namespace: service-mesh-hub
  name: smh-global-settings
spec:
  awsSettings:
    awsAccountSettings:
    - accountId: "410461945957"
      discoverySettings:
        appmesh:
          resourceSelectors:
          - matcher:
              regions:
              - us-east-2
        eks:
          resourceSelectors:
          - matcher:
              regions:
              - us-east-2
```
```
// Discovers everything for the account
apiVersion: settings.zephyr.solo.io/v1alpha1
kind: Settings
metadata:
  namespace: service-mesh-hub
  name: smh-global-settings
spec:
  awsSettings:
    awsAccountSettings:
    - accountId: "410461945957"
      discoverySettings:
        {}
```

pkg/settings/aws_selector.go
```diff
@@ -0,0 +1,180 @@
...
+		default:
+			return nil, UnknownSelectorType(resourceSelector)
+		}
+	}
+	return resourceSelectorsByRegion, nil
+}
+
+func (a *awsSelector) IsDiscoverAll(discoverySettings *zephyr_settings_types.DiscoverySettings) bool {
+	discoverAll := &zephyr_settings_types.DiscoverySettings{}
+	return discoverySettings.Equal(discoverAll)
```
GrahamGoudeau, line 62: I think, for improved UX, we should treat `discoverySettings == nil` the same as a non-nil but empty object. Then possibly add a boolean `disabled` flag in the `DiscoverySettings` object that defaults to false if it's not present.

pkg/settings/aws_selector_test.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	zephyr_settings_types "github.com/solo-io/service-mesh-hub/pkg/api/settings.zephyr.solo.io/v1alpha1/types"
+	"github.com/solo-io/service-mesh-hub/pkg/settings"
+	mock_aws "github.com/solo-io/service-mesh-hub/services/mesh-discovery/pkg/compute-target/aws/parser/mocks"
+)
+
+var _ = Describe("Utils", func() {
```
GrahamGoudeau, line 12: Outdated test name?

services/mesh-discovery/pkg/discovery/k8s-cluster/rest/eks/reconciler.go
```diff
@@ -150,11 +196,27 @@ func (e *eksReconciler) registerCluster(
...
+
+func (e *eksReconciler) fetchSelectorsByRegion(
+	ctx context.Context,
+	accountID string,
+) (settings_utils.AwsSelectorsByRegion, error) {
+	awsSettings, err := e.settingsClient.GetAWSSettingsForAccount(ctx, accountID)
+	if err != nil {
+		return nil, err
+	}
+	// "discoverySettings: {}" or "eks: {}" indicates discovery for all regions.
```
GrahamGoudeau, line 216: similar to my other comment (here and in discovery/mesh/rest/appmesh), I think we may want the UX of nil objects here equivalent to non-nil and empty, with specific `disable` flags available for use

services/mesh-discovery/pkg/compute-target/aws/api_handler.go
```diff
@@ -5,36 +5,42 @@ import (
...
 	"github.com/solo-io/service-mesh-hub/services/common/constants"
+	"github.com/solo-io/service-mesh-hub/services/mesh-discovery/pkg/compute-target/aws/clients/sts"
 	"go.uber.org/zap"
 	k8s_core_types "k8s.io/api/core/v1"
 )
 
 const (
-	ReconcileIntervalSeconds = 1           // TODO make this configurable by user
-	Region                   = "us-east-2" // TODO remove hardcode and replace with configuration
+	ReconcileIntervalSeconds = 1 // TODO make this configurable by user
```
GrahamGoudeau, line 20: time to pull this out into the settings now?

services/mesh-discovery/pkg/compute-target/aws/api_handler.go
```diff
@@ -93,10 +103,30 @@ func (a *awsCredsHandler) buildContext(parentCtx context.Context, computeTargetN
...
-		if err := reconciler.Reconcile(reconcilerCtx, creds, Region); err != nil {
-			logger.Errorw(fmt.Sprintf("Error reconciling for %s", reconciler.GetName()), zap.Error(err))
+		if err := reconciler.Reconcile(reconcilerCtx, creds, accountID); err != nil {
+			logger.Errorw(fmt.Sprintf("Error during reconcile for %s", reconciler.GetName()), zap.Error(err))
+		}
+	}
+}
+
+func (a *awsCredsHandler) fetchAWSAccount(creds *credentials.Credentials) (string, error) {
+	accountID, ok := a.credsToAccount[creds]
```
GrahamGoudeau, line 117: is this map intended to be used as a cache? I don't see anything reading from it except once at `ComputeTargetAdded` time, and every time that runs it's going to be a guaranteed cache miss

services/mesh-discovery/pkg/compute-target/aws/clients/sts/interfaces.go
```diff
@@ -0,0 +1,9 @@
+package sts
+
+import "github.com/aws/aws-sdk-go/service/sts"
+
+//go:generate mockgen -source interfaces.go -destination ./mocks/interfaces.go
+
+type STSClient interface {
```
GrahamGoudeau, line 7: can we get some docs on this? Is this making a real aws API call?

services/mesh-discovery/pkg/discovery/mesh/rest/appmesh/reconciler.go
```diff
@@ -52,46 +61,68 @@ func (a *appMeshDiscoveryReconciler) GetName() string {
...
+			continue
+		}
+		var nextToken *string
+		input := &appmesh.ListMeshesInput{
+			Limit:     NumItemsPerRequest,
+			NextToken: nextToken,
 		}
-		for _, appMeshRef := range appMeshes.Meshes {
-			discoveredMesh, err := a.convertAppMesh(appMeshRef, region)
+		for {
```
GrahamGoudeau, line 83: still a little nervous about this- could we at least get a debug-level log statement every time the loop runs, so we can detect an infinite loop?

pkg/clients/settings/interfaces.go
```diff
@@ -0,0 +1,14 @@
...
+	"context"
+
+	zephyr_settings_types "github.com/solo-io/service-mesh-hub/pkg/api/settings.zephyr.solo.io/v1alpha1/types"
+)
+
+//go:generate mockgen -source ./interfaces.go -destination ./mocks/mock_interfaces.go
+
+// Convenience wrapper around fetching the global Settings object.
+type SettingsHelperClient interface {
+	GetAWSSettingsForAccount(ctx context.Context, accountId string) (*zephyr_settings_types.AwsAccountSettings, error)
```
GrahamGoudeau, line 13: should prob document that this can return `(nil, nil)`

install/helm/charts/custom-resource-definitions/crds/config.zephyr.solo.io_v1alpha1_crds.yaml
```diff
@@ -0,0 +1,25 @@
+# Code generated by skv2. DO NOT EDIT.
+
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    helm.sh/hook: crd-install
```
GrahamGoudeau, line 7: we should probably make this optional in skv2... we definitely don't need this. I'll ask Scott about it

GrahamGoudeau: https://github.com/solo-io/skv2/pull/39

api/settings/v1alpha1/aws.proto
```diff
@@ -0,0 +1,49 @@
+syntax = "proto3";
+package settings.zephyr.solo.io;
```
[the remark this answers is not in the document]
harveyxia: I don't have strong opinions either. Does a separate API group give us more flexibility with versioning all Settings related CRD's separately?

services/mesh-discovery/pkg/compute-target/aws/api_handler.go
harveyxia: One complication is what to do if the value for `ReconcileIntervalSeconds` changes in the config. I don't think we can change the interval of the goroutine ticker after we invoke the goroutine, so that would mean stopping and restarting it, which would require additional plumbing (and we would need to introduce an EventWatcher on the settings CRD). Given the scope, we should file this as a separate issue I think

harveyxia: https://github.com/solo-io/service-mesh-hub/issues/690

services/mesh-discovery/pkg/compute-target/aws/api_handler.go
harveyxia: ah good catch

services/mesh-discovery/pkg/compute-target/aws/clients/sts/interfaces.go
harveyxia: yup, it's making a real API call, adding a clarifying comment

services/mesh-discovery/pkg/discovery/mesh/rest/appmesh/reconciler.go
harveyxia: added one for the EKS reconciler as well

api/settings/v1alpha1/settings.proto
```diff
@@ -0,0 +1,16 @@
...
+
+import "service-mesh-hub/api/settings/v1alpha1/aws.proto";
+
+import "gogoproto/gogo.proto";
+option (gogoproto.equal_all) = true;
+
+// Top level SMH user configuration object.
+message SettingsSpec {
+    AwsSettings aws_settings = 1;
+}
```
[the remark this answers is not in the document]
harveyxia: Great suggestion, looks much cleaner now

api/settings/v1alpha1/settings.proto
```diff
@@ -0,0 +1,43 @@
...
+        string account_id = 1;
+        // For unspecified or null fields, discovery will not run for the corresponding AWS resource type.
+        DiscoverySelector appmesh_discovery = 2;
+        DiscoverySelector eks_discovery = 3;
+    }
+
+    repeated AwsAccount aws = 1;
+}
+
+message SettingsStatus {
```
GrahamGoudeau, line 42: and just a reminder, we should add an empty (or with default values) Settings object to the smh helm chart

api/settings/v1alpha1/settings.proto
```diff
@@ -0,0 +1,43 @@
...
+                // AWS resource ARN that directly references a resource.
+                string arn = 1;
+                Matcher matcher = 2;
+            }
+        }
+
+        // AWS account ID.
+        string account_id = 1;
+        // For unspecified or null fields, discovery will not run for the corresponding AWS resource type.
+        DiscoverySelector appmesh_discovery = 2;
```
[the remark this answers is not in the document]
harveyxia: renamed to `mesh_discovery`

install/helm/charts/custom-resource-definitions/crds/settings.zephyr.solo.io_v1alpha1_crds.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    helm.sh/hook: crd-install
+  labels:
+    app: service-mesh-hub
+    app.kubernetes.io/name: service-mesh-hub
+  name: settings.settings.zephyr.solo.io
```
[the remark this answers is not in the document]
harveyxia: moved to `core`

PR #687, opened 2020-05-13
initial stab at building subset for multicluster domain routing in th…

christian-posta:
…e federation resolver

a diff impl of https://github.com/solo-io/service-mesh-hub/issues/671

may be more favorable than https://github.com/solo-io/service-mesh-hub/pull/681

### DIFF
--- a/services/mesh-networking/pkg/federation/resolver/meshes/istio/istio_federation_client.go
+++ b/services/mesh-networking/pkg/federation/resolver/meshes/istio/istio_federation_client.go
@@ -43,6 +43,7 @@ const (
 
 	EnvoySniClusterFilterName        = "envoy.filters.network.sni_cluster"
 	EnvoyTcpClusterRewriteFilterName = "envoy.filters.network.tcp_cluster_rewrite"
+	MultiClusterDRServiceEntryLabel  = "cluster"
 )
 
 type IstioFederationClient meshes.MeshFederationClient
@@ -152,6 +153,7 @@ func (i *istioFederationClient) FederateClientSide(
 	return i.setUpDestinationRule(
 		ctx,
 		clientForWorkloadMesh,
+		meshService,
 		serviceMulticlusterName,
 		installNamespace,
 	)
@@ -160,6 +162,7 @@ func (i *istioFederationClient) FederateClientSide(
 func (i *istioFederationClient) setUpDestinationRule(
 	ctx context.Context,
 	clientForWorkloadMesh client.Client,
+	meshService *zephyr_discovery.MeshService,
 	serviceMulticlusterName string,
 	installNamespace string,
 ) error {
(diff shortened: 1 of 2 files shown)


christian-posta:
@EItanya @harveyxia @GrahamGoudeau 
So this approach uses the existing subset information in the MeshService to build the subset info for the DestinationRule. I like this approach much better, however, there is one drawback. 

Although this does create the subset correctly for the DR, that subset name is not known to the Istio translator. When it creates the VirtualService in the multi-cluster scenario that routes to the correct subsets, we will have to query the DR to see what subsets it has and what labels match what is specified in the TrafficPolicy (because TrafficPolicy sees subsets not by name by a group of labels). So we end up in the same kind of race (was the service correctly federated before the traffic policy is is processed) in this scenario also.

EItanya:
> @EItanya @harveyxia @GrahamGoudeau
> So this approach uses the existing subset information in the MeshService to build the subset info for the DestinationRule. I like this approach much better, however, there is one drawback.
> 
> Although this does create the subset correctly for the DR, that subset name is not known to the Istio translator. When it creates the VirtualService in the multi-cluster scenario that routes to the correct subsets, we will have to query the DR to see what subsets it has and what labels match what is specified in the TrafficPolicy (because TrafficPolicy sees subsets not by name by a group of labels). So we end up in the same kind of race (was the service correctly federated before the traffic policy is is processed) in this scenario also.

So a solution to this is a function which names the subset based on the underlying labels, so that it would be the same for both `DestinationRules`. See [generateUniqueSubsetName](https://github.com/solo-io/service-mesh-hub/blob/89efdab66fc3c0218263ed777cf0fb7e65b07344/services/mesh-networking/pkg/routing/traffic-policy-translator/istio-translator/istio_translator.go#L662) in the istio translator. This same naming scheme should be used in the resolver.

christian-posta:
So that won't work as is, and doesn't solve the issue of the resource racing.

So the way MeshService determines subsets is by finding individual labels that have more than one value across the workloads and assuming that is a subset. So a subset by that definition is defined with a single label. We treat subsets in TrafficPolicy as being any arbitrary labels. We could change how subsets get built in MeshService but then we're back in the same problem of "what constitutes a subset"? An MxN projection? As discussed elsewhere proactive creating that is not desirable.

christian-posta:
After our call this morning, I think the approach that would work would be:

Let's get the TrafficPolicy re-write in [1] where we update the status of MeshService with the TPs that apply to that MeshService [2]. From there, the Federation client side resolver can use that information, see what subsets are being used, and then create the subsets for the multi-cluster destination rule.

This approach solves the race problem as well as having to compute MxN dest subsets proactively. 

[1]: https://github.com/solo-io/service-mesh-hub/pull/688
[2]: https://github.com/solo-io/service-mesh-hub/blob/smh-585-traffic-policy-story/api/discovery/v1alpha1/mesh_service.proto#L76

PR #688, opened 2020-05-13
Two out of three stages of traffic policy translation rewrite

GrahamGoudeau:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/585


services/mesh-networking/pkg/traffic-policy-temp/aggregation/reconciler.go
```diff
@@ -0,0 +1,359 @@
...
+
+		newlyValidatedPolicyState, validatedPolicyFound := uniqueStringToNewlyValidatedTrafficPolicy[identifierString]
+
+		// if the policy couldn't be validated or if it hasn't changed, keep the old state in the list
+		if !validatedPolicyFound || newlyValidatedPolicyState.Spec.Equal(previouslyRecordedPolicy.TrafficPolicySpec) {
+			policiesToRecordOnService = append(policiesToRecordOnService, previouslyRecordedPolicy)
+		} else {
+			// otherwise, there is a newly-valid state to record. Ensure that this new state can in fact be merged in with the rest of the list
+			var policiesToMergeWith []*zephyr_networking_types.TrafficPolicySpec
+			for i, tpIter := range previouslyRecordedPolicies {
```
harveyxia, line 160: Hmm, we're checking mergeability with all previously recorded policies including those whose statuses might have changed. If a previous TP's status changed to an error state, shouldn't it be excluded from the mergeability check? If a formerly valid TP is now erroneous, that may "free up" other TP's to be merged in

services/mesh-networking/pkg/traffic-policy-temp/aggregation/reconciler.go
```diff
@@ -0,0 +1,359 @@
...
+type aggregationReconciler struct {
+	trafficPolicyClient   zephyr_networking.TrafficPolicyClient
+	meshServiceClient     zephyr_discovery.MeshServiceClient
+	meshClient            zephyr_discovery.MeshClient
+	aggregator            Aggregator
+	translationValidators map[zephyr_core_types.MeshType]mesh_translation.TranslationValidator
+}
+
+func (a *aggregationReconciler) Reconcile(ctx context.Context) error {
+	allTrafficPolicies, err := a.trafficPolicyClient.ListTrafficPolicy(ctx)
```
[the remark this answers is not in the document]
harveyxia: After having read `determineNewStatuses`, I think we _should_ impose an ordering because the order in which we process the TP's to determine mergeability will affect which TP's are considered valid and which are in conflict, and we would want that to be deterministic

services/mesh-networking/pkg/traffic-policy-temp/aggregation/reconciler.go
```diff
@@ -0,0 +1,359 @@
...
+		err := a.updatePolicyStatusIfNecessary(ctx, policy, trafficPolicyToAllConflicts[policy], trafficPolicyToAllTranslationErrs[policy])
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (a *aggregationReconciler) determineNewStatuses(
```
harveyxia, line 119: My understanding of the current algorithm is as follows:

1. For all existing TP's `AllTPs`:
  a. Ignore if their state or spec has not changed
  b. Check mergeability with `AllTPs` and record any merge conflicts
  c. Check translatability with any existing TP's that have passed steps a, b, and c.
2. For all TP's that have not previously been processed in step 1, repeat steps a, b, and c

I'm not sure if step b is correct—shouldn't we be checking mergeability with only the TP's that we know are mergeable? To illustrate, if `previouslyRecordedPolicyA` has a merge conflict, a subsequent `previouslyRecordedPolicyB` will still be checking mergeability against a list containing `previouslyRecordedPolicyA`.

I wonder if the following approach could help us simplify the algorithm and make it easier to reason about:

1. Union all existing TP's and new TP's, `UnionTPs`
2. For each TP in `UnionTPs`:
  a. Ignore if it is already on the MeshService and both its state and spec are equivalent
  b. Check mergeability by attempting to merge them one by one into an accumulating list (exclude the ones that can't be merged and record the merge error).
  c. Check translatability on the mergeable output of step b

services/mesh-networking/pkg/traffic-policy-temp/aggregation/reconciler.go
```diff
@@ -0,0 +1,359 @@
...
+	ctx context.Context,
+	meshService *zephyr_discovery.MeshService,
+	newlyComputedMergeablePolicies []*zephyr_discovery_types.MeshServiceStatus_ValidatedTrafficPolicy,
+) error {
+	shouldUpdateStatus := false
+	if len(newlyComputedMergeablePolicies) != len(meshService.Status.ValidatedTrafficPolicies) {
+		shouldUpdateStatus = true
+	} else {
+		for i, newlyComputedPolicy := range newlyComputedMergeablePolicies {
+			shouldUpdateStatus = shouldUpdateStatus || !meshService.Status.ValidatedTrafficPolicies[i].Equal(newlyComputedPolicy)
```
harveyxia, line 307: How do we know that the indices of `meshService.Status.ValidatedTrafficPolicies` matches that of `newlyComputedMergeablePolicies`? Same concern applies to `updatePolicyStatusIfNecessary()` below. Perhaps we should transform one list into a map that the other list can key into to guarantee that we're comparing the relevant TP's

services/mesh-networking/pkg/traffic-policy-temp/validation/reconciler.go
```diff
@@ -0,0 +1,71 @@
...
+			logger.Debugf("Traffic policy %s.%s passed validation", trafficPolicy.GetName(), trafficPolicy.GetNamespace())
+		} else {
+			logger.Infof("Traffic policy %s.%s failed validation for reason: %+v", trafficPolicy.GetName(), trafficPolicy.GetNamespace(), validationErr)
+		}
+
+		if !trafficPolicy.Status.GetValidationStatus().Equal(newValidationStatus) {
+			trafficPolicy.Status.ValidationStatus = newValidationStatus
+
+			// also zero-out the conflict errors, since the state has changed and we don't know what it may conflict with now
+			trafficPolicy.Status.ConflictErrors = nil
```
harveyxia, line 61: Do zero-ing out `ConflictErrors` guarantee another round of conflict checking?

services/mesh-networking/pkg/traffic-policy-temp/aggregation/reconciler.go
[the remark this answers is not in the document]
GrahamGoudeau: I think, after offline conversation, we settled on that we *don't* need a deterministic ordering on the traffic policies returned by `ListTrafficPolicy`, because the ordering that we impose determines on the policies *as they are recorded on the service*. So we actually don't care what ordering comes out of this client.

services/mesh-networking/pkg/traffic-policy-temp/aggregation/aggregator.go
```diff
@@ -0,0 +1,265 @@
...
+// return a list of pairs:
+//    - the service (note that it will have the previously-recorded traffic policy state in its status)
+//           (services that have no traffic policies applying to them *will* be reflected in this list- their ServiceWithRelevantPolicies struct will have an empty `TrafficPolicies` field)
+//    - the traffic policies in the given snapshot that are associated with the above service.
+//           (This list must be reconciled with the existing state in the service's status)
+func (a *aggregator) GroupByMeshService(
+	trafficPolicies []*zephyr_networking.TrafficPolicy,
+	meshServices []*zephyr_discovery.MeshService,
+) (result []*ServiceWithRelevantPolicies, err error) {
+	for _, serviceIter := range meshServices {
```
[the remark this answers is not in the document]
harveyxia: I misunderstood the structure of the loops—the outer loop actually minimizes the search space required of the inner loop. Disregard my above comment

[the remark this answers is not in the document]
GrahamGoudeau: Talked offline, not sure it's possible to avoid quadratic here based on the implementation of the resource selector. Can revisit in the future when doing benchmarks

services/mesh-networking/pkg/traffic-policy-temp/aggregation/aggregator.go
```diff
@@ -0,0 +1,265 @@
...
+		return nil, err
+	}
+	return &mergeableHttpTrafficPolicy{
+		HttpMatcher:       httpMatcher,
+		SourceSelector:    trafficPolicySpec.SourceSelector,
+		TrafficPolicySpec: newTPSpec,
+	}, nil
+}
+
+func (a *aggregator) mergeTrafficPolicySpec(
```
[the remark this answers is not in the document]
GrahamGoudeau: Yeah totally. We'll have to think about this. I'm not sure that we'll want to do reflection here in such a tight loop, but I'm not sure what the other options are.

services/mesh-networking/pkg/traffic-policy-temp/validation/reconciler.go
GrahamGoudeau: Removed this, good call

services/mesh-networking/pkg/traffic-policy-temp/aggregation/reconciler.go
GrahamGoudeau: Talked offline: We're willing to accept that a reconcile iteration may erroneously produce a conflict state on a traffic policy temporarily. That state should only last until the next reconciliation iteration occurs, since the new state that we hadn't seen yet will be accounted for in the cache.

services/mesh-networking/pkg/traffic-policy-temp/aggregation/aggregator.go
```diff
@@ -51,78 +53,65 @@ func (a *aggregator) FindMergeConflict(
...
-		result = append(result, resultEntry)
+	return results, err
+}
+
+/*
+	Only attempt to add the HttpMatcher the list if it is non-nil.
+	Nil HttpRequestMatchers is a valid state in the Istio API, and therefore must
+	be treated as an additional case here
+*/
+func (*aggregator) appendHttpMatcherIfNonNil(
```
harveyxia, line 106: Nice

services/mesh-networking/pkg/traffic-policy-temp/aggregation/interfaces.go
```diff
@@ -18,24 +61,9 @@ type Aggregator interface {
 		meshService *zephyr_discovery.MeshService,
 	) *zephyr_networking_types.TrafficPolicyStatus_ConflictError
 
-	// return a list of pairs:
-	//    - the service (note that it will have the previously-recorded traffic policy state in its status)
-	//           (services that have no traffic policies applying to them *will* be reflected in this list- their ServiceWithRelevantPolicies struct will have an empty `TrafficPolicies` field)
-	//    - the traffic policies in the given snapshot that are associated with the above service.
-	//           (This list must be reconciled with the existing state in the service's status)
-	GroupByMeshService(
+	// return the policies tht have the given mesh service as a destination
```
harveyxia, line 64:
```suggestion
	// return the policies that have the given mesh service as a destination
```

services/mesh-networking/pkg/traffic-policy-temp/aggregation/interfaces.go
```diff
@@ -1,14 +1,57 @@
...
+	PolicyToConflictErrors   map[*zephyr_networking.TrafficPolicy][]*zephyr_networking_types.TrafficPolicyStatus_ConflictError
+	PolicyToTranslatorErrors map[*zephyr_networking.TrafficPolicy][]*zephyr_networking_types.TrafficPolicyStatus_TranslatorError
+}
+
+// This interface serves to abstract away the details of merging new entries into various objects' statuses. Its behavior needs to
+// ensure that we don't incur k8s api server calls when they are not necessary.
+// The logic depends on there being an idempotent order coming out of `PolicyCollector` if no updates have happened.
+// These methods can change the objects they are handed- they return true if the object was changed in memory and should be updated in the persistence layer
+type InMemoryStatusUpdater interface {
+	UpdateServicePolicies(
```
harveyxia, line 44: I'll leave this up to you—I wonder if we should use the term "mutate" instead of "update" to further emphasize that this method is just updating objects in memory and not issuing k8s updates.

PR #695, opened 2020-05-18
Give mesh-discovery permissions for Settings CRD. Continue scanning Appmesh pods if any error out.

harveyxia:
…ppmesh pods if any error out.

### DIFF
--- a/install/helm/charts/mesh-discovery/templates/clusterrole.yaml
+++ b/install/helm/charts/mesh-discovery/templates/clusterrole.yaml
@@ -17,4 +17,4 @@ rules:
     verbs: ["*"]
   - apiGroups: ["core.zephyr.solo.io"]
     resources: ["settings"]
-    verbs: ["*"]
+    verbs: ["get", "list", "watch"]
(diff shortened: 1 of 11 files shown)


install/helm/charts/mesh-discovery/templates/clusterrole.yaml
```diff
@@ -15,3 +15,6 @@ rules:
   - apiGroups: ["discovery.zephyr.solo.io"]
     resources: ["meshes", "meshworkloads", "meshservices", "kubernetesclusters"]
     verbs: ["*"]
+  - apiGroups: ["core.zephyr.solo.io"]
+    resources: ["settings"]
+    verbs: ["*"]
```
GrahamGoudeau, line 20: Do we need `*`on this? Can start with `get/list/watch`?

harveyxia: Good point, I wasn't sure because we've been using "*" for other SMH CRDs but I suppose this CRD is user-facing so we should be more careful with permissioning

PR #699, opened 2020-05-18
Aws discovery account

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/696


services/mesh-discovery/pkg/compute-target/aws/parser/interfaces.go
```diff
@@ -14,9 +14,11 @@ type AppMeshPod struct {
 }
 
 type AppMeshScanner interface {
-	ScanPodForAppMesh(pod *k8s_core_types.Pod) (*AppMeshPod, error)
+	ScanPodForAppMesh(
+		pod *k8s_core_types.Pod,
+		configMap *k8s_core_types.ConfigMap,
```
GrahamGoudeau, line 19: tl;dr, I think the configmap lookup and the method `parseAwsAccountFromConfigMap` should be moved into their own component

I think this interface is a little too complicated now- in order to use it, you have to 1. know what ObjectKey global var to use, 2. do a configmap lookup yourself, then 3. pass that nil-able unstructured config into this method. Then this method implementation becomes way more smart, it has to know how to parse that unstructured config.

I think we can encapsulate these different concerns a little tighter- if we had something like a
```go
func NewAwsAccountIdLookup(configMapClientFactory ConfigMapClientFactory)...

type AwsAccountIdLookup interface {
  // returns ("", nil) if no cluster-level account ID is set in any configmap
  GetClusterLevelAccountId(clusterScopedClient client.Client) (accountId string, err error)
}
```

then 
1. the lookup and parsing can all be contained inside that implementation, and
2. this method on `AppMeshScanner` can look like
```go
type AppMeshScanner interface {
  ScanPodForAppMesh(
    pod *k8s_core_types.Pod,
    clusterLevelAccountId string, // may be empty if not configured
  )
}
```

Also, def not strictly necessary for this PR, but if we wanted to lean on the type system even more to help inform/document how to use these interfaces, we could create a type alias for Aws account IDs. Then our two interfaces are linked by a common type, one producing it and one consuming it:
```go
type AwsAccountId string

type AwsAccountIdLookup interface {
  // returns ("", nil) if no cluster-level account ID is set in any configmap
  GetClusterLevelAccountId(clusterScopedClient client.Client) (AwsAccountId, error)
}
type AppMeshScanner interface {
  ScanPodForAppMesh(
    pod *k8s_core_types.Pod,
    clusterLevelAccountId AwsAccountId, // may be empty if not configured
  )
}
```

GrahamGoudeau: the rest of the PR lgtm

services/mesh-discovery/pkg/compute-target/aws/parser/appmesh_scanner.go
```diff
@@ -27,19 +30,34 @@ var (
...
+	ctx context.Context,
+	pod *k8s_core_types.Pod,
+	remoteClient client.Client,
+) (*AppMeshPod, error) {
 	var awsAccountID, region, appMeshName, virtualNodeName string
+	var err error
+	awsAccountID, err = a.awsAccountIdFetcher.GetEksAccountId(ctx, remoteClient)
+	if err != nil {
+		contextutils.LoggerFrom(ctx).Warnf("Error parsing aws-auth.kube-system ConfigMap: %+v")
+	}
```
GrahamGoudeau, line 60: big +1 on the fetching/parsing logic being wrapped up in that new interface

however, I do think that this pod scanner component should still be pure- there are a couple advantages I can see

1. this class should be agnostic of where the account ID is coming from (even more agnostic than hiding it behind an interface), which I think means receiving the fully-parsed string form of the account ID through the method args 
2. handing this component a `Client` is a pretty heavy hammer for the relatively small nail of parsing out image names and environment variables. Turning that back into a string would hugely simplify this logic and its tests (not that test complexity is what we're optimizing for, but it's generally an indicator of how complex the class is to use for real clients)

I've been thinking that the more we can abstract away our client interactions from our processing code, the better- and I think moving this `awsAccountIdFetcher` into the places that are already doing client-y things feels pretty reasonable here

services/mesh-discovery/pkg/compute-target/aws/parser/interfaces.go
```diff
@@ -13,11 +16,23 @@ type AppMeshPod struct {
...
+	) (*AppMeshPod, error)
 }
 
 type ArnParser interface {
 	ParseAccountID(arn string) (string, error)
 	ParseRegion(arn string) (string, error)
 }
+
+type AwsAccountIdFetcher interface {
+	GetEksAccountId(
```
GrahamGoudeau, line 34: some docs on this would be good, probably with a link to some amazon docs that mention the configmap or something

PR #707, opened 2020-05-19
Fix hyphens on helm install command

buzzsurfr:
Helm command had the wrong hyphen/dash and was giving error: 
```
Error: expected at most two arguments, unexpected arguments: —-namespace, service-mesh-hub
```

Traced to dash in place of hyphen, shown with hexdump (top was copied from docs, bottom is me inputting two hyphens)

```bash
❯ echo —- | hexdump
0000000 e2 80 94 2d 0a
0000005
❯ echo -- | hexdump
0000000 2d 2d 0a
0000003
```

### DIFF
--- a/docs/content/setup/_index.md
+++ b/docs/content/setup/_index.md
@@ -123,7 +123,7 @@ Note that the location of the Service Mesh Hub Helm charts is subject to change.
 Then install Service Mesh Hub into the `service-mesh-hub` namespace:
 
 ```shell
-helm install smh smh/service-mesh-hub —-namespace service-mesh-hub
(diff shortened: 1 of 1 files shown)


GrahamGoudeau:
Thanks for catching @buzzsurfr!

christian-posta:
Wow.. nice catch indeed! I copy/pasted that command from my notes in Evernote when I wrote that doc. It must have snuck in. Sorry for that @buzzsurfr

buzzsurfr:
No worries Christian! Looks like the build failed, but that may be the [quay outage](https://status.quay.io/incidents/kw2627bsdwd9)

GrahamGoudeau:
🚂 gonna force it through

PR #710, opened 2020-05-20
Meshctl appmesh eks demo cmd

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/708


cli/pkg/tree/demo/cleanup/appmesh_eks_cleanup.go
```diff
@@ -0,0 +1,35 @@
...
+
+const (
+	appmeshEksCleanupScript = `
+region=%s
+meshName=%s
+eksClusterName=%s
+awsAccountID=$(echo $(aws sts get-caller-identity --query 'Account'))
+
+if [ -z ${awsAccountID+x} ]; 
+then echo "Unable to fetch AWS account ID, check that your AWS credentials are configured properly." && exit 0 ;
```
GrahamGoudeau, line 21: should be exiting nonzero here

cli/pkg/tree/demo/cleanup/appmesh_eks_cleanup.go
```diff
@@ -0,0 +1,35 @@
...
+eksClusterName=%s
+awsAccountID=$(echo $(aws sts get-caller-identity --query 'Account'))
+
+if [ -z ${awsAccountID+x} ]; 
+then echo "Unable to fetch AWS account ID, check that your AWS credentials are configured properly." && exit 0 ; 
+else echo "Using AWS Account ID $awsAccountID" ; 
+fi
+
+# Delete Appmesh mesh
+aws appmesh delete-mesh --mesh-name=$meshName | cat
```
GrahamGoudeau, line 26: Some kind of comment here about how this is forcing non-interactive would be good, I think. It looks surprising as-is

cli/pkg/tree/demo/cleanup/appmesh_eks_cleanup.go
```diff
@@ -0,0 +1,35 @@
+package cleanup
+
+import (
+	"fmt"
+
+	"github.com/solo-io/service-mesh-hub/cli/pkg/common/exec"
+)
+
+func AppmeshEksCleanup(runner exec.Runner, region string, meshName string, eksClusterName string) error {
+	return runner.Run("bash", "-x", "-c", fmt.Sprintf(appmeshEksCleanupScript, region, meshName, eksClusterName))
```
GrahamGoudeau, line 10: should we be running with `-e` too?

cli/pkg/tree/demo/cleanup/istio_multicluster_cleanup.go
```diff
@@ -1,17 +1,15 @@
...
 import (
-	"context"
-
 	"github.com/solo-io/service-mesh-hub/cli/pkg/common/exec"
 )
 
-func DemoCleanup(ctx context.Context, runner exec.Runner) error {
-	return runner.Run("bash", "-c", cleanupDemoScript)
+func IstioMulticlusterCleanup(runner exec.Runner) error {
+	return runner.Run("bash", "-c", istioMulticlusterCleanupScript)
```
GrahamGoudeau, line 8: same comment, I think we should use -e. And that makes me wonder if we shouldn't abstract away the details of the bash invocations from each one of these. Seems reasonable that each one will have the pattern `bash -x -e -u -c ARG`

cli/pkg/tree/demo/init/appmesh_eks_demo.go
```diff
@@ -0,0 +1,121 @@
...
+*/
+const (
+	appmeshEksDemoScript = `
+awsAccountID=$(echo $(aws sts get-caller-identity --query 'Account'))
+region=%s
+clusterName=smh-demo-cluster-$(xxd -l8 -ps /dev/urandom)
+meshName=smh-demo-mesh-$(xxd -l8 -ps /dev/urandom)
+
+if [ -z ${awsAccountID+x} ]; 
+then echo "Unable to fetch AWS account ID, check that your AWS credentials are configured properly." && exit 0 ;
```
GrahamGoudeau, line 29: should be nonzero

cli/pkg/tree/demo/init/appmesh_eks_demo.go
```diff
@@ -0,0 +1,121 @@
...
+kubectl -n service-mesh-hub apply -f - <<EOF
+apiVersion: v1
+kind: Secret
+type: solo.io/register/aws-credentials
+metadata:
+  name: smh-demo-aws
+  namespace: service-mesh-hub
+data:
+  aws_access_key_id: $aws_access_key_id
+  aws_secret_access_key: $aws_secret_access_key
```
GrahamGoudeau, line 93: will these be dumped to the console because of -x? Same comment with the cert data up above

services/mesh-discovery/pkg/compute-target/aws/api_handler.go
```diff
@@ -112,7 +112,7 @@ func (a *awsCredsHandler) runReconcilers(
 }
 
 func (a *awsCredsHandler) fetchAWSAccount(creds *credentials.Credentials) (string, error) {
-	// Region does not matter for constructing the STS client because the account identity is region agnostic.
+	// AwsRegion does not matter for constructing the STS client because the account identity is region agnostic.
```
GrahamGoudeau, line 115: bad find/replace?

cli/pkg/tree/demo/cleanup/appmesh_eks_cleanup.go
harveyxia: Exiting non-zero causes a verbose `meshctl` error message to be outputted, and I don't expect this cmd to be used in an automated fashion, so I figured it'd be better to hide the verbosity

cli/pkg/tree/demo/init/appmesh_eks_demo.go
harveyxia: hiding

cli/pkg/options/flags.go
```diff
@@ -114,25 +114,25 @@ func AddMeshInstallFlags(cmd *cobra.Command, opts *Options) {
...
 }
 
-func AddMulticlusterDemoFlags(cmd *cobra.Command, opts *Options) {
+func AddDemoCleanupFlags(cmd *cobra.Command, opts *Options, profileNames []string) {
 	flags := cmd.PersistentFlags()
 
-	flags.StringVar(&opts.Demo.IstioMulticluster.RemoteContextName, "remote-context", "", "In multicluster demos where --use-kind is not specified, use this flag to specify what context to use from your kubeconfig for the remote cluster")
-	flags.StringVar(&opts.Demo.IstioMulticluster.RemoteClusterName, "remote-cluster-name", "", "Name under which to register the second cluster")
+	profileFlag := "profile"
+	flags.StringVar(&opts.DemoCleanup.Profile, profileFlag, "", fmt.Sprintf("Specify the demo profile, must be one of [%s]", strings.Join(profileNames, ", ")))
```
GrahamGoudeau, line 130: so, each profile may depend on a different set of binaries being available on the path. We ideally would be able to communicate to the user which binaries are required apart from running the command and seeing where it fails. I'm thinking that if we implement these as cobra subcommands, rather than a flat list of strings, we can document the prerequisites of each one in a better way, ie on the cobra command --help usage.

cli/pkg/tree/demo/cleanup/appmesh_eks_cleanup.go
harveyxia: making non-zero and hiding the error message

PR #713, opened 2020-05-21
consul discovery design doc

harveyxia:


harveyxia:
We've tentatively decided to simply use the datacenter name as the membership group identifier because we expect it to be less subject to change than the IP addresses of the consul servers. We'll make this assumption evident in our user docs.

PR #714, opened 2020-05-21
failover doc

yuval-k:


design-docs/global-services.md
```diff
@@ -0,0 +1,165 @@
...
+  hosts:
+  - reviews-global.default.global
+  location: MESH_INTERNAL
+  ports:
+  - name: http1
+    number: 9080
+    protocol: http
+  resolution: STATIC
+  addresses:
+  - 240.0.0.20
```
harveyxia, line 67: How is this static IP address determined?

design-docs/global-services.md
```diff
@@ -0,0 +1,165 @@
...
+        connect_timeout: 1s
+        lb_policy: CLUSTER_PROVIDED
+        cluster_type:
+          name: envoy.clusters.aggregate
+          typed_config:
+            "@type": type.googleapis.com/udpa.type.v1.TypedStruct
+            type_url: type.googleapis.com/envoy.config.cluster.aggregate.v2alpha.ClusterConfig
+            value:
+              clusters:
+              - outbound|9080||reviews.default.svc.cluster.local
```
harveyxia, line 103: Just to confirm, is this list ordered by decreasing priority?

design-docs/global-services.md
```diff
@@ -0,0 +1,165 @@
...
+      interval: 5s
+  trafficPolicy:
+    tls:
+      mode: ISTIO_MUTUAL
+```
+
+# Remote cluster
+On the remote cluster we will configure these objects, to redirect incoming traffic to the correct local service (pending final testing):
+```
+apiVersion: networking.istio.io/v1beta1
```
harveyxia, line 129: Could we also use an `EnvoyFilter` to perform this redirection? That's how [service federation](https://github.com/solo-io/service-mesh-hub/blob/master/services/mesh-networking/pkg/federation/resolver/meshes/istio/istio_federation_client.go#L441) is currently working.

What are the pros/cons, if any, to using a VirtualService vs. EnvoyFilter for this?

yuval-k: Better to use vs if you can. EnvoyFilter is a break glass configuration that should be avoided if possible

design-docs/global-services.md
yuval-k: Correct

design-docs/global-services.md
yuval-k: It is decided. 
It just needs to be unique.  It doesn't need to be a real ip address

yuval-k: In practice will probably scan existing sevice entries to generate a unique one

EItanya: We already have a component for this that @GrahamGoudeau designed
https://github.com/solo-io/service-mesh-hub/blob/master/services/mesh-networking/pkg/federation/dns/ip_assigner.go

design-docs/global-services.md
```diff
@@ -0,0 +1,165 @@
...
+spec:
+  selector:
+    istio: ingressgateway # use istio default ingress gateway
+  servers:
+  - port:
+      number: 80
+      name: http
+      protocol: HTTP
+    hosts:
+    - reviews-global.default.global
```
christian-posta, line 164: Do we need this `Gateway` resource? Could we just match on all *.global services?

design-docs/global-services.md
```diff
@@ -0,0 +1,165 @@
...
+we need it so `reviews-global.default.global` is resolvable.
+
+```
+apiVersion: networking.istio.io/v1alpha3
+kind: ServiceEntry
+metadata:
+  name: reviews-global
+spec:
+  hosts:
+  - reviews-global.default.global
```
christian-posta, line 59: Do we need to specify the namespace in the name? For global services, the namespace may not be consistent across clusters, so doesn't really mean much. Unless we're trying to create a global namespace for naming conventions that's not tied to Kubernetes namespaces?

PR #717, opened 2020-05-21
expose --values option for csr-agent Helm installation for `meshctl cluster register`

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/716


pkg/installers/csr/interfaces.go
```diff
@@ -15,6 +15,7 @@ type CsrAgentInstallOptions struct {
 	UseDevCsrAgentChart  bool // if true, look for a chart in .tgz form in _output/helm/charts/csr-agent
 	ReleaseName          string
 	RemoteWriteNamespace string
+	ValuesFiles          []string
```
GrahamGoudeau, line 18: I'm not sure what to do about this for auto-registration of EKS clusters, since the user won't have the opportunity to provide values... Not sure that needs to block this, but something to think about

harveyxia: Good point—we could expose something through discovery settings in the Settings CRD, but we need to think about how to represent the value overrides. Raw strings would be one naive approach but it wouldn't scale well with a large list of values

PR #723, opened 2020-05-22
make demo work in linux

yuval-k:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/692


ci/setup-kind.sh
```diff
@@ -16,15 +16,15 @@
...
 fi
 
 make clean
 
-# generate 16-character random suffix on these names
-managementPlane=management-plane-$(xxd -l16 -ps /dev/urandom)
-remoteCluster=target-cluster-$(xxd -l16 -ps /dev/urandom)
+# generate 1 allows to make several envs in parallel
+managementPlane=management-plane-$1
+remoteCluster=target-cluster-$1
```
GrahamGoudeau, line 27: why these changes?

yuval-k: xxd is not available on linux

PR #726, opened 2020-05-23
Fix ttlDays casting (#725)

### LINKED ISSUE #725: Virtual Mesh printer uses incorrect int to string cast
virtual_mesh_printer.go
`ttlDays := string(int(certgen.DefaultRootCertTTLDays))`

https://golang.org/ref/spec#Conversions_to_and_from_a_string_type

> Converting a signed or unsigned integer value to a string type yields a string containing the UTF-8 representation of the integer.

Results in CLI output:

`TTL: ŭ days`

u5surf:
Fix #725 
* Fix casting ttlDays in virtual_mesh_printer.go with strconv.Itoa
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/725

### DIFF
--- a/changelog/v0.4.12/fix-ttldate-cast.yaml
+++ b/changelog/v0.4.12/fix-ttldate-cast.yaml
@@ -1,0 +1,4 @@
+changelog:
+  - type: FIX
+    description: Fix casting ttlDays in virtual_mesh_printer.go with strconv.Itoa
+    issueLink: https://github.com/solo-io/service-mesh-hub/issues/725
--- a/cli/pkg/common/table_printing/test_goldens/virtual_mesh/virtual_mesh.golden
+++ b/cli/pkg/common/table_printing/test_goldens/virtual_mesh/virtual_mesh.golden
@@ -6,7 +6,7 @@
 | Display Name: my favorite virtual mesh      | - mesh-3                 | Certificate Authority:                   |   Message: This is a conflict         |
 |                                             |                          |   Type: Self Signed                      |                                       |
 |                                             |                          |   Org Name: my-org                       | Config Status:                        |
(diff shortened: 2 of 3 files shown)


GrahamGoudeau:
Thanks a lot for the fast fix @u5surf! We have a test case covering this actually, that'll need to get updated. Would you run `go test` in the package `cli/pkg/common/table_printing/` after setting this var to true:

https://github.com/solo-io/service-mesh-hub/blob/master/cli/pkg/common/table_printing/virtual_mesh_printer_test.go#L21

That will cause the test goldens to be updated by the running test. Then change that var back to false before committing again. 

Thanks again!

u5surf:
> Thanks a lot for the fast fix @u5surf! We have a test case covering this actually, that'll need to get updated. Would you run `go test` in the package `cli/pkg/common/table_printing/` after setting this var to true:
> 
> https://github.com/solo-io/service-mesh-hub/blob/master/cli/pkg/common/table_printing/virtual_mesh_printer_test.go#L21
> 
> That will cause the test goldens to be updated by the running test. Then change that var back to false before committing again.
> 
> Thanks again!

@GrahamGoudeau 
Thanks reviwing, I got it doing that.

GrahamGoudeau:
Looks good to me @u5surf! One last thing, if you could add a brief changelog entry in the style of this:

https://github.com/solo-io/service-mesh-hub/blob/master/changelog/v0.4.9/docs-fix-access-control-guide.yaml

marked as a FIX, with a brief description, and an `issueLink` pointing to `https://github.com/solo-io/service-mesh-hub/issues/725`, and put that file in `changelog/v0.4.12`. We have automation that ensures a corresponding file is present for each change. 

Then it'll be good to merge!

GrahamGoudeau:
Merged- thanks again @u5surf!

PR #733, opened 2020-05-27
Capture Appmesh specific data during MeshWorkload discovery

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/732
resolves https://github.com/solo-io/service-mesh-hub/issues/730


ci/setup-kind-single.sh
```diff
@@ -14,7 +14,7 @@ fi
 make clean
 
 # generate 16-character random suffix on these names
-managementPlane=management-plane-$(xxd -l16 -ps /dev/urandom)
+managementPlane=management-plane
```
EItanya, line 17: If this is going to remain constant, then should we change the cleanup to only delete this one as well?

api/discovery/v1alpha1/mesh_workload.proto
```diff
@@ -13,6 +13,20 @@ option (gogoproto.equal_all) = true;
...
 message MeshWorkloadSpec {
+    message Appmesh {
+        // k8s application container ports
+        message ContainerPort {
+            uint32 port = 1;
+            string protocol = 2;
+        }
+
+        // The value of the env var APPMESH_VIRTUAL_NODE_NAME on the Appmesh envoy proxy container
+        string virtualNodeName = 1;
```
EItanya, line 24: for consistency can we do `virtual_node_name`?
```suggestion
        string virtual_node_name = 1;
```

services/mesh-discovery/pkg/discovery/mesh-workload/k8s/appmesh/appmesh_mesh_workload_scanner.go
```diff
@@ -124,3 +130,18 @@ func (a *appMeshWorkloadScanner) buildMeshWorkloadName(deploymentName string, na
 	// TODO: https://github.com/solo-io/service-mesh-hub/issues/141
 	return fmt.Sprintf("%s-%s-%s-%s", "appmesh", deploymentName, namespace, clusterName)
 }
+
+func (a *appMeshWorkloadScanner) getContainerPorts(
+	deployment *k8s_apps_types.Deployment,
+) []*zephyr_discovery_types.MeshWorkloadSpec_Appmesh_ContainerPort {
+	var ports []*zephyr_discovery_types.MeshWorkloadSpec_Appmesh_ContainerPort
```
EItanya, line 137: super nit: declare capacity since we have it.
```suggestion
	ports := make([]*zephyr_discovery_types.MeshWorkloadSpec_Appmesh_ContainerPort, 0, len(container.Ports))
```

api/discovery/v1alpha1/mesh_workload.proto
```diff
@@ -13,6 +13,20 @@ option (gogoproto.equal_all) = true;
...
+
+        // Needed for declaring Appmesh VirtualNode listeners
+        repeated ContainerPort ports = 2;
```
GrahamGoudeau, line 27: might we need the container ports more generally? might be a case to be made for moving this out of the appmesh-specific object. Not a big deal though, I'm fine either way

services/mesh-discovery/pkg/discovery/mesh-workload/k8s/appmesh/appmesh_mesh_workload_scanner.go
harveyxia: the length is actually the number of container ports across all containers, computing that would require looping through the containers anyways

api/discovery/v1alpha1/mesh_workload.proto
harveyxia: Good point, I thought about that but figured it's better to scope the metadata to a specific mesh until we find one more non-appmesh use case, at which point we can hoist it out

PR #754, opened 2020-06-04
use pinned versions from go mod instead of updating

yuval-k:
all the tools should be pinned using tools.go
we shouldnt use the `-u` flag so we always get the same version


yuval-k:
i can't seem to get go.mod / go.sum to a stable place :/

PR #757, opened 2020-06-05
show the user all conflicts from a traffic policy

yuval-k:


services/mesh-networking/pkg/traffic-policy-temp/aggregation/aggregator.go
```diff
@@ -10,7 +12,7 @@ import (
 )
 
 var (
-	TrafficPolicyConflictError = eris.New("Found conflicting TrafficPolicy")
+	TrafficPolicyConflictError = func(s string) error { return eris.Errorf("Found conflicting TrafficPolicy %s", s) }
```
harveyxia, line 15:
```suggestion
	TrafficPolicyConflictError = func(s string) error { return eris.Errorf("Found the following conflicts in TrafficPolicy: %s", s) }
```

PR #762, opened 2020-06-05
Appmesh EKS e2e

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/752


pkg/metadata/appmesh.go
```diff
@@ -1,13 +1,20 @@
...
+	"strings"
 
+	zephyr_discovery "github.com/solo-io/service-mesh-hub/pkg/api/discovery.zephyr.solo.io/v1alpha1"
+)
+
+// The VirtualService name must be a resolvable hostname. For k8s this means <svc-name>.<svc-namespace>
+// Reference: https://docs.aws.amazon.com/app-mesh/latest/userguide/virtual_services.html
 func BuildVirtualServiceName(meshService *zephyr_discovery.MeshService) string {
-	return meshService.Spec.GetKubeService().GetRef().GetName() + "-" + meshService.Spec.GetKubeService().GetRef().GetNamespace()
+	return meshService.Spec.GetKubeService().GetRef().GetName() + "." + meshService.Spec.GetKubeService().GetRef().GetNamespace()
```
harveyxia, line 12: This fixes the problem of "-" delimiters not being resolvable k8s DNS names

pkg/aws/translation/translation_reconciler.go
```diff
@@ -46,57 +46,34 @@ func (a *appmeshTranslationReconciler) Reconcile(
...
 	TODO: wire up this method when SMH API exposes sidecar configuration options
-	For every services declared as a target in at least one AccessControlPolicy, create an Appmesh VirtualService
-	that routes to the VirtualNodes for the services' backing workloads.
-
-	For every workload declared as a source in at least one AccessControlPolicy, create an Appmesh VirtualNode
-	with VirtualServices corresponding to ACP declared upstream services as backends.
+	Do not permit workloads to communicate to any other service in the mesh.
 */
-func (a *appmeshTranslationReconciler) reconcileWithEnforcedAccessControl(
+func (a *appmeshTranslationReconciler) reconcileWithLimitedRoutability(
```
harveyxia, line 60: The goal here is to prevent workloads from talking to any services (i.e. restore the Appmesh state to the default). I discovered upon testing that deleting VirtualNodes does not update the envoy configuration. Instead we need to update the VirtualNodes to remove the backends.

ci/bookinfo.yaml
```diff
@@ -325,6 +325,12 @@ spec:
         volumeMounts:
         - name: tmp
           mountPath: /tmp
+      # Include curl container for e2e testing, allows sending traffic mediated by the proxy sidecar
+      - name: curl
+        image: curlimages/curl@sha256:aa45e9d93122a3cfdf8d7de272e2798ea63733eeee6d06bd2ee4f2f8c4027d7c
```
yuval-k, line 330: with changes like this, its good to add a README.md here or in the e2e test folder explaining why this was done.

pkg/aws/translation/translation_reconciler.go
yuval-k: what happens in the case that the virtual node already had backends before it was added? 
i.e. do we take ownership of the virtual node?

test/e2e/appmesh_eks_test.go
```diff
@@ -0,0 +1,312 @@
...
+
+const (
+	AwsAccountId          = "410461945957"
+	Region                = "us-east-2"
+	AppmeshArn            = "arn:aws:appmesh:us-east-2:410461945957:mesh/smh-e2e-test"
+	EksArn                = "arn:aws:eks:us-east-2:410461945957:cluster/smh-e2e-test"
+	AppmeshName           = "smh-e2e-test"
+	EksClusterName        = "smh-e2e-test"
+	SmhNamespace          = "service-mesh-hub"
+	AppmeshInjectionLabel = "appmesh.k8s.aws/sidecarInjectorWebhook=enabled"
```
yuval-k, line 33: Would be good to have a README explaining all these, and how to recreate them if needed

test/e2e/appmesh_eks_test.go
```diff
@@ -0,0 +1,312 @@
...
+
+	buildSecretYaml = func(awsAccessKeyId, awsSecretAccessKey string) string {
+		return fmt.Sprintf(`
+apiVersion: v1
+kind: Secret
+type: solo.io/register/aws-credentials
+metadata:
+  name: %s
+  namespace: %s
+data:
```
yuval-k, line 106: you can use string_data to not need to base64.
i think this is also more secure, as github will be able to santize the secrets form the logs if they ever appear there

test/e2e/appmesh_eks_test.go
```diff
@@ -0,0 +1,312 @@
...
+`, secretObjKey.Name, secretObjKey.Namespace, awsAccessKeyId, awsSecretAccessKey)
+	}
+
+	kubeClusterName = metadata.BuildEksKubernetesClusterName(EksClusterName, Region)
+)
+
+func getEksKubeContext(ctx context.Context) KubeContext {
+	eg, ctx := errgroup.WithContext(ctx)
+
+	r, w, err := os.Pipe()
```
yuval-k, line 118: i dont see these used - can they be removed?

test/e2e/appmesh_eks_test.go
```diff
@@ -0,0 +1,312 @@
...
+
+	err = eg.Wait()
+	if err != nil {
+		dumpState()
+	}
+
+	Expect(err).NotTo(HaveOccurred())
+
+	// Use current context
+	return NewKubeContext("")
```
yuval-k, line 142: yes it possible to give an explicit context here?

test/e2e/appmesh_eks_test.go
```diff
@@ -0,0 +1,312 @@
...
+	eksContext := getEksKubeContext(ctx)
+	eksContext.DeleteNamespace(ctx, ns)
+}
+
+var _ = Describe("Appmesh EKS ", func() {
+	AfterEach(func() {
+		testLabels := map[string]string{"test": "true"}
+		opts := &client.DeleteAllOfOptions{}
+		opts.LabelSelector = labels.SelectorFromSet(testLabels)
+		opts.Namespace = "service-mesh-hub"
```
yuval-k, line 212: `opts` is not used here. is it used elsewhere?

test/e2e/appmesh_eks_test.go
```diff
@@ -0,0 +1,312 @@
...
+			"60s", "1s").
+			ShouldNot(BeNil())
+	}
+
+	var curlReviewsWithExpectedOutput = func(expectedString string, shouldExpect bool) {
+		productPageDeployment := "productpage-v1"
+		ctx := context.Background()
+		eksContext := getEksKubeContext(ctx)
+		eksContext.WaitForRollout(ctx, generatedNamespace, productPageDeployment)
+		if shouldExpect {
```
yuval-k, line 273: we can make this more readable. maybe?
```
eventuallyCurl := Eventually(func() string {
				return eksContext.Curl(
					context.Background(),
					generatedNamespace,
					productPageDeployment,
					"curl",
					fmt.Sprintf("http://reviews.%s:9080/reviews/1", generatedNamespace))
			}, "120s", "1s")
if (...){
eventuallyCurl.Should(..)
}else{
eventuallyCurl.ShouldNot(..)
}
```

test/e2e/appmesh_eks_test.go
```diff
@@ -0,0 +1,312 @@
...
+		opts := &client.DeleteAllOfOptions{}
+		opts.LabelSelector = labels.SelectorFromSet(testLabels)
+		opts.Namespace = "service-mesh-hub"
+	})
+
+	// Fetch base64 encoded AWS credentials from environment
+	var registerAwsSecret = func() {
+		awsAccessKeyId := os.Getenv("AWS_ACCESS_KEY_ID")
+		awsSecretAccessKey := os.Getenv("AWS_SECRET_ACCESS_KEY")
+		awsAccessKeyIdBase64 := base64.StdEncoding.EncodeToString([]byte(awsAccessKeyId))
```
yuval-k, line 219: as said above, base64 may not be needed

test/e2e/helpers.go
```diff
@@ -34,6 +35,20 @@ func StatusOfTrafficPolicy(tp v1alpha1.TrafficPolicy, kc KubeContext) v1alpha1ty
 	return newtp.Status.GetTranslationStatus().GetState()
 }
 
+func KubeClusterShouldExist(key client.ObjectKey, kc KubeContext) func() *zephyr_discovery.KubernetesCluster {
```
yuval-k, line 38: should this be called `KubeClusterShouldExist`?
that way the assertion reads better:
```
		Eventually(KubeCluster(...), "30s", "1s").ShouldNot(BeNil()) 
```

pkg/aws/translation/translation_reconciler.go
harveyxia: Yeah in that case SMH takes ownership and will update it if needed

ci/bookinfo.yaml
harveyxia: Added some notes to `test/e2e/README.md`

test/e2e/appmesh_eks_test.go
harveyxia: ah there is, just found the `alias` option [here](https://docs.aws.amazon.com/cli/latest/reference/eks/update-kubeconfig.html)

test/e2e/appmesh_eks_test.go
harveyxia: this whole block isn't needed

pkg/aws/translation/translation_reconciler.go
yuval-k: ok, that's fine for now, but we might want to re-visit this soon

test/e2e/cluster_lock.go
```diff
@@ -0,0 +1,60 @@
...
+
+	lock := &clusterLock{started: make(chan struct{})}
+	err = mgr.Add(lock)
+	Expect(err).NotTo(HaveOccurred())
+	fmt.Fprintln(GinkgoWriter, "Waiting to acquire cluster lock...")
+	select {
+	case <-ctx.Done():
+		fmt.Fprintln(GinkgoWriter, "Releasing cluster lock")
+	case <-time.After(timeout):
+		fmt.Fprintln(GinkgoWriter, "Timed out waiting for cluster lock")
```
yuval-k, line 56: should there be a behavior for releasing and time out?
i.e. on timeout we should probably call Fail?
any action needed to release the lock?

test/e2e/appmesh_eks_test.go
```diff
@@ -0,0 +1,293 @@
...
+	// https://github.com/istio/istio/blob/1.5.0/samples/bookinfo/src/productpage/productpage.py#L60
+	eksKubeContext.SetDeploymentEnvVars(ctx, generatedNamespace, "productpage-v1", "productpage", map[string]string{"SERVICES_DOMAIN": generatedNamespace})
+	eksKubeContext.SetDeploymentEnvVars(ctx, generatedNamespace, "reviews-v1", "reviews", map[string]string{"SERVICES_DOMAIN": generatedNamespace})
+	eksKubeContext.SetDeploymentEnvVars(ctx, generatedNamespace, "reviews-v2", "reviews", map[string]string{"SERVICES_DOMAIN": generatedNamespace})
+	eksKubeContext.SetDeploymentEnvVars(ctx, generatedNamespace, "reviews-v3", "reviews", map[string]string{"SERVICES_DOMAIN": generatedNamespace})
+	return generatedNamespace
+}
+
+func cleanupAppmeshEksEnvironment(ns string) {
+	if useExisting := os.Getenv("USE_EXISTING"); useExisting != "" {
```
yuval-k, line 167: most of cleanup is done if USE_EXISTING is not empty? i.e. we are re-using an existing environment? should this be the other way? i.e. clean up resources that we created

test/e2e/cluster_lock.go
harveyxia: The nice thing about using the leader election logic is that no manual release is needed—when the running test ends, the configmap will automatically be up for acquisition by other tests

test/e2e/appmesh_eks_test.go
harveyxia: yup that's correct—this cleanup is only needed if we're reusing the management-plane cluster for multiple iterations of the test, otherwise the cluster will just be torn down anyways so no need to clean anything up

PR #763, opened 2020-06-08
Update setup-kind.sh is used to solve the problem of King running failure

synuwxy:
The default version of k8s under Linux is 1.18，set version to 1.17.5

Using the default kind version will cause istiod to run error

see: https://github.com/solo-io/service-mesh-hub/issues/700
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/718


harveyxia:
@synuwxy Thanks for the contribution!

PR #773, opened 2020-06-12
refactor skv2 codegen

ilackarms:


harveyxia:
The new skv2 `Upsert` methods have different semantics than the `UpsertSpec` methods that they're replacing. `UpsertSpec` [copies over all non-Spec data](https://github.com/solo-io/service-mesh-hub/blob/e536c0aa50a34e69c238f01a3b260c7fa667672f/pkg/api/discovery.smh.solo.io/v1alpha1/clients.go#L269) from the existing obj to the desired obj before calling `Update`.

Would we need to utilize the transitionFuncs to preserve the above semantics?

harveyxia:
Chatted offline, [this portion](https://github.com/solo-io/skv2/blob/master/pkg/controllerutils/upsert.go#L57-L59) of the skv2 `Upsert` method handles updating the resource version, so it should be semantically equivalent to the old `UpsertSpec` methods

PR #774, opened 2020-06-15
add stubs for skv2 reconcilers

ilackarms:


harveyxia:
Mesh, MeshWorkload, MeshService, VirtualMesh, TrafficPolicy, and AccessControlPolicy can be switched to a single-cluster reconciler.

The only multicluster reconcilers we need are for Pod, Deployment, and Service

PR #783, opened 2020-06-17
initial wiring

yuval-k:


pkg/mesh-networking/run.go
```diff
@@ -85,3 +88,15 @@ func startComponents(meshNetworkingContext wire.MeshNetworkingContext) func(cont
 		return nil
 	}
 }
+
+func startTrafficPolicyReconciler(ctx context.Context, meshNetworkingContext wire.MeshNetworkingContext) {
```
harveyxia, line 92: I would leave another comment here marking this as a temporary approach

pkg/mesh-networking/traffic-policy-temp/translation/framework/snapshot/reconcilers/destination_rule.go
```diff
@@ -68,7 +69,7 @@ func (v *destinationRuleReconcilerBuilder) Build() (DestinationRuleReconciler, e
 		return nil, eris.New("Must either configure this reconciler to be cluster-scoped or explicitly scope it to a namespace/label")
 	}
 
-	if v.clusterScoped && (v.namespace != "" || len(v.labels) > 0) {
+	if v.clusterScoped && v.namespace != "" {
 		return nil, eris.New("Cannot be cluster-scoped and scoped to a namespace/label")
```
harveyxia, line 73:
```suggestion
		return nil, eris.New("Cannot be cluster-scoped and scoped to a namespace")
```

harveyxia: Also the top-level comment in this file needs to be updated, specifically the line `// a reconciler can either be whole-cluster scoped or scoped to a namespace and/or labels`

pkg/mesh-networking/run.go
```diff
@@ -18,7 +18,7 @@ func Run(ctx context.Context) {
 
 	// build all the objects needed for multicluster operations
 	meshNetworkingContext, err := wire.InitializeMeshNetworking(
-		contextutils.WithLogger(ctx, "access_control_enforcer"),
+		ctx,
```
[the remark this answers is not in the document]
yuval-k: yea it got me confused..

PR #813, opened 2020-07-06
Versioned docs

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/811


.github/workflows/docs.yaml
```diff
@@ -34,31 +34,7 @@ jobs:
...
-      # Found on github community: https://github.com/ameydev/gke-kubectl-action
-      uses: ameydev/gke-kubectl-action@v1.02
-      env:
-        PROJECT_ID: ${{ secrets.PROJECT_ID }}
-        APPLICATION_CREDENTIALS: ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS }}
-        CLUSTER_NAME: ${{ secrets.GKE_CLUSTER_NAME }}
-        ZONE_NAME: us-central1-a
-      with:
-        args: -n docs rollout restart deployment/service-mesh-hub-docs-latest
+        FIREBASE_TOKEN: ${{ secrets.FIREBASE_TOKEN }}
```
kdorosh, line 40: i don't think we actively use firebase for docs hosting anymore. is moving to this an intentional decision? if so, why?

yuval-k: it is;
the current docs cluster is in a bad state, and deploying to it is complicated.
we want to try a model where all versions are publish to a single static website

ci/build-docs-site.sh
```diff
@@ -0,0 +1,89 @@
...
+# This script generates a versioned docs website for Service Mesh Hub and
+# deploys to Firebase.
+###################################################################################
+
+set -ex
+
+# Update this array with all versions of SMH to include in the versioned docs website.
+declare -a versions=(
+"master"
+"0.5.0"
```
ilackarms, line 13: i wonder if we could read this data from somewhere, maybe by picking a naming convention for our tags, and then grabbing them dynamically from github? 

or maybe there's a better place we can maintain this data where we don't need to manually update a script

ci/build-docs-site.sh
```diff
@@ -0,0 +1,89 @@
...
+      "**/node_modules/**"
+    ],
+    "rewrites": [
+      {
+        "source": "/",
+        "destination": "/master/index.html"
+      },
+      {
+        "source": "/latest",
+        "destination": "/0.5.0/index.html"
```
ilackarms, line 34: was hardcoding `0.5.0` a mistake?  shouldn't we be generating this for the currrent release? 

i'm starting to think that we should maintain a release file somewhere, which indexes all our releases. if you think that makes sense

ci/build-docs-site.sh
```diff
@@ -0,0 +1,89 @@
...
+docsSiteDir=$workingDir/ci/docs-site
+repoDir=$workingDir/ci/service-mesh-hub-temp
+
+mkdir $docsSiteDir
+echo $firebaseJson > $docsSiteDir/firebase.json
+
+git clone git@github.com:solo-io/service-mesh-hub.git $repoDir
+
+# Generates a data/Solo.yaml file with $1 being the specified version.
+function generateHugoVersionsYaml() {
```
ilackarms, line 53: would you mind adding a `make` target that essentially accomplishes this function? i like having a make target if it's gonna modify something in the current repo (so i can set it back, etc.)

ilackarms: this could go in the docs Makefile

harveyxia: Discussed offline, because this function isn't modifying files in the main repo (instead it's modifying a file in the repo cloned during the execution of this script), we don't need a make target

ci/build-docs-site.sh
harveyxia: I had intended for this to be manually updated as well, but the latest commit prevents us from having to modify this script, instead reading declared versions from `docs/version.json`.

ci/build-docs-site.sh
harveyxia: Good call, I create a `docs/version.json` to house version data, the script reads from that file

PR #829, opened 2020-07-14
Export skv2 groups

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/828

### DIFF
--- a/generate.go
+++ b/generate.go
@@ -3,19 +3,14 @@ package main
 import (
 	"log"
 
-	"github.com/solo-io/service-mesh-hub/pkg/common/constants"
-	"github.com/solo-io/skv2/codegen"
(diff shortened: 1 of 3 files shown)


generate.go
```diff
@@ -45,87 +40,30 @@ func run() error {
...
 		AnyVendorConfig: protoImports,
 		ManifestRoot:    smhCrdManifestRoot,
-		Groups:          groups,
+		Groups:          codegen.SmhGroups,
 		RenderProtos:    true,
 	}
 }
 
-func makeCsrCommand() codegen.Command {
+func makeCsrCommand() skv2_codegen.Command {
```
harveyxia, line 53: I left the CsrAgent group here because I don't think it's likely that other projects will need to reference this

PR #838, opened 2020-07-21
E2e refactor

ilackarms:
main entrypoint, dockerfile, helm chart, makefile, CI, wiring for e2e test


.circleci/config.yml
```diff
@@ -0,0 +1,29 @@
+# Golang CircleCI 2.0 configuration file
```
EItanya, line 1: Why the swtich to circle?  github actions can do this easily

PR #843, opened 2020-07-27
Merge repos decorator unit tests

harveyxia:
Depends on (i.e. is a superset of the changes in) #842


harveyxia:
changes already contained in #842

PR #851, opened 2020-08-03
E2e federation + shared trust

ilackarms:
this PR finishes implementing shared trust using a certificate-agent installed to the remote cluster.

includes an e2e test for federation


pkg/certificates/agent/reconciliation/cert_agent_reconciler.go
```diff
@@ -0,0 +1,204 @@
...
+	// reset & update status
+	issuedCertificate.Status.ObservedGeneration = issuedCertificate.Generation
+	issuedCertificate.Status.Error = ""
+
+	// state-machine style processor
+	switch issuedCertificate.Status.State {
+	case v1alpha2.IssuedCertificateStatus_FINISHED:
+		// ensure issued cert secret exists, nothing to do for this issued certificate
+		if _, err := inputSecrets.Find(issuedCertificate.Spec.IssuedCertificateSecret); err == nil {
+			return nil
```
harveyxia, line 101: There's a chance the cache might not return the written IssuedCertificate secret (because of weak read-after-write consistency), in which case the entire flow will be unnecessarily reinitiated. Is this something we should be concerned about /  if so is there anything we can do to decrease the likelihood of this occurring?

pkg/certificates/issuer/reconciliation/cert_issuer_reconciler.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/solo-io/skv2/pkg/ezkube"
+	"github.com/solo-io/skv2/pkg/multicluster"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type certIssuerReconciler struct {
+	ctx           context.Context
+	builder       input.Builder
+	mcClient      multicluster.Client
+	masterSecrets corev1.SecretClient
```
harveyxia, line 24: Why are we fetching secrets using a client vs. reading them from the input snapshot?

pkg/certificates/issuer/reconciliation/cert_issuer_reconciler.go
```diff
@@ -0,0 +1,118 @@
...
+	signingCA := secrets.RootCADataFromSecretData(signingCertificateSecret.Data)
+
+	// generate the issued cert PEM encoded bytes
+	signedCert, err := utils.GenCertForCSR(
+		certificateRequest.Spec.CertificateSigningRequest,
+		signingCA.PrivateKey,
+		signingCA.RootCert,
+	)
+	if err != nil {
+		return eris.Wrapf(err, "failed to find master signing certificate matching issued request %v", sets.Key(issuedCertificate))
```
harveyxia, line 107:
```suggestion
		return eris.Wrap(err, "failed to generate issued cert")
```

pkg/mesh-networking/translation/istio/mesh/access/access_policy_translator.go
```diff
@@ -21,7 +21,8 @@ type Translator interface {
 	Translate(
 		mesh *discoveryv1alpha2.Mesh,
 		virtualMesh *discoveryv1alpha2.MeshStatus_AppliedVirtualMesh,
-	) v1beta1sets.AuthorizationPolicySet
+		outputs output.Builder,
```
harveyxia, line 24: I would update the comment to mention that `outputs` is mutated with the newly translated output resources. Same for the other translators that have been similarly refactored

pkg/meshctl/install/helm/installer.go
```diff
@@ -0,0 +1,200 @@
...
+func (i Installer) InstallChart() error {
+	kubeConfig := i.KubeConfig
+	kubeContext := i.KubeContext
+	chartUri := i.ChartUri
+	namespace := i.Namespace
+	releaseName := i.ReleaseName
+	valuesFile := i.ValuesFile
+	verbose := i.Verbose
+
+	actionConfig, settings, err := newActionConfig(kubeConfig, kubeContext, namespace)
```
harveyxia, line 47: Out of curiosity was there a particular reason we couldn't depend on the Helm utils in `github.com/solo-io/go-utils/installutils/helminstall`?

pkg/meshctl/install/helm/installer.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+
+		logrus.Infof("finished upgrading chart as release: %+v", release)
+
+	} else {
+		// release does not exist, perform install
+
+		client := action.NewInstall(actionConfig)
+		client.ReleaseName = releaseName
+		client.Namespace = namespace
```
harveyxia, line 87: Will this work if the namespace doesn't already exist?

harveyxia: The go-utils package I mentioned above [will create the namespace if it doesn't exist](https://github.com/solo-io/go-utils/blob/master/installutils/helminstall/installer.go#L53)

pkg/certificates/common/secrets/cert_secrets.go
```diff
@@ -0,0 +1,73 @@
+package secrets
+
+const (
+	/*
+		TODO(ilackarms): document the expected structure of secrets (required for VirtualMeshes  using a user-provided root CA)
```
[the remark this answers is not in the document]
ilackarms: yes

api/certificates/issued_certificate.proto
```diff
@@ -51,11 +54,31 @@ message IssuedCertificateStatus {
...
+    // the current state of the IssuedCertificate workflow,
+    // reported by the agent.
+    State state = 3;
+
+    // possible states in which an IssuedCertificate can exist
+    enum State {
+        // the IssuedCertificate has yet to be picked up by the agent.
+        PENDING = 0;
+
+        // the agent has created a local private key
```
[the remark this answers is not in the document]
ilackarms: i've actually seen the opposite in the past, which is why i write my comments like this. let's address when we start generating our docs

pkg/certificates/agent/reconciliation/cert_agent_reconciler.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+		// otherwise, restart the workflow from PENDING
+		fallthrough
+	case v1alpha2.IssuedCertificateStatus_FAILED:
+		// restart the workflow from PENDING
```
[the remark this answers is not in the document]
ilackarms: i believe the flow should only fail due to an internal error (components miscommunicating) or an invalid setup performed by the user. do you have a possible scenario in mind?

pkg/certificates/issuer/reconciliation/cert_issuer_reconciler.go
ilackarms: the case here is a bit special. inputs (and the snapshot builder) for the cert issuer come from managed clusters. here we actually need to read secrets from the master cluster. the option was either to create a complex (and imo convoluted) snapshot which includes both master-cluster and managed-cluster resources, or simply use a client to access the master cluster.

i don't see a huge impact here as we don't need to "watch" the master cluster secret - the issuedCertificate is the source of truth for this workflow and will be updated should the workflow ever need to be repeated (such as in the case of cert rotation).

does not follow our input/output snapshot convention strictly, but i feel it's a reasonable case that's better suited to direct client access

pkg/meshctl/install/helm/installer.go
ilackarms: i went with this model for its simplicity over that code, as well as the ability to perform upgrades when a release already exists (rather than erroring out)

pkg/meshctl/install/helm/installer.go
ilackarms: yes, there is code for handling namespace creation that gets run before this

PR #862, opened 2020-08-11
Federated subsets

ilackarms:


pkg/mesh-networking/translation/istio/decorators/decorators.go
```diff
@@ -0,0 +1,116 @@
...
+
+type RegisterField func(fieldPtr, val interface{}) error
+
+/*
+	Interface definitions for decorators which take VirtualMesh as an input and
+	decorate a given output resource.
+*/
+
+// a VirtualMeshDestinationRuleDecorator modifies the DestinationRule based on a VirtualMesh which applies to the MeshService.
+type VirtualMeshDestinationRuleDecorator interface {
```
harveyxia, line 66: What's the reason for defining all decorator interfaces at `pkg/mesh-networking/translation/istio/decorators` vs previously where we were defining the interfaces in an input CRD scoped package, e.g. `pkg/mesh-networking/translation/istio/decorators/trafficpolicy`?

Are decorators no longer scoped to a single input CRD?

pkg/mesh-networking/translation/istio/decorators/trafficshift/traffic_shift_decorator.go
```diff
@@ -211,13 +188,20 @@ func (d *trafficShiftDecorator) translateSubset(
 		uniqueSubsets = append(uniqueSubsets, subsetLabels)
 	}
 
-	for _, policy := range appliedPolicies {
-		for _, destination := range policy.GetSpec().GetTrafficShift().GetDestinations() {
-			if subsetLabels := destination.GetKubeService().GetSubset(); len(subsetLabels) > 0 {
-				appendUniqueSubset(subsetLabels)
+	allMeshServices.List(func(service *discoveryv1alpha2.MeshService) bool {
```
harveyxia, line 191: Interesting use of list + filter, it's a bit of a weird convention though—why not just list all mesh services and then range?

ilackarms: to skip iterating twice. not overly concerned about this and i can switch it to the other approach if preferred. since we are inserting into a map, ordering doesn't matter (when ordering does matter, you want the result of List() because it's sorted)

pkg/mesh-networking/translation/istio/decorators/decorators.go
ilackarms: i started with this change because i was originally planning on adding decorators to the federation translator (in which case we'd be using a decorator which applies a VirtualMesh to a meshservice). 

i chose not to go that route in the end, but i decided to leave in this bit of refactoring because i believe it's something we may want in the near future (in particular if we extend the existing decorators to work with federated & failover services)

i'm ok with changing it back if you have any objections, but it's my suspicion this will come up again and be useful to us then

PR #864, opened 2020-08-12
use matrix strategy for testing istio minor versions

harveyxia:


ci/setup-funcs.sh
```diff
@@ -152,7 +152,6 @@ spec:
         enabled: true
       podDNSSearchNamespaces:
       - global
-      - '{{ valueOrDefault .DeploymentMeta.Namespace "default" }}.global'
```
harveyxia, line 155: Support for `valueOrDefault` [appears to have been dropped in Istio 1.6.x](https://github.com/istio/istio/issues/23711), I'm not sure if this is needed for testing Istio 1.5 (the e2e tests pass without it). @ilackarms let me know if you have context here

ilackarms: this yaml was a copy-paste from the original smh; if the tests pass, my inclination is that we don't need it

PR #870, opened 2020-08-14
Merge repos one virtualmesh per mesh

harveyxia:


.circleci/config.yml
```diff
@@ -9,6 +9,8 @@ jobs:
   build:
     machine:
       image: ubuntu-1604:201903-01
+      # need to upgrade plan before using large executor
+      # resource_class: large
```
joekelley, line 13: do we need this upgrade? Looks like CI is green as-is.

.github/workflows/pull_request.yaml
```diff
@@ -25,6 +25,13 @@ jobs:
           key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
           restore-keys: |
             ${{ runner.os }}-go-
+      - name: Install Protoc
+        uses: solo-io/setup-protoc@master
+        with:
+          version: '3.6.1'
+      - name: Check code gen
+        run: |
+          ./ci/check-code-gen.sh
```
joekelley, line 34: In other projects we've made this a separate job from unit tests. That way we get better parallelization and a better bird-eye-view of why CI is failing.

api/networking/v1alpha2/virtual_mesh.proto
```diff
@@ -149,7 +150,7 @@ message VirtualMeshStatus {
     // the state of the overall resource.
     // will only show accepted if it has been successfully
     // applied to all target meshes.
-    ApprovalState state = 2;
+    ApprovalStatus status = 2;
 
     // The status of the VirtualMesh for each Mesh to which it has been applied.
     // A TrafficPolicy may be Accepted for some Meshes and rejected for others.
```
joekelley, line 156: While we're here, this comment seems out of place

api/networking/v1alpha2/virtual_mesh.proto
```diff
@@ -149,7 +150,7 @@ message VirtualMeshStatus {
     // the state of the overall resource.
     // will only show accepted if it has been successfully
     // applied to all target meshes.
```
joekelley, line 152: nit: let's update this comment to describe status rather than state

pkg/common/bootstrap/bootstrap.go
```diff
@@ -20,12 +20,17 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 )
 
+type StartParameters struct {
```
joekelley, line 23: mega-nit: I see structs named like "StartOptions" or "StartOpts" in many of the kube/controller libraries we interact with. In this case though, many of these seem to be required, so perhaps "Options" is misleading. wdyt?

.circleci/config.yml
harveyxia: I just left this here for reference in case we ever need/want to use larger machines

api/networking/v1alpha2/virtual_mesh.proto
```diff
@@ -141,18 +142,17 @@ message VirtualMeshSpec {
...
+    // If the observedGeneration does not match generation, the controller has not received the most
     // recent version of this resource.
     int64 observed_generation = 1;
 
-    // the state of the overall resource.
-    // will only show accepted if it has been successfully
+    // The status of the overall resource. It will only show accepted if it has been successfully
     // applied to all target meshes.
-    ApprovalState state = 2;
+    ApprovalStatus status = 2;
```
ilackarms, line 152: why this rename?

ilackarms: the convention i've been adhering to is as follows: 
- `status` represents the kubernetes subresource `status` which aggregates all information about the current state of the object. 
- `state` represents the current *state* of the object as an enum. 

i'd like to try to avoid overloading the term `status` if possible.

pkg/mesh-networking/approval/approver_test.go
```diff
@@ -5,14 +5,15 @@ import (
...
+	corev1 "github.com/solo-io/skv2/pkg/api/core.skv2.solo.io/v1"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	discoveryv1alpha2 "github.com/solo-io/service-mesh-hub/pkg/api/discovery.smh.solo.io/v1alpha2"
 	discoveryv1alpha2sets "github.com/solo-io/service-mesh-hub/pkg/api/discovery.smh.solo.io/v1alpha2/sets"
 	"github.com/solo-io/service-mesh-hub/pkg/api/networking.smh.solo.io/input"
 	"github.com/solo-io/service-mesh-hub/pkg/api/networking.smh.solo.io/output"
-	"github.com/solo-io/service-mesh-hub/pkg/api/networking.smh.solo.io/v1alpha2"
+	networkingv1alpha2 "github.com/solo-io/service-mesh-hub/pkg/api/networking.smh.solo.io/v1alpha2"
```
ilackarms, line 16: why this change? 

in `discovery`, the discovery api is called `v1alpha2`, because it's the "local" api group.

the same thing is true in `networking`.

api/networking/v1alpha2/virtual_mesh.proto
harveyxia: I thought we needed AcceptanceOrder, but turns out we don't— @ilackarms makes the point that since the algorithm has the invariant of never permitting overlapping VirtualMeshes to be accepted.

PR #875, opened 2020-08-19
Merge repos docs

harveyxia:


.github/workflows/docs.yaml
```diff
@@ -0,0 +1,48 @@
...
+    name: Generate versioned docs site
+    runs-on: ubuntu-18.04
+    steps:
+    - name: Set up Go 1.14
+      uses: actions/setup-go@v1
+      with:
+        go-version: 1.14
+      id: go
+    - name: Install Protoc
+      uses: solo-io/setup-protoc@master
```
EItanya, line 22: Can we pin to a non-master version?

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+    - name: Update deps
+      run: |
+        make install-go-tools
+
+    - name: Generate versioned docs site
+      run: ./ci/build-docs-site.sh
+
+    - name: Deploy to Firebase
+      uses: w9jds/firebase-action@master
```
EItanya, line 42: same as above

docs/conf/conf.d/default.conf
```diff
@@ -0,0 +1,29 @@
...
+  location / {
+    root   /usr/share/nginx/html/no_auth;
+    index  index.html index.htm;
+    try_files $uri $uri/index.html @docs_error;
+  }
+  location @docs_error {
+    # this rewrite expects a "docerr" endpoint to be available with a service that logs the failed request path and returns a friendly response
+    # rewrite ^ $scheme://$host/docerr/$uri break;
+
+    # this form of rewrite could be used to point to a static site that is packaged with the docs or, as in it's current form, redirect to the latest page of web-assembly-hub
```
EItanya, line 21: `web-assembly-hub`
Do we still need this if we aren't using nginx?

harveyxia: nope, removing

.github/workflows/docs.yaml
harveyxia: Unfortunately you can't pin to a commit. Pinning to master in this case seems acceptable because it's a Solo-managed Github action, we'll receive communication about it before anyone pushes breaks to master (or shortly after), given that other projects are depending on master as well

PR #879, opened 2020-08-20
explicitly set output clusters in mesh translator

ilackarms:


harveyxia:
https://github.com/solo-io/skv2/pull/119 is merged, so let's depend on that commit in master, `go get github.com/solo-io/skv2@b998dc4f14993d5c033ab81b6cb96656a7b9c285`

PR #903, opened 2020-08-25
Add FailoverService traffic shift type.

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/899


pkg/mesh-networking/translation/istio/decorators/trafficshift/traffic_shift_decorator.go
```diff
@@ -169,6 +186,28 @@ func (d *trafficShiftDecorator) buildKubeTrafficShiftDestination(
...
+	weight uint32,
+) (*networkingv1alpha3spec.HTTPRouteDestination, error) {
+	failoverService, err := d.failoverServices.Find(failoverServiceRef)
+	if err != nil {
+		return nil, eris.Wrapf(err, "invalid traffic shift destination %s, FailoverService not found", sets.Key(failoverServiceRef))
+	}
+
+	httpRouteDestination := &networkingv1alpha3spec.HTTPRouteDestination{
+		Destination: &networkingv1alpha3spec.Destination{
+			Host: failoverService.Spec.Hostname,
```
ilackarms, line 200: aren't failover services required to have a `.global` suffix? just wondering if we've performed that check ourselves anywhere (or we appended it ourselves?)

pkg/mesh-networking/translation/istio/traffictarget/virtualservice/virtual_service_translator.go
```diff
@@ -87,6 +86,11 @@ func (t *translator) Translate(
 			}
 		}
 
+		// Avoid appending an HttpRoute that will have no affect, which occurs if no decorators mutate the baseRoute with any non-match config.
+		if equalityutils.Equals(initializeBaseRoute(policy.Spec), baseRoute) {
```
ilackarms, line 90: can you provide some context here?

harveyxia: If the decorators above don't mutate `baseRoute`, we shouldn't append any `HttpRoutes`, otherwise we'll end up with a VirtualService consisting of a host with a traffic shift to that same host, with no other config. This behavior can be reproduced with a TrafficPolicy that only declares OutlierDetection.

pkg/mesh-networking/translation/istio/decorators/trafficshift/traffic_shift_decorator.go
harveyxia: Thanks for reminding me, I added that to the list of FailoverService validation checks

PR #907, opened 2020-08-25
Traffic policy guide

ned1313:
Adding a guide on creating a basic Traffic Policy resource for the Pet Store application.


docs/content/guides/traffic_policy.md
```diff
@@ -0,0 +1,185 @@
...
+  - retries:
+      attempts: 5
+      perTryTimeout: 0.005s
+    route:
+    - destination:
+        host: petstore.default.svc.cluster.local
+    timeout: 0.100s
+```
+
+As we can see above, the proper retry and timeout settings have been applied to the VirtualService from the Service Mesh Hub TrafficPolicy. This feature can be extended to configure many services across multiple service meshes and clusters.
```
joekelley, line 181: let's also add a note that many other features can be configured via the traffic policy (e.g. fault injection, traffic mirroring) and link users to the API https://docs.solo.io/service-mesh-hub/latest/reference/api/traffic_policy/#networking.smh.solo.io.TrafficPolicySpec

ned1313: Added.

PR #914, opened 2020-08-26
Osm docs

ned1313:
Updating the OSM guide for clarity and tested functionality.


docs/content/guides/osm/_index.md
```diff
@@ -0,0 +1,173 @@
...
+``` 
+
+This command will create a local Kind cluster, install OSM, install Service Mesh Hub, and deploy a sample application.
+
+### Install manually
+
+If you prefer to install the components yourself, first you will need to install OSM using the `osm` CLI tool.
+
+```shell
+osm install --enable-metrics-stack=false --deploy-zipkin=false
```
EItanya, line 40: I think we should remove these flags for the purpose of the demo, and the command, but not the CI script. It is meant to save CPU in CI and when running tests, but most users will probably want metrics and tracing.

docs/content/guides/osm/_index.md
```diff
@@ -0,0 +1,173 @@
...
+meshctl cluster register --cluster-name management-cluster --remote-context management-cluster-context
+```
+
+Finally, we will deploy the sample application.
+
+```shell
+kubectl create ns bookstore
+kubectl create ns bookthief 
+kubectl create ns bookwarehouse 
+kubectl create ns bookbuyer
```
EItanya, line 64: Also need to run `osm add <ns>` for all of the namespaces

docs/content/guides/osm/_index.md
```diff
@@ -0,0 +1,173 @@
...
+      service_accounts:
+      - cluster_name: mgmt-cluster
+        name: bookthief
+        namespace: bookthief
+      - cluster_name: mgmt-cluster
+        name: bookstore-v1
+        namespace: bookstore
+      - cluster_name: mgmt-cluster
+        name: bookstore-v2
+        namespace: bookstore
```
EItanya, line 132: I'm curious why these 2 sources are necessary, the bookstore should never need to talk to the bookstore. You said it doesn't work without these?

PR #937, opened 2020-09-02
Upgrade to new skv2 registration interface

harveyxia:
Depends on https://github.com/solo-io/skv2/pull/135

Replace go mod dependency with released version after the above merges and is released
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/950


pkg/meshctl/install/agent/install_cert_agent.go
```diff
@@ -1,84 +0,0 @@
-package agent
```
harveyxia, line 1: This whole file was unused so I removed it

pkg/meshctl/utils/flag_utils.go
```diff
@@ -1,10 +1,39 @@
...
+		return nil, nil, err
+	}
+	remoteKubeCfg, err = loader.GetClientConfigForContext(m.kubeConfigPath, m.RemoteContext)
+	if err != nil {
+		return nil, nil, err
+	}
+	return mgmtKubeCfg, remoteKubeCfg, nil
+}
+
+func AddMgmtRemoteKubeConfigFlags(kubeConfigOptions *MgmtRemoteKubeConfigOptions, flags *pflag.FlagSet) {
```
EItanya, line 35:
```suggestion
func AddRemoteKubeConfigFlags(kubeConfigOptions *MgmtRemoteKubeConfigOptions, flags *pflag.FlagSet) {
```
?

pkg/meshctl/install/helm/internal/memory_rest_client_getter.go
```diff
@@ -0,0 +1,92 @@
+package internal
+
+// Copied from https://github.com/solo-io/go-utils/blob/2af7be12c6cfd50ed5890f59fbf03b4b404c5707/installutils/helminstall/internal/memory_backed_rest_client_getter.go#L1
```
EItanya, line 3: Should we put this code somewhere shared and accessible so we don't need to keep copy pasting it around?

pkg/common/registration/registrant.go
```diff
@@ -22,7 +23,15 @@ var smhRbacRequirements = func() []rbacv1.PolicyRule {
 	return policyRules
 }()
 
-type Registrant struct {
+//go:generate mockgen -source ./registrant.go -destination ./mocks/mock_registrant.go
+
+type ClusterRegistrant interface {
+	RegisterCluster(ctx context.Context) error
+	RegisterProviderCluster(ctx context.Context, providerInfo *v1alpha1.KubernetesClusterSpec_ProviderInfo) error
```
EItanya, line 30: Should we just combine these into one `RegisterCluster(ctx context.Context, providerInfo *v1alpha1.KubernetesClusterSpec_ProviderInfo) error` since that's how the impl works anyway.

pkg/meshctl/install/helm/internal/memory_rest_client_getter.go
harveyxia: Good question, I wasn't sure about making it public because I'm not sure about the status of `go-utils`. It seems like we're moving away from depending on it?

ilackarms (approved):

PR #956, opened 2020-09-11
Validation schema

harveyxia:
do not merge, depends on https://github.com/solo-io/skv2/pull/24
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/512


changelog/v0.7.6/crd-validation-schemas.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+  - type: NEW_FEATURE
```
ashleywang1, line 2: If we change the protos (int64 to uint32), does that make it a breaking change?

harveyxia: If the type change was on a user-supplied field, then strictly speaking it would be a breaking change. But the only int64's in SMH protos were on status fields, which are set by our internal logic, so it's not breaking from the user's perspective

harveyxia: Amending my comment—it is a breaking change for the user if they upgrade to this code, since their existing CRDs would break when read in by the new logic

joekelley:
Marking as WIP until https://github.com/solo-io/skv2/pull/24 is merged and released

api/networking/v1alpha2/traffic_policy.proto
```diff
@@ -161,36 +161,26 @@ message TrafficPolicySpec {
...
-
         /*
             The _fixedDelay_ field is used to indicate the amount of delay in seconds.
             The optional _percentage_ field can be used to only delay a certain
             percentage of requests. If left unspecified, all request will be delayed.
         */
-        message Delay {
-            oneof http_delay_type {
-                /*
+        oneof fault_injection_type {
```
harveyxia, line 169: This refactor was needed because the cuelang parser wasn't able to parse nested oneofs. Conveniently, I think this refactored structure is clearer

ilackarms (approved):

PR #969, opened 2020-09-24
policy status workload list fix

ashleywang1:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/967


ilackarms (approved):

pkg/mesh-networking/translation/utils/selectorutils/selector_utils.go
```diff
@@ -38,7 +37,7 @@ func IdentityMatchesWorkload(selectors []*v1alpha2.IdentitySelector, workload *d
 	for _, selector := range selectors {
 		kubeWorkload := workload.Spec.GetKubernetes()
 		if kubeWorkload != nil {
-			if kubeWorkloadIdentityMatcher := selector.KubeIdentityMatcher; kubeWorkloadIdentityMatcher != nil {
+			if kubeWorkloadIdentityMatcher := selector.GetKubeIdentityMatcher(); kubeWorkloadIdentityMatcher != nil {
```
harveyxia, line 40: a stylistic convention—we prefer to access by field instead of using the nil-safe getter. This approach communicates to the reader that the base object is expected to be non-nil

harveyxia: Amending my comment, this convention doesn't apply here because we _do not_ have the guarantee that all selectors are non-nil. The user could theoretically pass in `null` as a selector, which would appear as nil here.

So we could either use the field accessors with a nil check at the beginning, or use the getter methods. I'm fine with using the getter methods

harveyxia (approved):

PR #976, opened 2020-09-28
fix Issuedcertificate pod restarts

ilackarms:
fixes the issue by extracting `podsToBounce` to its own CRD. thus, when `podsToBounce` changes (when `networking` resyncs), it will not update the `issuedCertificate` and not cause the cert issuance workflow to restart
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/975


pkg/mesh-networking/translation/istio/mesh/mtls/mtls_translator.go
```diff
@@ -94,26 +108,98 @@ func (t *translator) Translate(
...
+		return t.configureSharedTrust(
+			mesh,
+			trustModel.Shared,
+			virtualMesh.Ref,
+			istioOutputs,
+			localOutputs,
+			mtlsConfig.AutoRestartPods,
+		)
+	case *v1alpha2.VirtualMeshSpec_MTLSConfig_Limited:
+		return eris.Errorf("limited trust not supportedin version %v of service mesh hub", version.Version)
```
harveyxia, line 144:
```suggestion
		return eris.Errorf("limited trust not supported in version %v of service mesh hub", version.Version)
```

pkg/mesh-networking/translation/istio/mesh/mtls/mtls_translator.go
```diff
@@ -168,28 +264,42 @@ func (t *translator) Translate(
...
+		Namespace: agentNamespace,
+		// write to the mesh cluster
+		ClusterName: istioMesh.GetInstallation().GetCluster(),
+		Labels:      metautils.TranslatedObjectLabels(),
+	}
+
 	// get the pods that need to be bounced for this mesh
-	podsToBounce := getPodsToBounce(mesh, t.workloads, mtlsConfig.AutoRestartPods)
+	podsToBounce := getPodsToBounce(mesh, t.workloads, autoRestartPods)
+	var (
```
harveyxia, line 278: Just curious, is it conventional to use `var ( )` blocks within functions vs. just repeating inline `var foo ...` declarations?

ilackarms: i think for a large list of vars it makes sense. i don't know that there are any well known conventions around this.

harveyxia (approved):

PR #990, opened 2020-10-05
Osm 0.4.0 test

ashleywang1:
Do not merge!


ci/setup-funcs.sh
```diff
@@ -276,7 +276,7 @@ function install_osm() {
   ROLLOUT="${K} rollout status deployment --timeout 300s"
 
   # install in permissive mode for testing
-  osm install
+  osm install --enable-permissive-traffic-policy
```
EItanya, line 279: In v0.3.0 this would break traffic splitting, as traffic splitting required restricted traffic policies

ashleywang1: Yup! Removed

PR #1001, opened 2020-10-08
Settings crd

harveyxia:


api/settings/v1alpha2/settings.proto
```diff
@@ -0,0 +1,27 @@
...
+message SettingsSpec {
+
+    // Configure mTLS settings for TrafficTargets.
+    MTLS mtls = 1;
+
+    // mTLS settings.
+    message MTLS {
+
+        // If true, by default enable mTLS for all TrafficTargets unless explicitly disabled by TrafficPolicies.
+        bool default_mtls = 1;
```
joekelley, line 18: should this be as simple as on/off, or should we also allow for simple TLS? https://istio.io/latest/docs/reference/config/networking/destination-rule/#ClientTLSSettings-TLSmode

harveyxia: Good question—we currently only use `ISTIO_MUTUAL` for mTLS. @ilackarms any thoughts?

harveyxia:
Closed in favor of https://github.com/solo-io/service-mesh-hub/pull/1004

api/settings/v1alpha2/settings.proto
harveyxia: Discussed offline, we'll import all of the Istio TLS modes, [reference](https://istio.io/latest/docs/reference/config/networking/destination-rule/#ClientTLSSettings-TLSmode)

PR #1004, opened 2020-10-09
Configure mtls

harveyxia:
Depends on https://github.com/solo-io/service-mesh-hub/pull/1001


pkg/common/bootstrap/bootstrap.go
```diff
@@ -36,7 +38,6 @@ type StartReconciler func(
 ) error
 
 // bootstrap options for starting discovery
-// TODO: wire these up to Settings CR
```
joekelley, line 39: nit: I interpreted this comment as "wire up MetricsBindPort, MasterNamespace, and VerboseMode to a settings CRD", which isn't addressed by this change.

api/settings/v1alpha2/settings.proto
```diff
@@ -0,0 +1,30 @@
...
+option (gogoproto.equal_all) = true;
+
+import "service-mesh-hub/api/networking/v1alpha2/validation_state.proto";
+import "service-mesh-hub/api/networking/v1alpha2/traffic_policy.proto";
+
+// Configure global settings and defaults.
+message SettingsSpec {
+
+    // Configure default mTLS settings for TrafficTargets (MTLS declared in TrafficPolicies take precedence)
+    .networking.smh.solo.io.TrafficPolicySpec.MTLS mtls = 1;
```
joekelley, line 15: I regret that this didn't occur to me when reviewing #1001 -- what do you think about having this default configured on the virtual mesh? Since a virtual mesh is a pre-requisite for federation, we're sure to have one when creating destination rules. Additionally, this way users will be able to specify default behavior on the VMesh level rather than globally, creating an opportunity for more granular control.

That way, semantically, the settings CRD can be used for controller settings (e.g. metrics bind port, watch namespaces, etc) rather than mesh configuration settings.

What do you think?

joekelley: Perhaps giving users the ability to specify a global default on the Settings CRD is sufficient for now and we could add an option at the virtual mesh in a follow up if demand arises. My concern with using the settings CRD to influence translation behavior is that we also have the outstanding TODO to wire the metrics bind address, write namespace, and other startup-time config to the settings CRD, and I want to avoid having the Settings CRD become a catch-all for high-level config options.

harveyxia: I share the same intuition here, that Settings isn't the best CRD for default MTLS settings, and that it would naturally belong under the VirtualMesh. However, DestinationRules are needed for translating subset routing config, regardless of whether the traffic target is grouped into a VirtualMesh (i.e. not every traffic target will have a VirtualMesh)

pkg/common/bootstrap/bootstrap.go
joekelley: discussed offline -- we probably don't want these values specified in the settings CRD. instead they can be options on the startup command and/or env vars.

install/helm/service-mesh-hub/templates/settings.yaml
```diff
@@ -0,0 +1,11 @@
...
+{{- $settings := $.Values.settings}}
+
+kind: Settings
+apiVersion: settings.smh.solo.io/v1alpha2
+metadata:
+  name: settings
+  namespace: {{ $.Release.Namespace }}
+spec:
+  mtls:
+    defaultMtls: {{ $settings.defaultMtls }}
```
joekelley: afaict this will be empty by default, which would be invalid. we should make sure there's a default in values.yaml

pkg/mesh-networking/translation/istio/decorators/outlierdetection/outlier_detection_decorator.go
```diff
@@ -85,3 +92,26 @@ func (d *outlierDetectionDecorator) translateOutlierDetection(
 		MaxEjectionPercent:    maxEjectionPercent,
 	}
 }
+
+func (d *outlierDetectionDecorator) translateTlsSettings(
+	trafficPolicy *v1alpha2.TrafficPolicySpec,
+) *networkingv1alpha3spec.ClientTLSSettings {
+	// If TrafficPolicy doesn't specify mTLS configuration, use global default populated upstream during initialization.
+	if trafficPolicy.Mtls == nil {
```
joekelley, line 100: we should also check that `trafficPolicy.Mtls.Istio` is non-nil

pkg/mesh-networking/translation/istio/decorators/outlierdetection/outlier_detection_decorator.go
```diff
@@ -85,3 +92,26 @@ func (d *outlierDetectionDecorator) translateOutlierDetection(
...
+}
+
+func MapIstioTlsMode(tlsMode v1alpha2.TrafficPolicySpec_MTLS_Istio_TLSmode) networkingv1alpha3spec.ClientTLSSettings_TLSmode {
+	switch tlsMode {
+	case v1alpha2.TrafficPolicySpec_MTLS_Istio_SIMPLE:
+		return networkingv1alpha3spec.ClientTLSSettings_SIMPLE
+	case v1alpha2.TrafficPolicySpec_MTLS_Istio_ISTIO_MUTUAL:
+		return networkingv1alpha3spec.ClientTLSSettings_ISTIO_MUTUAL
+	default:
+		return networkingv1alpha3spec.ClientTLSSettings_DISABLE
```
joekelley, line 115: wdyt about erroring if we fail to match on the tlsMode here so that we don't have an insecure default for TLS?

cmd/service-mesh-hub/main.go
```diff
@@ -33,6 +34,8 @@ func (opts *bootstrapOpts) addToFlags(flags *pflag.FlagSet) {
 	flags.StringVarP(&opts.MasterNamespace, "namespace", "n", metav1.NamespaceAll, "if specified restricts the master manager's cache to watch objects in the desired namespace.")
 	flags.Uint32Var(&opts.MetricsBindPort, "metrics-port", 9091, "port on which to serve Prometheus metrics. set to 0 to disable")
 	flags.BoolVar(&opts.VerboseMode, "verbose", true, "enables verbose/debug logging")
+	flags.StringVar(&opts.SettingsName, "settings-name", defaults.DefaultSettingsName, "The name of the Settings object this controller should use.")
```
ashleywang1, line 37: In gloo, we know for sure that its Settings object is named 'default' and is in the 'gloo-system'. Could we restrict the location of the Settings object here to the name 'default' in the 'service-mesh-hub' namespace?

pkg/mesh-networking/reconciliation/networking_reconciler.go
```diff
@@ -5,11 +5,15 @@ import (
 	"fmt"
 	"time"
 
+	settingsv1alpha2 "github.com/solo-io/service-mesh-hub/pkg/api/settings.smh.solo.io/v1alpha2"
```
ashleywang1, line 8: rename smhsettings to be consistent with import in 'smh_schemes.go'

harveyxia: Ah it's actually `smh_schemes.go` that's inconsistent with our convention, which is to to use `<smh><networking|discovery><version>` depending on which conflicts are present.

I changed `smh_schemes.go`

cmd/service-mesh-hub/main.go
harveyxia: The name/namespace of the Settings object is configurable in Gloo also, https://github.com/solo-io/gloo/blob/5c3bc06ca220dc407ce626d8e0893561312f151c/pkg/utils/setuputils/flags.go#L29-L30. The logic here follows the same convention, configurable with default name/namespace

api/settings/v1alpha2/settings.proto
antonioberben: SMH could copy the approach from Istio: A user specifies its configuration within its `IstioOperator` CR. Then, the controller creates a ConfigMap with the required Istio configuration which is then used by `istiod`.

So, in SMH, there is the `Mesh` CRD where the global settings/config could be attached. And the controller will create the confimap which the it is mounted by the discovery or networking pods.

ilackarms (approved):

PR #1005, opened 2020-10-09
Meshctl debug snapshot

ashleywang1:
```
meshctl debug snapshot --help
The output can be piped into a command like jq. For example:
meshctl debug snapshot discovery input | jq '.'

Usage:
  meshctl debug snapshot [flags]
  meshctl debug snapshot [command]

Available Commands:
  discovery   for the discovery pod
  networking  for the networking pod

Flags:
  -f, --file string   file to write output to
  -h, --help          help for snapshot
      --json          display the entire json snapshot The output can be piped into a command like jq. For example:
                       meshctl debug snapshot discovery input | jq '.'
      --zip string    zip file output

Use "meshctl debug snapshot [command] --help" for more information about a command.
```

The default output looks like:
```
{
  "kind": "configMaps",
  "value": [
    {
      "name": "cluster-autoscaler-status",
      "namespace": "kube-system",
      "cluster": "smh-mgmt"
    },
    {
      "name": "cluster-autoscaler-status",
      "namespace": "kube-system",
      "cluster": "smh-remote"
    },
    {
      "name": "coredns",
      "namespace": "istio-system",
      "cluster": "smh-mgmt"
    },
    {
      "name": "coredns",
      "namespace": "istio-system",
      "cluster": "smh-remote"
    },
...
```
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/992


harveyxia:
Would it be possible to output the snapshots to stdout? That way, users can pipe it to a utility of their choice for viewing/parsing

pkg/meshctl/commands/debug/snapshot/debug_snapshot.go
```diff
@@ -0,0 +1,60 @@
...
+
+func debugSnapshot(ctx context.Context) error {
+	freePort, err := cliutils.GetFreePort()
+	if err != nil {
+		fmt.Println(err.Error())
+		return err
+	}
+	localPort := strconv.Itoa(freePort)
+
+	for _, podName := range []string{"discovery", "networking"} {
```
harveyxia, line 33: Since these snapshots can be quite big, what do you think about parameterizing `discovery` and `networking` into another subcommand? E.g. `meshctl debug snapshot networking` would output just the networking snapshot

ashleywang1: `meshctl debug snapshot networking input` as well, since we differentiate input and output snapshots.

pkg/meshctl/commands/debug/snapshot/debug_snapshot.go
```diff
@@ -0,0 +1,208 @@
...
+	)
+	AddDebugSnapshotFlags(cmd.PersistentFlags(), opts)
+	return cmd
+}
+
+func Networking(ctx context.Context, opts *DebugSnapshotOpts) *cobra.Command {
+	pods := []string{networking}
+	cmd := &cobra.Command{
+		Use:   "networking",
+		Short: "for the networking pod",
```
jenshu, line 70:
```suggestion
		Short: "Input and output snapshots for the networking pod",
```

pkg/meshctl/commands/debug/snapshot/debug_snapshot.go
```diff
@@ -0,0 +1,208 @@
...
+		Output(ctx, opts, pods),
+	)
+	return cmd
+}
+
+func Discovery(ctx context.Context, opts *DebugSnapshotOpts) *cobra.Command {
+	pods := []string{discovery}
+	cmd := &cobra.Command{
+		Use:   "discovery",
+		Short: "for the discovery pod",
```
jenshu, line 83:
```suggestion
		Short: "Input and output snapshots for the discovery pod",
```

pkg/meshctl/commands/debug/snapshot/debug_snapshot.go
```diff
@@ -0,0 +1,208 @@
...
+		Input(ctx, opts, pods),
+		Output(ctx, opts, pods),
+	)
+	return cmd
+}
+
+func Input(ctx context.Context, opts *DebugSnapshotOpts, pods []string) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "input",
+		Short: "input snapshot",
```
jenshu, line 95: looks like `pods` is only ever set to one string, `discovery` or `networking`. can we make the `pods` arg a single string instead of array, and then we can set the `Short` description to something like `fmt.Sprintf("Input snapshot for the %s pod", pod)`

pkg/meshctl/commands/debug/snapshot/debug_snapshot.go
```diff
@@ -0,0 +1,208 @@
...
+			return debugSnapshot(ctx, opts, pods, []string{input})
+		},
+	}
+	return cmd
+}
+
+func Output(ctx context.Context, opts *DebugSnapshotOpts, pods []string) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "output",
+		Short: "output snapshot",
```
jenshu, line 106: same comment here as above

pkg/meshctl/commands/debug/snapshot/debug_snapshot.go
```diff
@@ -0,0 +1,208 @@
...
+			fileName := fmt.Sprintf("%s-%s-snapshot.json", podName, snapshotType)
+			var snapshotStr string
+			if opts.json {
+				snapshotStr = snapshot
+			} else {
+				err = ioutil.WriteFile(fileName, []byte(snapshot), filePermissions)
+				if err != nil {
+					return err
+				}
+				pipedCmd := "cat " + fileName + " | jq '" + query + "'"
```
jenshu, line 144: will this error if user doesn't have `jq` installed? (if so, we should add that as a prerequisite in the docs)

pkg/meshctl/commands/debug/snapshot/debug_snapshot.go
```diff
@@ -0,0 +1,208 @@
...
+	if opts.zip != "" {
+		err = zip(fs, dir, opts.zip)
+	}
+	return nil
+}
+
+func getSnapshot(ctx context.Context, localPort, podName, snapshotType string) string {
+	snapshot, portFwdCmd, err := cliutils.PortForwardGet(ctx,
+		"service-mesh-hub", "deploy/"+podName,
+		localPort, kubePort, false, "/snapshots/"+snapshotType)
```
jenshu, line 183: is this "service-mesh-hub" namespace configurable? can we read the value from somewhere, or have the user pass it in to the command?

pkg/meshctl/commands/debug/snapshot/debug_snapshot.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/solo-io/go-utils/debugutils"
+	"github.com/solo-io/go-utils/tarutils"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+const (
+	filePermissions = 0644
+	kubePort        = "9091"
```
jenshu, line 23: is this port configurable? can we read the value from somewhere, or have the user pass it in?

ashleywang1: I'll make both the namespace and port configurable, but hide it from the docs, since they are supposed to be debug features and the normal user shouldn't even be aware that they exist.

pkg/meshctl/commands/debug/snapshot/debug_snapshot.go
```diff
@@ -0,0 +1,226 @@
...
+		},
+	}
+	return cmd
+}
+
+func debugSnapshot(ctx context.Context, opts *DebugSnapshotOpts, pods, types []string) error {
+	// Check prerequisite jq is installed
+	_, err := exec.Command("which", "jq").Output()
+	if err != nil {
+		fmt.Printf("Error: Could not find jq! Please install it from https://stedolan.github.io/jq/download/\n")
```
jenshu, line 130: thinking about this again, i wonder if we can just fall back to `json.MarshalIndent(data, "", "  ")` if jq isn't installed, instead of requiring the user to install jq?

ashleywang1: We're also using jq to parse the data down into a smaller format! We won't be pretty-printing the entire JSON string because that would be too large.

pkg/meshctl/commands/debug/snapshot/debug_snapshot.go
```diff
@@ -0,0 +1,226 @@
...
+		Input(ctx, opts, discovery),
+		Output(ctx, opts, discovery),
+	)
+	return cmd
+}
+
+func Input(ctx context.Context, opts *DebugSnapshotOpts, pod string) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "input",
+		Short: fmt.Sprintf("input snapshot for the %s pod", pod),
```
jenshu, line 107: nit: casing  `input` -> `Input`

pkg/meshctl/commands/debug/snapshot/debug_snapshot.go
```diff
@@ -0,0 +1,226 @@
...
+			return debugSnapshot(ctx, opts, []string{pod}, []string{input})
+		},
+	}
+	return cmd
+}
+
+func Output(ctx context.Context, opts *DebugSnapshotOpts, pod string) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "output",
+		Short: fmt.Sprintf("output snapshot for the %s pod", pod),
```
jenshu, line 118: nit: casing  `output` -> `Output`

pkg/common/defaults/port.go
```diff
@@ -0,0 +1,3 @@
+package defaults
+
+var MetricsPort uint32 = 9091
```
jenshu, line 3: this can be a const

pkg/meshctl/commands/debug/snapshot/debug_snapshot.go
```diff
@@ -0,0 +1,226 @@
...
+	metricsBindPort uint32
+	namespace       string
+}
+
+func AddDebugSnapshotFlags(flags *pflag.FlagSet, opts *DebugSnapshotOpts) {
+	flags.BoolVar(&opts.json, "json", false, "display the entire json snapshot. The output can be piped into a command like jq (https://stedolan.github.io/jq/tutorial/). For example:\n meshctl debug snapshot discovery input | jq '.'")
+	flags.StringVarP(&opts.file, "file", "f", "", "file to write output to")
+	flags.StringVar(&opts.zip, "zip", "", "zip file output")
+	flags.BoolVar(&opts.verbose, "verbose", false, "enables verbose/debug logging")
+	flags.Uint32Var(&opts.metricsBindPort, "port", defaults.MetricsPort, "metrics port")
```
jenshu, line 51: there are a few other places in SMH code that use 9091, can you change those to use this new defaults const too?

jenshu (approved):

PR #1022, opened 2020-10-16
Remove links that don't work

ashleywang1:
We have some clever logic in solo-kit to figure out the correct doc link for Gloo, but unfortunately service-mesh-hub uses skv2 instead of solo-kit. This is a temporary workaround, but we definitely should figure out a way in skv2 to calculate the correct link from the proto file.


docs/content/reference/api/workload.md
```diff
@@ -86,9 +86,9 @@ Information describing a Kubernetes-based workload (e.g. a Deployment or DaemonS
 
 | Field | Type | Label | Description |
 | ----- | ---- | ----- | ----------- |
-| controller | [core.skv2.solo.io.ClusterObjectRef](#core.skv2.solo.io.ClusterObjectRef) |  | Resource ref to the underlying kubernetes controller which is managing the pods associated with the workloads. It has the generic name controller as it can represent a deployment, daemonset, or statefulset. |
-| podLabels | [][WorkloadSpec.KubernetesWorkload.PodLabelsEntry](#discovery.smh.solo.io.WorkloadSpec.KubernetesWorkload.PodLabelsEntry) | repeated | These are the labels directly from the pods that this controller owns. NB: these labels are read directly from the pod template metadata.labels defined in the workload spec. We need these to determine which services are backed by this workload. |
```
jenshu, line 90: this applies to all files: since there aren't any links, I wonder if we should list out the full package name for now, like we are doing for `core.skv2.solo.io.ClusterObjectRef` below

jenshu: (if we plan to fix the docs links soon, you can disregard this comment)

jenshu (approved):

PR #1019, opened 2020-10-15
Smh extensions: networking API

ilackarms:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/1018

PR #1036, opened 2020-10-23
Smh extensions impl

ilackarms:
based on #1019 
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/1018


pkg/mesh-networking/translation/istio/traffictarget/destinationrule/destination_rule_translator.go
```diff
@@ -176,7 +176,7 @@ func (t *translator) initializeDestinationRule(
 	}
 
 	// Initialize Istio TLS mode with default declared in Settings
-	istioTlsMode, err := tls.MapIstioTlsMode(mtlsDefault.Istio.TlsMode)
+	istioTlsMode, err := tls.MapIstioTlsMode(mtlsDefault.GetIstio().GetTlsMode())
```
harveyxia, line 179: Do we want the hardcoded default to be `Istio_DISABLE`? If so, why?

test/e2e/istio/README.md
```diff
@@ -0,0 +1,24 @@
+# Istio E2e Tests
```
harveyxia, line 1: Thank you for adding this!

pkg/common/utils/grpc/dial.go
```diff
@@ -0,0 +1,61 @@
...
+			grpc_retry.WithCodes(codes.Unavailable),
+			grpc_retry.WithMax(10),
+			grpc_retry.WithBackoff(grpc_retry.BackoffLinear(time.Second * 2)),
+		}
+		connectionBackoff := backoff.DefaultConfig
+		connectionBackoff.MaxDelay = time.Second * 2
+		opts = append(opts,
+			grpc.WithConnectParams(grpc.ConnectParams{Backoff: connectionBackoff}),
+			grpc.WithStreamInterceptor(grpc_middleware.ChainStreamClient(
+				//WithResetConnectionBackoffStream(),
```
harveyxia, line 45: Is this and L49 still needed?

pkg/mesh-networking/extensions/clients.go
```diff
@@ -0,0 +1,166 @@
...
+
+// Clients provides a convenience wrapper for a set of clients to communicate with multiple Extension Servers
+type Clients []v1alpha1.NetworkingExtensionsClient
+
+func NewClientsFromSettings(ctx context.Context, extensionsServerOptions []*v1alpha2.NetworkingExtensionsServer) (Clients, error) {
+	var extensionsClients Clients
+	for _, extensionsServer := range extensionsServerOptions {
+		extensionsServerAddr := extensionsServer.GetAddress()
+		if extensionsServerAddr == "" {
+			return nil, eris.Errorf("must specify extensions server address")
```
harveyxia, line 30: We should perform this validation in `pkg/common/settings/validate.go` and report it onto the Settings status

pkg/mesh-networking/extensions/clients.go
```diff
@@ -0,0 +1,166 @@
...
+		}
+		notification, err := notifications.Recv()
+		if err != nil {
+			return err
+		}
+		pushFn(notification)
+	}
+}
+
+// Clientset provides a handle to fetching a set of cahced gRPC clients
```
harveyxia, line 90:
```suggestion
// Clientset provides a handle to fetching a set of cached gRPC clients
```

pkg/mesh-networking/reconciliation/networking_reconciler.go
```diff
@@ -47,6 +52,13 @@ type networkingReconciler struct {
 	totalReconciles    int
 	verboseMode        bool
 	settingsRef        v1.ObjectRef
+	extensionClients   extensions.Clientset
+	reconciler         skinput.SingleClusterReconciler
+}
+
+// pushNotificationId is a special identifier for a reconcile event triggered by an extension server pushing a notification
+var pushNotificationId = &v1.ObjectRef{
+	Name: "push-notification-event",
```
harveyxia, line 61: It might be useful to identify the extension server instance with some sort of ID or name—we can possibly add that data to the `PushNotification` proto message

pkg/mesh-networking/translation/istio/traffictarget/destinationrule/destination_rule_translator.go
ilackarms: there's no way to tell between `Istio_DISABLE` being explicitly set by the user vs using the zero value because it's been left empty. to set a different default, we'd need to change the order of the enums in the proto.

all we're doing here is allowing a default (rather than erroring out) if the field is not provided.

long term i think the ideal default behavior is actually to skip creating the DestinationRule if the user does not explicitly specify an mTLS policy && no policies exist which require a DR to be created (like we do with the virtual services)

harveyxia: In the case where the user leaves the mtlsDefault field nil, this code will default to `Istio_DISABLE`, I was wondering whether we want that to default to something else

> long term i think the ideal default behavior is actually to skip creating the DestinationRule if the user does not explicitly specify an mTLS policy && no policies exist which require a DR to be created (like we do with the virtual services)

this makes sense to me, can you create an issue to track it?

pkg/mesh-networking/reconciliation/networking_reconciler.go
ilackarms: i think we can address this in a follow-up if need be. what do you think?

harveyxia (approved):

PR #1037, opened 2020-10-23
Write only debug logs when app mesh CRDs are not present

joekelley:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/1024


harveyxia (dismissed):

joekelley:
we're going to look into a more robust solution for this, where users can opt-in to warnings for particular mesh types on particular clusters

PR #1038, opened 2020-10-23
Add fqdns

ashleywang1:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/1035


api/discovery/v1alpha2/traffic_target.proto
```diff
@@ -68,6 +68,9 @@ message TrafficTargetStatus {
     // The set of Access Policies that have been applied to this TrafficTarget
     repeated AppliedAccessPolicy applied_access_policies = 4;
 
+    // the local and (if applicable) remote fully qualified domain names
+    repeated string fqdns = 5;
```
harveyxia, line 72: Maybe we should explicitly denote the local and remote FQDNs using separate fields, this removes any assumption that local FQDNs have a certain suffix (e.g. `.cluster.local`)

harveyxia (approved):

PR #1046, opened 2020-10-30
update selector docs

harveyxia:


api/networking/v1alpha2/selectors.proto
```diff
@@ -9,98 +9,45 @@ import "gogoproto/gogo.proto";
...
+        /*
+            If specified, all labels must exist on k8s Service.
+            When used in a networking policy, omission matches any labels.
+            When used in a Role, a wildcard `"*"` must be explicitly used to match any label key and/or value.
+        */
         map<string, string> labels = 1;
-        // If specified, match k8s Services if they exist in one of the specified namespaces. If not specified, match on any namespace.
+
+        /*
+            If specified, match k8s Services if they exist in one of the specified namespaces. If not specified, match on any namespace.
```
jenshu, line 31:
```suggestion
            If specified, match k8s Services if they exist in one of the specified namespaces.
```

jenshu: same for all other fields: i would remove the "If not specified, match on any..." from the first line, since the behavior differs between roles and policies

harveyxia: ah thanks for catching that, forgot to remove

jenshu (approved):

PR #1067, opened 2020-11-11
restore hack script for installing meshctl

joekelley:


hack/install.sh
```diff
@@ -0,0 +1,71 @@
...
+fi
+
+for smh_version in $GLOO_MESH_VERSIONS; do
+
+tmp=$(mktemp -d /tmp/smh.XXXXXX)
+filename="meshctl-${OS}-amd64"
+url="https://github.com/solo-io/gloo-mesh/releases/download/${smh_version}/${filename}"
+
+if curl -f ${url} >/dev/null 2>&1; then
+  echo "Attempting to download meshctl version ${smh_version}"
```
jenshu, line 29: nit: we should rename the `smh`

joekelley: good catch

PR #1070, opened 2020-11-11
Add autogenerated changelog (both open source and enterprise) to doc gen step

ryantking:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/1041


changelog/v0.10.3/generate_changelog.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+  - type: NEW_FEATURE
```
ryantking, line 2: Is this the correct type? Or should it be NON_USER_FACING?

harveyxia: This is correct—this is a new user facing feature

harveyxia: good question—we don't have a changelog type for docs, i think we've been inconsistently using NEW_FEATURE and FIX, nbd either way

docs/docsgen/generate_reference_docs.go
```diff
@@ -194,6 +273,86 @@ func collectDescriptors(protoDir, outDir string, filter func(file *model.Descrip
 	return gendoc.NewTemplate(protokit.ParseCodeGenRequest(golangRequest)), nil
 }
 
+func generateChangelog(root string, opts ChangelogOptions) error {
+	if strings.ToLower(os.Getenv("RELEASE")) != "true" {
```
harveyxia, line 277: It might be nice to also generate changelogs for `main`, so that we have an easy place to point at fixes that have made it into the `main` branch. @ryantking @joekelley what do you guys think?

joekelley: I like the idea, but I think we'd want to have a separate section on the changelog page that contains "staged but not yet available" release notes. That way users won't be met with a list of features they can't use yet when they first go to the changelog page

ryantking: I'm into it, will have to look at the complexity since the changelogs will have to come from a different source since there's no release

harveyxia (approved):

PR #1078, opened 2020-11-12
Add meshctl dashboard command

ryantking:
BOT NOTES: 
resolves https://github.com/solo-io/service-mesh-hub/issues/1053

### DIFF
--- a/pkg/meshctl/commands/dashboard/dashboard.go
+++ b/pkg/meshctl/commands/dashboard/dashboard.go
@@ -69,7 +69,7 @@ func forwardDashboard(ctx context.Context, kubeconfigPath, kubectx, namespace st
 	return nil
 }
 
-func getStaticPort(ctx context.Context, kubeconfigPath, kubectx, kubens string) (string, error) {
+func getStaticPort(ctx context.Context, kubeconfigPath, kubectx, namespace string) (string, error) {
 	cfg, err := kubeconfig.GetRestConfigWithContext(kubeconfigPath, kubectx, "")
 	if err != nil {
 		return "", err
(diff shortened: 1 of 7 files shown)


pkg/meshctl/commands/dashboard/dashboard.go
```diff
@@ -0,0 +1,157 @@
...
+type options struct {
+	kubeconfig  string
+	kubecontext string
+	kubens      string
+	port        uint32
+}
+
+func (o *options) addToFlags(flags *pflag.FlagSet) {
+	utils.AddManagementKubeconfigFlags(&o.kubeconfig, &o.kubecontext, flags)
+	flags.StringVar(&o.kubens, "kubes-namespace", "gloo-mesh", "The namespace that the Gloo Mesh UI is deployed in")
```
harveyxia, line 49: I would name this argument just `namespace` to stay consistent with our other commands

pkg/meshctl/commands/dashboard/dashboard.go
```diff
@@ -0,0 +1,157 @@
...
+	flags.Uint32VarP(&o.port, "port", "p", 8090, "The local port to forward to the dashboard")
+}
+
+var ConsoleNotFoundError = errors.New("Console image not found. Your Gloo Mesh enterprise install may be a bad state.")
```
harveyxia, line 53: nit: our convention is to place variable declarations towards the top of the file

pkg/meshctl/commands/dashboard/dashboard.go
```diff
@@ -0,0 +1,157 @@
...
+
+func forwardDashboard(ctx context.Context, kubeconfigPath, kubectx, kubens string, localPort uint32) error {
```
harveyxia, line 55: nit, since "namespace" implies k8s:

```suggestion
func forwardDashboard(ctx context.Context, kubeconfigPath, kubectx, namespace string, localPort uint32) error {
```

pkg/meshctl/commands/dashboard/dashboard.go
```diff
@@ -0,0 +1,157 @@
...
+			fmt.Printf("No Gloo Mesh dashboard found as part of the installation in namespace %s. "+
+				"The full dashboard is part of Gloo Mesh enterprise by default. ", kubens)
+		}
+
+		return "", err
+	}
+
+	var staticPort string
+	for _, container := range dep.Spec.Template.Spec.Containers {
+		if container.Name == "console" {
```
harveyxia, line 93: Are we confident that the container name and port name won't change? Is there a more robust / future-proof approach (for instance, if it's safer to assume that there will only ever be a single container with a single port for the UI deployment we could just grab the first container and first port found)

ryantking: There are currently 3 containers in the pod (console, API server, and envoy). Is adding a flag to specify container name acceptable?

harveyxia: Ah I see. I don't have strong opinions on this, I'm fine leaving the container and port hardcoded. @joekelley thoughts?

joekelley: This approach is fine with me

joekelley (approved):

PR #1080, opened 2020-11-13
Feat/add limited trust

### LINKED ISSUE #910: Add limited trust configuration capabilities to Service Mesh Hub
**Problem Statement**

The current SMH trust model implies changing the root certificate in all managed clusters. This is not a desired trust model, especially when dealing with multiple entities, each with it's own security requirements

**Proposal**

Each mesh should have it's own trust domain, and different TLS exchanges should happen at the cluster edge.

A possible solution would be configuring the ingress gateway in each cluster as an **edge proxy** that supports mTLS connection from the request originator, and initiates a new different mTLS connection to forward the request to the desired server.

This would help in the context where different clusters use different authentication protocols, and would limit the scope of any security change to the gateways.

Edge Proxy

Problems to solve:
- Managing certificates from multiple CA sources
- Managing different types of ingress gateways
- Passing/Persisting the configuration to the Ingress Gateways

*Steps*
- [ ] Propose implementation general design
- [ ] Discuss proposal with the community
- [ ] Start implementing this feature for Istio first (since Istio is already supported by SMH)

mimatache:
implement limited trust in `gloo-mesh`.

The scope is to enable cross cluster communication via Istio without changing the root trust domain in each cluster. This will allow each mesh to configure its own security, and we will provide another security domain for the inter cluster communication.

In order to switch from using the shared trust model to the limited trust model you need to apply the following configuration change:
```diff
  apiVersion: networking.smh.solo.io/v1alpha2
 kind: VirtualMesh
 metadata:
   name: virtual-mesh
   namespace: service-mesh-hub
 spec:
   mtlsConfig:
     autoRestartPods: true
-    shared:
+    limited:
       rootCertificateAuthority:
         generated: {}
   federation: {}
   meshes:
   - name: istiod-istio-system-mgmt-cluster
     namespace: service-mesh-hub
   - name: istiod-istio-system-remote-cluster
     namespace: service-mesh-hub
```

The existing shared trust certificate logic will generate the certificates for the inter cluster trust domain.

resolves #910


mimatache:
Will update the PR after fixing the merge conflicts.

PR #1081, opened 2020-11-13
Enterprise fixes 2

ilackarms:
includes 2 fixes:
- do not call extensions server during `applier` translation
- make ci scripts callable from another repo (enterpirse repo)

also makes the scripts re-runnable (will not exit if cluster / namespace creation fails). i can undo this if desired (but definitely helps with local development)

### DIFF
--- a/ci/setup-kind.sh
+++ b/ci/setup-kind.sh
@@ -50,18 +50,18 @@ else
 
   # NOTE(ilackarms): we run the setup_kind clusters sequentially due to this bug:
   # related: https://github.com/kubernetes-sigs/kind/issues/1596
-  create_kind_cluster ${mgmtCluster} 32001  || echo SKIPPED
+  create_kind_cluster ${mgmtCluster} 32001
   install_istio ${mgmtCluster} 32001 &
 
-  create_kind_cluster ${remoteCluster} 32000  || echo SKIPPED
+  create_kind_cluster ${remoteCluster} 32000
   install_istio ${remoteCluster} 32000 &
 
   wait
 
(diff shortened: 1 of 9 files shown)


ci/setup-kind.sh
```diff
@@ -48,18 +50,18 @@ else
 
   # NOTE(ilackarms): we run the setup_kind clusters sequentially due to this bug:
   # related: https://github.com/kubernetes-sigs/kind/issues/1596
-  create_kind_cluster ${mgmtCluster} 32001
+  create_kind_cluster ${mgmtCluster} 32001  || echo SKIPPED
```
harveyxia, line 53: To make this extra safe, we can check for the specific error string `ERROR: node(s) already exist for a cluster with the name "mgmt-cluster"`

ci/setup-kind.sh
```diff
@@ -48,18 +50,18 @@ else
...
 
-  create_kind_cluster ${remoteCluster} 32000
+  create_kind_cluster ${remoteCluster} 32000  || echo SKIPPED
   install_istio ${remoteCluster} 32000 &
 
   wait
 
   # create istio-injectable namespace
-  kubectl --context kind-${mgmtCluster} create namespace bookinfo
+  kubectl --context kind-${mgmtCluster} create namespace bookinfo  || echo SKIPPED
```
harveyxia, line 62: Same comment as above, we can check for the error `Error from server (AlreadyExists): namespaces "bookinfo" already exists`

harveyxia (approved):

Issue #1020, opened 2020-10-16
Add meshctl install enterprise


joekelley:
Add an `install enterprise` command to allow users to install Service Mesh Hub Enterprise via meshctl. The command should have a user experience similar to that of `glooctl install gateway enterprise` (i.e. provide a license key flag) and be backed by the chart published via https://github.com/solo-io/service-mesh-hub-enterprise-helm.

PR #1083, opened 2020-11-16
Add meshctl install enterprise command

ryantking:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1020

### DIFF
--- a/pkg/meshctl/commands/install/enterprise/enterprise.go
+++ b/pkg/meshctl/commands/install/enterprise/enterprise.go
@@ -2,10 +2,6 @@ package enterprise
 
 import (
 	"context"
-	"fmt"
-	"io"
-	"io/ioutil"
-	"net/http"
 
 	"github.com/rotisserie/eris"
 	"github.com/solo-io/gloo-mesh/pkg/meshctl/commands/install/internal/flags"
@@ -14,7 +10,6 @@ import (
 	"github.com/solo-io/gloo-mesh/pkg/meshctl/registration"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-	"gopkg.in/yaml.v2"
 )
 
 func Command(ctx context.Context) *cobra.Command {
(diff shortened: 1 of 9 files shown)


pkg/meshctl/commands/install/enterprise/enterprise.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+type options struct {
+	flags.Options
+	licenseKey string
+}
+
+func (o *options) addToFlags(flags *pflag.FlagSet) {
+	o.AddToFlags(flags)
+	flags.StringVar(&o.licenseKey, "license", "", "Enterprise license key")
```
harveyxia, line 36: We can additionally use `cobra.MarkFlagRequired()` here to ensure that the user invokes the command with the flag

pkg/meshctl/commands/install/enterprise/enterprise.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+var NoLicenseError = errors.New("Gloo Mesh Enterprise requries a license key.")
```
harveyxia, line 39: nit: move var declarations to top of files

pkg/meshctl/commands/install/enterprise/enterprise.go
```diff
@@ -0,0 +1,68 @@
...
+	installer.Values["license.key"] = opts.licenseKey
+	if err := installer.InstallGlooMeshEnterprise(ctx); err != nil {
+		return eris.Wrap(err, "installing gloo-mesh-enterprise")
+	}
+	if err := installer.InstallGlooMeshUI(ctx); err != nil {
+		return eris.Wrap(err, "installing gloo-mesh-ui")
+	}
+	if err := installer.InstallRbacWebHook(ctx); err != nil {
+		return eris.Wrap(err, "installing rbac-webhook")
+	}
```
harveyxia, line 55: The enterprise helm chart already contains the gloo-mesh UI, and it's not there now but should be updated to also include the RBAC webhook, so we can remove this block

harveyxia (commented):
Looks really good! Clean and clear approach for refactoring the install code to support OSS and enterprise installation

pkg/meshctl/commands/install/install.go
```diff
@@ -26,100 +20,20 @@ func Command(ctx context.Context) *cobra.Command {
...
-	gloomeshVersion := opts.version
-	if opts.version == "" {
-		gloomeshVersion = version.Version
-	}
-	if gloomeshChartUri == "" {
-		gloomeshChartUri = fmt.Sprintf(gloomesh.GlooMeshChartUriTemplate, gloomeshVersion)
-	}
-
-	err := gloomesh.Installer{
-		HelmChartPath:  gloomeshChartUri,
```
harveyxia, line 87: You missed this field when refactoring, and that's why the [e2e tests are failing](https://github.com/solo-io/gloo-mesh/pull/1083/checks?check_run_id=1407696484#step:10:425)

ryantking: I originally left this out because it was overriding the enterprise helm chart path. Will make this a param for the function so the logic can be performed by the caller

pkg/meshctl/commands/install/internal/flags/flags.go
```diff
@@ -51,9 +54,16 @@ func (o *Options) AddToFlags(flags *pflag.FlagSet) {
...
 }
 
-func (o *Options) GetInstaller(chartPath string) gloomesh.Installer {
+func (o *Options) GetInstaller(chartUriTemplate string) gloomesh.Installer {
+	// User-specified chartPath takes precedence over specified version.
+	if o.Version == "" {
+		o.Version = version.Version
+	}
+	if o.ChartPath == "" {
+		o.ChartPath = fmt.Sprintf(gloomesh.GlooMeshEnterpriseChartUriTemplate, o.Version)
```
harveyxia, line 63:
```suggestion
		o.ChartPath = fmt.Sprintf(chartUriTemplate, o.Version)
```

harveyxia (approved):

Issue #872, opened 2020-08-17
repor Write errors to the input resources that manage the output resource


ilackarms:
**Is your feature request related to a problem? Please describe.**

When a user config object is invalid, its status is updated with an error. However, if a user config object is valid but writing the translated object fails, the error is only logged and not reported back to the crd status.

**Describe the solution you'd like**

Write errors should be written back to the status of each CRD that was responsible for writing the translated object.

**Additional context**

The `output.Snapshot.Apply()` methods accept an `errHandler` which can be used to handle write errors when individual output objects fail to write to the cluster: https://github.com/solo-io/service-mesh-hub/blob/merge-repos/pkg/api/networking.smh.solo.io/output/snapshot.go#L62

ryantking:
Adding some more information here after diving into this with @harveyxia and @ilackarms.

Two things need to happen in order to get the desired functionality.

1. Parent resources to a resource need to be tracked. This can be done in memory, but doing it via an annotation is more desirable since that information will be useful in other circumstances.
2. Create an error handler that can find the parent(s) in the input snapshot and update their statuses with any error.

PR #1087, opened 2020-11-18
Write network errors to input resources

ryantking:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/872


pkg/mesh-networking/reconciliation/errhandler.go
```diff
@@ -0,0 +1,38 @@
...
+		r.Status.Errors = append(r.Status.Errors, e.AppendingErrHandler.Errors().Error())
+	case *v1alpha2.FailoverService:
+		r.Status.Errors = append(r.Status.Errors, e.AppendingErrHandler.Errors().Error())
+	case *v1alpha2.TrafficPolicy:
+		r.Status.Errors = append(r.Status.Errors, e.AppendingErrHandler.Errors().Error())
+	case *v1alpha2.AccessPolicy:
+		r.Status.Errors = append(r.Status.Errors, e.AppendingErrHandler.Errors().Error())
+	}
+
+	if _, err := controllerutils.UpdateStatus(context.TODO(), e.client, resource); err != nil {
```
harveyxia, line 35: `inputSnap.SyncStatuses` is [invoked later in the code path](https://github.com/solo-io/gloo-mesh/blob/310f802e998c4f5cf571d0d8007bedc2e7dd60fe/pkg/mesh-networking/reconciliation/networking_reconciler.go#L153), which will perform the write. So we should only need to update the status in memory here

pkg/mesh-networking/reconciliation/errhandler.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+var _ output.ErrorHandler = errHandler{}
+
+type errHandler struct {
+	client client.Client
+	errhandlers.AppendingErrHandler
+}
+
+func (e errHandler) HandleWriteError(resource ezkube.Object, err error) {
```
harveyxia, line 21: As written, this will collect and return errors for all input resources. I think we can implement a new `ErrorHandler` entirely, one that simply appends the error to the specific associated input resource's `Status.Errors` like you have written below

harveyxia (approved):

harveyxia (dismissed):
trying to revert the accidental approval

pkg/mesh-networking/reconciliation/errhandler.go
```diff
@@ -0,0 +1,98 @@
...
+	annotations := resource.GetAnnotations()
+	parentsStr, ok := annotations[metautils.ParentLabelkey]
+	if !ok {
+		return
+	}
+	var allParents map[string][]*v1.ObjectRef
+	if err := json.Unmarshal([]byte(parentsStr), &allParents); err != nil {
+		return
```
harveyxia, line 29: Can we surface this error somehow, maybe through logging?

pkg/mesh-networking/reconciliation/errhandler.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	for gvk, parents := range allParents {
+		switch gvk {
+		case discoveryv1alpha2.Mesh{}.GVK().String():
```
harveyxia, line 34: The only parent CRDs are the networking types, we shouldn't need to handle discovery types here

pkg/mesh-networking/reconciliation/networking_reconciler.go
```diff
@@ -169,7 +168,7 @@ func (r *networkingReconciler) applyTranslation(ctx context.Context, in input.Sn
 		return err
 	}
 
-	errHandler := errhandlers.AppendingErrHandler{}
+	errHandler := errHandler{}
```
harveyxia, line 171: nit: maybe create a constructor? would make it slightly easier to extend if the error handler requires other dependencies

pkg/mesh-networking/translation/istio/mesh/access/access_policy_translator.go
```diff
@@ -54,8 +54,8 @@ func (t *translator) Translate(
 	}
 	clusterName := istioMesh.Installation.Cluster
 	installationNamespace := istioMesh.Installation.Namespace
-	globalAuthPolicy := buildGlobalAuthPolicy(installationNamespace, clusterName)
-	ingressGatewayAuthPolicies := buildAuthPoliciesForIngressgateways(installationNamespace, clusterName, istioMesh.IngressGateways)
+	globalAuthPolicy := buildGlobalAuthPolicy(installationNamespace, clusterName, mesh)
```
harveyxia, line 57: The VirtualMesh, not the Mesh, is the parent for the Istio AuthorizationPolicies

pkg/mesh-networking/translation/istio/mesh/failoverservice/failover_service_translator.go
```diff
@@ -132,14 +132,26 @@ func (t *translator) translate(
 		if err != nil {
 			errsForMesh = multierror.Append(errsForMesh, err)
 		}
+		if err := metautils.AppendParent(serviceEntry, mesh, mesh.GVK()); err != nil {
```
harveyxia, line 135: The FailoverService is the parent

pkg/mesh-networking/translation/istio/mesh/federation/federation_translator.go
```diff
@@ -219,6 +219,10 @@ func (t *translator) Translate(
 					Ports:      ports,
 				},
 			}
+			if err := metautils.AppendParent(se, ref, mesh.GVK()); err != nil {
```
harveyxia, line 222: VirtualMesh is the parent

pkg/mesh-networking/translation/istio/mesh/federation/federation_translator.go
```diff
@@ -219,6 +219,10 @@ func (t *translator) Translate(
...
+				// TODO(ryantking): Handle error
+				return
+			}
 
 			// Translate VirtualServices for federated TrafficTargets
 			vs := t.virtualServiceTranslator.Translate(in, trafficTarget, clientIstio.Installation, reporter)
```
harveyxia, line 228: We also need to invoke `AppendParent` for VirtualServices and DestinationRules. More generally, to any resource being added to the output snapshot

pkg/mesh-networking/translation/istio/mesh/mtls/mtls_translator.go
```diff
@@ -299,7 +306,13 @@ func (t *translator) constructIssuedCertificate(
 			IssuedCertificateSecret:  istioCaCerts,
 			PodBounceDirective:       podBounceRef,
 		},
-	}, podBounceDirective
+	}
+	if err := metautils.AppendParent(issuedCert, rootCaSecret, discoveryv1alpha2.Mesh{}.GVK()); err != nil {
```
harveyxia, line 310: 1. VirtualMesh is the parent
2. Also need to append parents for PodBounceDirective
3. Might be easier to read to place the `AppendParent` invocations next to L183-184 where we add issuedCerts and podBounceDirectives to the istio output snapshot

pkg/mesh-networking/translation/istio/traffictarget/authorizationpolicy/authorization_policy_translator.go
```diff
@@ -74,6 +74,11 @@ func (t *translator) Translate(
 		authPolicy.Spec.Rules = append(authPolicy.Spec.Rules, rule)
 	}
 
+	if err := metautils.AppendParent(authPolicy, trafficTarget, trafficTarget.GVK()); err != nil {
```
harveyxia, line 77: We should invoke this in the [Istio TrafficTargetTranslator](https://github.com/solo-io/gloo-mesh/blob/d619caabbe7838a4141ca04cca1eab323bc337f0/pkg/mesh-networking/translation/istio/traffictarget/istio_traffic_target_translator.go#L76), same comment for DestinationRuleTranslator and VirtualServiceTranlator

pkg/mesh-networking/translation/smi/traffictarget/access/traffic_target_translator.go
```diff
@@ -119,6 +119,10 @@ func (t *translator) Translate(
 				},
 			},
 		}
+		if err := metautils.AppendParent(trafficTarget, target, target.GVK()); err != nil {
```
harveyxia, line 122: Similar comment as above, these should be invoked in the top level [SMI traffic target translator](https://github.com/solo-io/gloo-mesh/blob/5d3ea22d594b3fbeb83ceb930f0c4459fedeadc6/pkg/mesh-networking/translation/smi/traffictarget/smi_traffic_target_translator.go#L55)

pkg/mesh-networking/reconciliation/errhandler.go
ryantking: The logging I've seen invoked uses `contextutils.LoggerFrom(ctx).Etc`, but there is no context here, should I pass it in, or are there other logging options?

I would also like to use this to handle the errors in `AppendParent` so that way that function does not need to return an error.

pkg/mesh-networking/reconciliation/errhandler.go
```diff
@@ -1,81 +1,87 @@
...
 	"github.com/solo-io/skv2/pkg/ezkube"
 )
 
-var _ output.ErrorHandler = errHandler{}
+var _ output.ErrorHandler = &errHandler{}
 
 type errHandler struct {
-	inp input.Snapshot
+	ctx  context.Context
+	inp  input.Snapshot
```
harveyxia, line 22: nit: we've been using `in` as shorthand for input snaphsot

pkg/mesh-networking/reconciliation/errhandler.go
```diff
@@ -85,14 +91,6 @@ func (e errHandler) HandleWriteError(resource ezkube.Object, err error) {
...
-}
-
-func (e errHandler) HandleListError(err error) {
-	panic("unimplemented")
-}
-
-func (e errHandler) Errors() error {
-	return nil
+func (e *errHandler) Errors() error {
+	return e.errs
```
harveyxia, line 95:
```suggestion
	return e.errs.ErrorOrNil()
```

We need to invoke this to safely convert to an `error` interface

pkg/mesh-networking/translation/istio/mesh/access/access_policy_translator.go
```diff
@@ -88,10 +102,10 @@ func buildAuthPoliciesForIngressgateways(
 				},
 			},
 		}
-		if err := metautils.AppendParent(ap, mesh, mesh.GVK()); err != nil {
-			// TODO(ryantking): Handle error
-			return nil
-		}
+
+		// Append the virtual mesh as a parent to each output access policy
```
harveyxia, line 106: I would move this logic to L70 to group it with the other `AppendParent` invocation for easier reading, and colocate it next to `outputs.AddAuthorizationPolicies`. This would also prevent us from having to extend the function signature of this method.

pkg/mesh-networking/translation/istio/mesh/failoverservice/failover_service_translator.go
```diff
@@ -132,35 +132,32 @@ func (t *translator) translate(
...
+		metautils.AppendParent(t.ctx, serviceEntry, failoverService.GetRef(), v1alpha2.FailoverService{}.GVK())
+		metautils.AppendParent(t.ctx, envoyFilter, failoverService.GetRef(), v1alpha2.FailoverService{}.GVK())
+
 		serviceEntries = append(serviceEntries, serviceEntry)
-		destinationRules = append(destinationRules, destinationRule)
 		envoyFilters = append(envoyFilters, envoyFilter)
+
+		// add destination rule if necessary
+		if destinationRule != nil {
+			metautils.AppendParent(t.ctx, destinationRule, failoverService.GetRef(), v1alpha2.FailoverService{}.GVK())
```
harveyxia, line 158: 1. There are other places in translation where a translator may return a nil object if that object has no configuration effect, so we should include the nil check in `AppendParent` itself (just return if the obj is nil)

2. Same comment as above, though it would incur additional loops, I think it would be clearer to invoke `AppendParent` next to invocations of `output.Add...` since this can be considered a post-processing step that's not critical to the main configuration translation.

pkg/mesh-networking/translation/istio/mesh/mtls/mtls_translator.go
```diff
@@ -180,6 +180,11 @@ func (t *translator) configureSharedTrust(
 		agentInfo.AgentNamespace,
 		autoRestartPods,
 	)
+
+	// Append the virtual mesh as a parent to the output resources
+	metautils.AppendParent(t.ctx, issuedCertificate, virtualMeshRef, v1alpha2.VirtualMesh{}.GVK())
+	metautils.AppendParent(t.ctx, podBounceDirective, virtualMeshRef, v1alpha2.VirtualMesh{}.GVK())
+
```
harveyxia, line 187: We should also add a parent annotation to the secret created on L236

pkg/mesh-networking/reconciliation/errhandler.go
```diff
@@ -0,0 +1,96 @@
...
+	if err := json.Unmarshal([]byte(parentsStr), &allParents); err != nil {
+		contextutils.LoggerFrom(e.ctx).Errorf("internal error: could not unmarshal %q annotation", metautils.ParentLabelkey)
+		return
+	}
+
+	for gvk, parents := range allParents {
+		switch gvk {
+		case v1alpha2.VirtualMesh{}.GVK().String():
+			for _, parentVMesh := range parents {
+				for _, vmesh := range e.in.VirtualMeshes().List(func(vmesh *v1alpha2.VirtualMesh) bool {
```
ilackarms, line 60: instead of iterating through the `List`, we can use the `Find` function

ryantking: Do we need to do anything with the returned error from `Find`, or can we just check if its `nil` then append the error?

ilackarms (dismissed):

ilackarms (dismissed):

ilackarms (approved):

PR #1099, opened 2020-11-25
Istio ingress discovery config

jenshu:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1094


api/settings/v1alpha2/settings.proto
```diff
@@ -18,9 +18,36 @@ message SettingsSpec {
...
     repeated NetworkingExtensionsServer networking_extension_servers = 2;
+
+    // Istio-specific discovery settings
+    Istio istio = 3;
+
+    message Istio {
+        // Workload labels and TLS port name used during discovery to detect ingress gateways for a mesh.
+        // If not specified, will fall back to the default values specified by `IngressGatewayMatcher`.
+        // To override the labels and/or TLS ports for a specific cluster, use `ingressGatewayMatcherOverrides`.
+        IngressGatewayMatcher ingress_gateway_matcher = 1;
```
ilackarms, line 29: instead of `matcher`, how about `detector`? i think it lines up with our semantics (and internal language) a bit more nicely. matchers elsewhere in gloo-mesh refer to matching requests

api/settings/v1alpha2/settings.proto
```diff
@@ -18,9 +18,36 @@ message SettingsSpec {
...
+
+        // Override the ingress gateway matchers on a per-cluster basis. The key to the map is a k8s cluster name.
+        // If an entry is found for a given cluster, it will be used. Otherwise, will fall back to `ingressGatewayMatcher`.
+        // If the cluster's `IngressGatewayMatcher` entry does not specify all the fields, it will fall back to the
+        // `ingressGatewayMatcher` values for the missing field(s).
+        map<string, IngressGatewayMatcher> ingress_gateway_matcher_overrides = 3;
```
ilackarms, line 35: instead of providing 2 fields, one with a defualt and one with cluster specific options, what about only providing the map field, and permitting a special-case key `*` to represent the default option?

jenshu:
had merge issues. will start a new PR

PR #1136, opened 2020-12-10
Make istio ingress gateway discovery configurable

jenshu:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1094


api/settings/v1alpha2/settings.proto
```diff
@@ -18,9 +18,33 @@ message SettingsSpec {
...
+
+    // Istio-specific discovery settings
+    Istio istio = 3;
+
+    message Istio {
+        // Ingress gateway detectors for each cluster. The key to the map is either a k8s cluster name or the wildcard
+        // `*` meaning all clusters. If an entry is found for a given cluster, it will be used. Otherwise, the
+        // wildcard entry will be used if it exists. Lastly, we will fall back to the default values.
+        // If any entry does not specify all the fields, it will fall back to the wildcard or default values for
+        // the missing field(s).
```
harveyxia, line 30: up for debate, but I feel this is a bit redundant since default values are discussed below

pkg/common/defaults/istio.go
```diff
@@ -6,3 +6,8 @@ package defaults
 var DefaultGatewayWorkloadLabels = map[string]string{
 	"istio": "ingressgateway",
 }
+
+const (
+	// https://istio.io/docs/ops/deployment/requirements/#ports-used-by-istio
```
harveyxia, line 11: might be worth specifying "As of istio version ...." since their doc links don't explicitly include the version

pkg/mesh-discovery/translation/discovery_translator.go
```diff
@@ -15,7 +15,7 @@ var DefaultDependencyFactory = internal.DependencyFactoryImpl{}
 // the translator "reconciles the entire state of the world"
 type Translator interface {
 	// translates the Input Snapshot to an Output Snapshot
-	Translate(ctx context.Context, in input.RemoteSnapshot) (discovery.Snapshot, error)
+	Translate(ctx context.Context, inRemote input.RemoteSnapshot, inLocal input.LocalSnapshot) (discovery.Snapshot, error)
```
harveyxia, line 18: Applying the principle of least knowledge, can we pass down the settings object instead of the entire local snapshot? This way we can:

1. Detect errors while fetching the singleton settings object earlier in the call stack
2. Clarifies the function signatures by explicitly indicating that the settings object is needed

harveyxia (approved):

PR #1174, opened 2020-12-21
ignore Helm missing CRD error

harveyxia:


pkg/meshctl/install/helm/installer.go
```diff
@@ -110,11 +110,12 @@ func (i Installer) InstallChart(ctx context.Context) error {
 		client.DryRun = dryRun
 
 		release, err := client.Run(chartObj, parsedValues)
-		if err != nil {
+		// ignore missing CRD error due to known Helm limitation, https://github.com/helm/helm/issues/7449
+		if err != nil && !strings.Contains(err.Error(), "unable to recognize \"\": no matches for kind") {
```
jenshu, line 114:
```suggestion
		if err != nil && !(dryRun && strings.Contains(err.Error(), "unable to recognize \"\": no matches for kind")) {
```
should we check for dryRun?

harveyxia: good call!

jenshu (approved):

Issue #1167, opened 2020-12-18
Istio mtls translator nil-check


harveyxia:
Nil dereference happens [here in the mtls translator](https://github.com/solo-io/gloo-mesh/blob/8c7ec10557074e9c185afed540a833865a30cf0e/pkg/mesh-networking/translation/istio/mesh/mtls/mtls_translator.go#L186) because the [pod bounce directive](https://github.com/solo-io/gloo-mesh/blob/8c7ec10557074e9c185afed540a833865a30cf0e/pkg/mesh-networking/translation/istio/mesh/mtls/mtls_translator.go#L288) can be nil if there aren't any pods that need bouncing.

PR #1185, opened 2020-12-30
extract DiscoverySettings to a separate message

yuval-k:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1186


changelog/v0.11.4/discovery-setting.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+  - type: FIX
```
joekelley, line 2: Let's mark this as breaking since it restructures the settings API

pkg/mesh-discovery/agent/discovery_reconciler.go
```diff
@@ -0,0 +1,191 @@
+package agent
```
joekelley, line 1: This file is unused and appears to be out of scope for this PR

changelog/v0.11.4/discovery-setting.yaml
```diff
@@ -0,0 +1,4 @@
...
+    issueLink: https://github.com/solo-io/gloo-mesh-enterprise/issues/98
```
joekelley, line 3: could you create and link to an issue in this repo instead?

pkg/mesh-discovery/agent/discovery_reconciler.go
yuval-k: right! copy paste error

changelog/v0.11.4/discovery-setting.yaml
yuval-k: will do. what's the advantage of doing so?

ilackarms (dismissed):

changelog/v0.11.4/discovery-setting.yaml
joekelley: Thanks! This way non-solo users will be able to access a working issue link in the generated release notes, as in `#1167` here https://github.com/solo-io/gloo-mesh/releases/tag/v0.11.2.

Not very impactful in this case, but some issues might have extensive discussion that's worth referencing.

joekelley (approved):

Issue #1182, opened 2020-12-28
Changelogs not appearing on "main" docs site


joekelley:
See https://docs.solo.io/gloo-mesh/latest/reference/changelog/open_source/ -- "latest" changelogs aren't rendered, but I'm able to view changelogs for all other served docs docs versions (main, 0.10.10, etc).

ashleywang1:
There are no changelogs past a certain point on:

<img width="452" alt="Screen Shot 2021-08-24 at 11 09 31 AM" src="https://user-images.githubusercontent.com/4549947/130642081-37c4a942-7dba-46f4-ae37-f235d9a529cb.png">
https://docs.solo.io/gloo-mesh/main/reference/changelog/

<img width="823" alt="Screen Shot 2021-08-24 at 11 09 09 AM" src="https://user-images.githubusercontent.com/4549947/130642014-8c3fa18d-8dfe-4b9c-bfc0-7e6bee6b9866.png">

Issue #545, opened 2020-04-08
Docs release process (latest)


EItanya:
When a new release is cut (e.g. 0.7.1 -> 0.7.2) an additional commit must be merged to package 0.7.2 docs. See https://github.com/solo-io/service-mesh-hub/commit/c5c3c19a66d574e769cf32bb913e40368a2f1cce#diff-366e6256bfcf6bf1aae5ee7b9e343475

We should automate this process such that versioned docs are updated on release.

joekelley:
Generate the `version.json` file in the docsgen step; have it read the latest patch of all minors starting with some fixed version (maybe 0.7.x). 

We should be mindful of when the script runs, whether that be on commit or on release.

PR #1188, opened 2021-01-04
Better doc release process

ryantking:
Solves two issues:
- Remove the need for hardcoded versions.json file, use git tags to build the list (greatest patch for each minor version past a given minimum version, 0.7.x for now) such that once a commit to main or release happens, the doc site is immediately updated.
- Uses the local changelog folder instead of the GitHub release body to build the changelog to solve a race condition where the GH release body wasn't updated until after docs were published, which previously required a second commit to master.

BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1182
resolves https://github.com/solo-io/gloo-mesh/issues/545


kdorosh:
Food for thought: one thing I'll note as a positive of using today's behavior (github API as source of truth) is that we can make manual edits to releases that already went out.

For example, Gloo Edge has had releases such as v1.6.0-beta10 (see in [docs](https://docs.solo.io/gloo-edge/latest/reference/changelog/enterprise/) generated from GH API) that had regressions. We could easily update the release notes by hand and call attention to it up front.

To do this and properly call attention up front while using local changelog as source of truth, may require some work in our changelog parsing utils and thus also an update to our changelog-bot

joekelley:
Can you share a screenshot of how the unreleased section of the changelog will render?

ryantking:
@joekelley It just shows the unreleaed versions:

main branch changelog

ryantking:
After going over this in more detail and some of the implications originally not seen, we are opting to close this for now and use the original GitHub release API style of changelog generation. We are keeping this branch around as a reference for when we revisit changelog generation at a later date.

PR #1191, opened 2021-01-06
Fix changelog gneeration for latest

ryantking:
Fixes the issue with the latest site being blank by disabling docs publication on releases, a subsequent commit to `main` to update the `versions.json` will be needed to trigger it.

Also makes a tweak where the docsgen script does not use flags. The use of flags was a poor design choice (by me) originally because the docs sites for old versions are built by checking out a tag then running the script from that tag, so if flags were later changed in the `ci/build-docs-site.sh` script, then that would break old versions. The new change uses an environment variable for skipping changelog generation and parses the git repo for version info instead of requiring flags.

BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1182


codegen/docs/docsgen.go
```diff
@@ -11,17 +10,7 @@ import (
 	"github.com/solo-io/gloo-mesh/pkg/meshctl/commands"
 )
 
-//go:generate go run docsgen.go
```
ryantking, line 14: @joekelley This `go:generate` is redundant with the `make generated-reference-docs` routine, which is part of `make generated-doe` so I removed it. Does this make sense to you?

kdorosh (commented):
lgtm

joekelley (approved):

PR #1200, opened 2021-01-11
Flat network api

EItanya:


api/discovery/v1alpha2/traffic_target.proto
```diff
@@ -53,6 +55,17 @@ message TrafficTargetSpec {
...
+        // A series of IP addresses and their associated ports.
+        // The list of ip + port pairs is the cartesian product of the following lists
+        message EndpointSubsets {
+
+            repeated string ip_addresses = 1;
+
+            repeated KubeServicePort ports = 2;
+        }
+
+        repeated EndpointSubsets endpoints = 6;
```
ilackarms, line 67: what Subset does mean in this context?

api/networking/v1alpha2/virtual_mesh.proto
```diff
@@ -125,6 +125,8 @@ message VirtualMeshSpec {
             // in that Virtual Mesh.
             .google.protobuf.Empty permissive = 1;
         };
+
+        bool flat_network = 2;
```
ilackarms, line 129: needs comment

api/networking/v1alpha2/virtual_mesh.proto
```diff
@@ -125,6 +125,12 @@ message VirtualMeshSpec {
             // in that Virtual Mesh.
             .google.protobuf.Empty permissive = 1;
         };
+
+        // Whether or not the virtual mesh is part of a flat network or not. If true, then all multicluster
+        // traffic will be routed directly to the Service Endpoints of the target workloads, rather than
+        // routing through an ingress gateway.
+        // NOTE: This feature will not work if the clusters are not pre-configured to live on the same network.
```
harveyxia, line 132:
```suggestion
        // If true, all multicluster traffic will be routed directly to the Service Endpoints of the traffic targets, 
        // rather than through an ingress gateway.
        // NOTE: This feature will not work if the clusters are not pre-configured to live on the same network.
```

api/discovery/v1alpha2/traffic_target.proto
harveyxia: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#endpointsubset-v1-core

ilackarms (approved):

PR #1204, opened 2021-01-13
Networking with custom overrides

ilackarms:
the purpose of this PR is to expose the networking controller as a module that can be imported to go projects and injected with custom behavior.

currently we are seeking to make the following possible to override:

- how networking's reconcilers get called (e.g. not necessarily on kubernetes storage events)
- how networking's outputs get written (not necessarily to a kubernetes backend)
- translation logic

this allows for extensibility of the Networking controller from within a single binary, and will be Solo's approach moving forward on implementing Enterprise functionality on top of the core Networking controller


codegen/helm/chart.go
```diff
@@ -170,6 +170,7 @@ func networkingOperator() model.Operator {
 			"--verbose",
 			"--disallow-intersecting-config={{ $.Values.disallowIntersectingConfig }}",
 			"--watch-output-types={{ $.Values.watchOutputTypes }}",
+			"--local-mode={{ $.Values.settings.relay.enabled }}",
```
harveyxia, line 173: do we anticipate any other use cases for running networking in local mode? if not, maybe we should call this `relay-mode`

pkg/mesh-networking/reconciliation/networking_reconciler.go
```diff
@@ -58,6 +60,7 @@ type networkingReconciler struct {
 	reconciler                 skinput.InputReconciler
 	remoteResourceVerifier     verifier.ServerResourceVerifier
 	disallowIntersectingConfig bool
+	relayEnabled               atomic.Bool
```
harveyxia, line 63: Why does this need to be atomic?

pkg/mesh-networking/start.go
```diff
@@ -23,41 +23,40 @@ import (
 
 type NetworkingOpts struct {
 	*bootstrap.Options
-	disallowIntersectionConfig bool
+	disallowIntersectingConfig bool
```
harveyxia, line 26: thanks :)

pkg/mesh-networking/start.go
```diff
@@ -23,41 +23,40 @@ import (
...
 	watchOutputTypes           bool
+	localMode                  bool
 }
 
 func (opts *NetworkingOpts) AddToFlags(flags *pflag.FlagSet) {
-	flags.BoolVar(&opts.disallowIntersectionConfig, "disallow-intersecting-config", false, "if enabled, Gloo Mesh will detect and report errors when outputting service mesh configuration that overlaps with existing config not managed by Gloo Mesh")
-	flags.BoolVar(&opts.watchOutputTypes, "watch-output-types", true, "if disabled, Gloo Mesh will not resync upon changes to service mesh config managed by Gloo Mesh")
+	flags.BoolVar(&opts.disallowIntersectingConfig, "disallow-intersecting-config", false, "if true, Gloo Mesh will detect and report errors when outputting service mesh configuration that overlaps with existing config not managed by Gloo Mesh")
+	flags.BoolVar(&opts.watchOutputTypes, "watch-output-types", true, "if true, Gloo Mesh will resync upon changes to the service mesh config output by Gloo Mesh")
+	flags.BoolVar(&opts.localMode, "local-mode", false, "if true, Gloo Mesh will un not attempt to directly read/write to registered clusters. This option should be enabled when using Relay for multi cluster I/O.")
```
harveyxia, line 34:
```suggestion
	flags.BoolVar(&opts.localMode, "local-mode", false, "if true, Gloo Mesh will not attempt to directly read/write to registered clusters. This option should be enabled when using Relay for multi cluster I/O.")
```

pkg/mesh-networking/reconciliation/networking_reconciler.go
ilackarms: so go race detector will not complain (it gets read and written in separate goroutines)

pkg/mesh-networking/start.go
```diff
@@ -13,86 +17,226 @@ import (
...
+		opts.SyncCertificateIssuerInputStatuses = func(ctx context.Context, snapshot certissuerinput.Snapshot) error {
+			return snapshot.SyncStatusesMultiCluster(ctx, parameters.McClient, certissuerinput.SyncStatusOptions{
+				CertificateRequest: true,
+			})
+		}
+	}
+}
+
+// custom entryoint for the
+func StartExtended(ctx context.Context, opts *NetworkingOpts, extensions ExtensionOpts) error {
```
harveyxia, line 182: typo: `StartExtended` -> `StartExtender`

pkg/mesh-networking/start.go
```diff
@@ -13,86 +17,226 @@ import (
...
+		*opts.Options,
+		extensions.DisableMultiCluster,
+	)
 }
 
 type networkingStarter struct {
-	disallowIntersectingConfig bool
-	watchOutputTypes           bool
+	*NetworkingOpts
+	ExtensionOpts
```
harveyxia, line 198: nit: can we be consistent about using either pointers or values

pkg/mesh-networking/start.go
```diff
@@ -13,86 +17,226 @@ import (
...
+		// sync statuses to multicluster clients (default)
+		opts.SyncCertificateIssuerInputStatuses = func(ctx context.Context, snapshot certissuerinput.Snapshot) error {
+			return snapshot.SyncStatusesMultiCluster(ctx, parameters.McClient, certissuerinput.SyncStatusOptions{
+				CertificateRequest: true,
+			})
+		}
+	}
+}
+
+// custom entryoint for the
```
harveyxia, line 181: comment needs to be completed

pkg/mesh-networking/start.go
harveyxia: since it's an entrypoint, can we move this function to the top of the file next to `Start`?

pkg/mesh-networking/translation/networking_translator.go
```diff
@@ -23,44 +23,44 @@ import (
...
 
 type OutputSnapshots struct {
-	istio   istiooutput.Snapshot
-	appmesh appmeshoutput.Snapshot
-	smi     smioutput.Snapshot
-	local   localoutput.Snapshot
+	Istio   istiooutput.Snapshot
+	Appmesh appmeshoutput.Snapshot
+	Smi     smioutput.Snapshot
+	Local   localoutput.Snapshot
```
harveyxia, line 29: Why are these now now exposed? I don't see any references to them outside this package

ilackarms: for access inside of enterprise repo

pkg/mesh-networking/start.go
ilackarms: wrt, `StartExtender`, it's not an extender we're starting, rather an _extended_ version of networking

api/settings/v1alpha2/settings.proto
```diff
@@ -53,6 +53,7 @@ message RelaySettings {
 
     // Connection info for the Relay Server. Gloo Mesh will fetch discovery resources from this server
     // and push translated outputs to this server.
+    // Note: currently this field has no effect as the relay server runs in-process of the networking pod.
     GrpcServer server = 2;
```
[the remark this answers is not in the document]
ilackarms: keeping it around because i expect us to extract the relay server in the near future

harveyxia (approved):

PR #1224, opened 2021-01-25
docs: add open source license attribution

saiskee:
Adds makefile target to generated-code to create open source attribution docs for gloo mesh. This allows us to create a doc that lists licenses that our dependencies use.


joekelley (commented):
Looks good! I'm wondering if there's a way this could be made more portable though -- I imagine the CLI implemented here would be useful in other repos more or less as-is

go.mod
```diff
@@ -63,13 +63,14 @@ require (
 	github.com/sirupsen/logrus v1.7.0
 	github.com/solo-io/anyvendor v0.0.1
 	github.com/solo-io/external-apis v0.1.4
-	github.com/solo-io/go-utils v0.20.0
+	github.com/solo-io/go-list-licenses v0.0.4-0.20210127191650-c1072fce4ee2
```
joekelley, line 66: Could we pin to a release here?

joekelley (approved):

PR #1226, opened 2021-01-26
Disambiguate GVKs in snapshots

joekelley:
When two Kinds are alike in different Group/Versions, our snapshot interfaces break. This PR bumps skv2 to a version that supports unambiguous snapshot interfaces.


joekelley:
Will need to bump to skv2 0.18 when https://github.com/solo-io/skv2/pull/189 merges in order for this to be ready to go

PR #1253, opened 2021-02-05
Flat network subsets

EItanya:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1255


api/discovery/v1alpha2/traffic_target.proto
```diff
@@ -24,16 +24,10 @@ message TrafficTargetSpec {
     // The mesh with which this traffic target is associated.
     .core.skv2.solo.io.ObjectRef mesh = 2;
 
-    message KubeService {
+    // The discovered workloads which serve as backends for this TrafficTarget
+    repeated .core.skv2.solo.io.ObjectRef workloads = 3;
```
jenshu, line 28: should this go in the Status, since it's the result of applying the workload_selector_labels? (or is it something different?)

joekelley, line 28: is this an optimization to replace a function call during translation to find the backing workloads?

pkg/mesh-discovery/translation/traffictarget/detector/traffictarget_detector.go
```diff
@@ -126,82 +137,113 @@ func getMeshForKubeService(
...
+	backingWorkloads := workloadutils.FindBackingWorkloads(kubeService, meshWorkloads)
+	// if discovery is enabled, do not return
+	if backingWorkloads.Length() == 0 {
+		return false
+	}
+	handleWorkloadDiscoveredMesh(ctx, tt, backingWorkloads, kubeService, endpoints)
+	return true
+}
+
+func handleWorkloadDiscoveredMesh(
```
joekelley, line 155: This seems like a responsibility for the [workload detector](https://github.com/solo-io/gloo-mesh/blob/5d3ea22d594b3fbeb83ceb930f0c4459fedeadc6/pkg/mesh-discovery/translation/workload/detector/meshworkload_detector.go#L58) -- wdyt?

api/discovery/v1alpha2/traffic_target.proto
EItanya: This is just the mesh workloads which back the service, not the result of applying config

EItanya: Yes

jenshu: is it representing the same workloads selected via the workload_selector_labels?
what if the `workloads` conflict with the `workload_selector_labels`?

EItanya: Yes, it's those same workloads

pkg/mesh-discovery/translation/traffictarget/detector/traffictarget_detector.go
EItanya: Yea, I think I vastly overcomplicated it

api/discovery/v1alpha2/traffic_target.proto
joekelley: It's likely that in the future we'll have new types of traffic targets which aren't backed by workloads. wdyt about moving to the KubeService message?

EItanya: Ok let's discuss. I pulled it out because I assumed that every service we represent would have to be backed by some set of concrete workloads. Whether or not this field has data, I think that assumption will always hold.

api/discovery/v1alpha2/workload.proto
```diff
@@ -25,6 +26,17 @@ message WorkloadSpec {
     // Appmesh specific metadata.
     AppMesh app_mesh = 5;
 
+    // An endpoint representing the direct method of communication with a given workload
+    message Endpoint {
+        // Leaving this message open for further endpoint info
```
harveyxia, line 31: I think we can take this comment out since it will appear in user facing docs

api/discovery/v1alpha2/workload.proto
```diff
@@ -25,6 +26,17 @@ message WorkloadSpec {
...
+
+        string ip_address = 1;
```
harveyxia, line 33: Can we add a docstring for this (even if it's simplistic)

api/discovery/v1alpha2/workload.proto
```diff
@@ -25,6 +26,17 @@ message WorkloadSpec {
...
+    }
+
+    // A series of IP addresses which correspond to the underlying workloads which this WorkloadSpec owns
```
harveyxia, line 36:
```suggestion
    // A set of IP addresses exposed by the workload
```

api/discovery/v1alpha2/workload.proto
```diff
@@ -25,6 +26,17 @@ message WorkloadSpec {
     // Appmesh specific metadata.
     AppMesh app_mesh = 5;
 
+    // An endpoint representing the direct method of communication with a given workload
```
harveyxia, line 29:
```suggestion
    // An endpoint exposed by the workload
```

api/discovery/v1alpha2/workload.proto
```diff
@@ -25,6 +26,17 @@ message WorkloadSpec {
...
+    message Endpoint {
```
harveyxia, line 30: Can we move this message definition to the bottom alongside the others?

pkg/mesh-discovery/reconciliation/discovery_reconciler.go
```diff
@@ -101,7 +101,7 @@ func Start(
 	}
 
 	if clusters != nil {
-		// running in master mode; our reconciler should watch local and remote resources
+		// running in management mode; our reconciler should watch local and remote resources
```
harveyxia, line 104:
```suggestion
		// running in non-relay mode; our reconciler should watch local and remote resources
```

pkg/mesh-discovery/translation/traffictarget/detector/traffictarget_detector.go
```diff
@@ -56,44 +51,50 @@ func NewTrafficTargetDetector() TrafficTargetDetector {
...
+	)
+	if !validTrafficTarget {
+		return nil
+	}
+
+	return trafficTarget
 }
 
-func getMeshForKubeService(
+func addMeshForKubeServiceIfValid(
```
harveyxia, line 90:
```suggestion
func addMeshForKubeService(
```

pkg/mesh-discovery/translation/traffictarget/detector/traffictarget_detector.go
```diff
@@ -56,44 +51,50 @@ func NewTrafficTargetDetector() TrafficTargetDetector {
...
 	ctx context.Context,
+	tt *v1alpha2.TrafficTarget,
```
harveyxia, line 92: Can we just pass the traffic target and not also the kubeService? In the caller we can use a conditional or a switch statement to only invoke this for kube-service-backed traffic targets. This also sets up the pattern for non-kube-service backed traffic target processing.

pkg/mesh-discovery/translation/traffictarget/detector/traffictarget_detector.go
```diff
@@ -126,82 +127,48 @@ func getMeshForKubeService(
...
+	if validMesh != nil {
+		tt.Spec.Mesh = validMesh
+		return true
+	}
 
-		// derive subsets from backing workloads
-		kubeService.Subsets = findSubsets(backingWorkloads)
+	// if no mesh was found from the annotation, check the workloads
+	backingWorkloads := workloadutils.FindBackingWorkloads(kubeService, meshWorkloads)
+	// if discovery is enabled, do not return
```
harveyxia, line 137: what does this comment mean?

pkg/mesh-discovery/translation/traffictarget/detector/traffictarget_detector.go
harveyxia: I think we can keep the logic of populating backing workloads and subsets here (in the traffic target detector). But I would structure the logic like so:

1. find the backing workloads (already happening on L136)
2. set the backing workload refs on the traffic target (can either inline this after L136 or give it a function like `setBackingWorkloads`)
3. find the subsets from the backing workloads (name this `findSubsets`)

pkg/mesh-discovery/utils/workloadutils/backing_workloads.go
```diff
@@ -9,15 +9,14 @@ import (
 func FindBackingWorkloads(
 	service *v1alpha2.TrafficTargetSpec_KubeService,
 	meshWorkloads v1alpha2sets.WorkloadSet,
-) v1alpha2.WorkloadSlice {
-
-	var result []*v1alpha2.Workload
+) v1alpha2sets.WorkloadSet {
```
harveyxia, line 12: Can we revert this back to returning a slice? I don't see the set being used

pkg/mesh-discovery/translation/workload/detector/meshworkload_detector.go
```diff
@@ -82,12 +97,16 @@ func (d workloadDetector) DetectWorkload(workload types.Workload, meshes v1alpha
...
 			},
-			Mesh: meshRef,
+			Mesh:      meshRef,
+			Endpoints: endpoints,
 		},
 	}
 }
 
-func (d workloadDetector) getMeshForPods(pods corev1sets.PodSet, meshes v1alpha2sets.MeshSet) *v1alpha2.Mesh {
+func (d workloadDetector) getMeshForPods(
```
harveyxia, line 106: Why are we separating parameters per line here, in L119, and L134 but not L182? If we want to adopt this strict convention we should discuss it first.

Also, my preference would be to limit the changes only to the logic pertinent to the issue, this makes it easier to review and track changes later through git blame.

api/discovery/v1alpha2/traffic_target.proto
harveyxia: I think it makes sense to include workloads in the spec root since "workload" does not imply k8s, but we should add a comment stating "A TrafficTarget is not guaranteed to have backing workloads, e.g. when fronting external services" or something to that effect.

pkg/mesh-discovery/translation/traffictarget/detector/traffictarget_detector.go
EItanya: I already moved it back. It simplifies the logic quite a bit

api/discovery/v1alpha2/workload.proto
EItanya: I put it where I did because I find protos much easier to grok when the messages are right above the references. I would love to discuss moving to this pattern where possible

pkg/mesh-discovery/translation/traffictarget/detector/traffictarget_detector.go
EItanya: There's no reason to do those sorts of checks since this is specifically a kubernetes discovery component, so the kube service will always exist

pkg/mesh-discovery/utils/workloadutils/backing_workloads.go
EItanya: How do we decide when to use each one? In general the use of slice vs. set seemed pretty arbitrary here, and since sets are very nice to work with I figured it was worth using them whenever possible

pkg/mesh-discovery/translation/traffictarget/detector/traffictarget_detector.go
EItanya: Although, I think I may want to move it back because of a detail I just read in the Endpoints documentation. Endpoints includes health information, which is something we want to have included in our endpoints

pkg/mesh-discovery/translation/traffictarget/detector/traffictarget_detector.go
```diff
@@ -161,54 +168,50 @@ func handleWorkloadDiscoveredMesh(
...
-				sub.IpAddresses = append(sub.IpAddresses, addr.IP)
+
+			for _, workload := range backingWorkloads {
+				kubeWorkload := workload.Spec.GetKubernetes()
+				if kubeWorkload == nil {
+					continue
+				}
+				// Check if TargetRef points to a child of a backing workload to get the labels
+				if addr.TargetRef.Namespace == kubeWorkload.GetController().GetNamespace() &&
+					strings.HasPrefix(addr.TargetRef.Name, kubeWorkload.GetController().GetName()+"-") {
```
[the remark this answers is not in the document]
EItanya: Yea that makes sense to me. My issues was this: I was trying to avoid doing this computation since we already do it in the workload detector. I see a couple ways to do this:
1. Pass all of the workload types into the TrafficTarget detector and do this computation again
2. Somehow store this information on the workload

harveyxia:
(leaving this here for posterity)

e2e tests appear to be timing out due to hardware constraints, the liveness probe for various deployments are failing, [for example](https://github.com/solo-io/gloo-mesh/pull/1253/checks?check_run_id=1863673344#step:10:2404). Perhaps it's time to move these e2e to circle

harveyxia (dismissed):

pkg/mesh-discovery/translation/traffictarget/detector/traffictarget_detector.go
```diff
@@ -126,76 +131,110 @@ func getMeshForKubeService(
...
-					sub.IpAddresses = append(sub.IpAddresses, addr.IP)
+
+	for _, epSub := range endpoint.Subsets {
+		sub := &v1alpha2.TrafficTargetSpec_KubeService_EndpointsSubset{}
+		for _, addr := range epSub.Addresses {
+			addr := addr
+			if addr.TargetRef == nil {
+				contextutils.LoggerFrom(ctx).Debugf(
+					"skipping endpoint address %v because targetRef is nil",
+					addr.TargetRef,
```
ilackarms, line 190: this logline will always log `<nil>` (addr.TargetRef == nil)

how could target ref be nil here?

EItanya: Whoops, will log the encapsulating object. The ref is optional in kubernetes, so need to check

harveyxia (approved):

PR #1264, opened 2021-02-10
meshctl utilities

harveyxia:


pkg/meshctl/utils/port_forward.go
```diff
@@ -0,0 +1,146 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/portforward"
+	"k8s.io/client-go/transport/spdy"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Open a port-forward against the specified deployment. Returns the stop channel and the local port.
+// If localPort is unspecified, a free port will be chosen at random.
+// Close the port forward by closing the returned channel.
+func PortForwardFromDeployment(
```
ashleywang1, line 27: Is there any way (this can be a future PR) that this code can fit in here instead? https://github.com/solo-io/go-utils/blob/master/cliutils/port_forward.go

It seems like a useful function across multiple repos.

harveyxia: definitely—i started by adding it here to prove out it's utility before promoting to a shared lib

pkg/meshctl/utils/port_forward.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	// the following code is based on this reference, https://github.com/kubernetes/client-go/issues/51
+	roundTripper, upgrader, err := spdy.RoundTripperFor(config)
+	if err != nil {
+		return nil, "", err
+	}
+
+	path := fmt.Sprintf("/api/v1/namespaces/%s/pods/%s/portforward", podNamespace, podName)
+	hostIP := strings.TrimLeft(config.Host, "htps:/")
```
joekelley, line 78: should this be https?

harveyxia: no weirdly, it errored when i made it `https`, I didn't dig into the reason

joekelley (approved):

joekelley (approved):

PR #1268, opened 2021-02-11
Add init-plugin-manager command

ryantking:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/129

PR #1269, opened 2021-02-12
Meshctl plugin fixes

ryantking:
Fixes an issue where an exec error was given on all not found commands since they were checked for in the system path.
Fixes the outstanding comments on https://github.com/solo-io/gloo-mesh/pull/1268


pkg/meshctl/commands/initpluginmanager/initpluginmanager.go
```diff
@@ -18,7 +18,8 @@ func Command(ctx context.Context) *cobra.Command {
 	opts := &options{}
 	cmd := &cobra.Command{
 		Use:   "init-plugin-manager",
-		Short: "Install the plugin manager",
+		Short: "Install the GLoo Mesh Enterprise CLI plugin manager",
```
joekelley, line 21: nit: GLoo -> Gloo

joekelley (approved):

Issue #538, opened 2020-04-08
un-nest HTTP Method in TrafficPolicy HTTP Matcher, convert to string


EItanya:
currently, request matcher requires the user to write something like this:

```
method:
  method: GET
```

which forces a redundant level of nesting just so we can make our code use an Enum and be nullable. While I appreciate the stronger guarantees this makes for our own code, I'd like to see the user experience look like this:

```
method: GET
```

in other words, i don't think we should place additional burden on our users just to make the work on our backend (slightly) less complicated.

after discussing with @harveyxia I think we should revert the HTTP Method matcher to a string and perform manual validation in the translator (rather than relying on the `jsonpb` parser, which currently performs validation when parsing the enum itself).

joekelley:
done in https://github.com/solo-io/gloo-mesh/pull/1287/files#diff-41d76041a6ae498319dbcb47496b7d097fff108471ff618e43a8c1a062953951R61

PR #1287, opened 2021-02-19
API 1.0

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/142
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/167


api/networking/v1alpha2/traffic_policy.proto
```diff
@@ -50,6 +50,104 @@ message TrafficPolicySpec {
...
+            Specifies a set of URL query parameters which requests must match in entirety (all query params must match).
+            The router will check the query string from the *path* header against all the specified query parameters
+        */
+        repeated QueryParameterMatcher query_parameters = 5;
+
+        // HTTP Method/Verb to match on. If none specified, the matcher will ignore the HTTP Method
+        HttpMethod method = 6;
+
+        // Express an optional HttpMethod by wrapping it in a nillable message.
+        message HttpMethod {
```
joekelley, line 93: what do you think about removing this in favor of a simple inline string? we've had an issue open for this for a while, now's the time if we want to make the change https://github.com/solo-io/gloo-mesh/issues/538

api/networking/v1alpha2/traffic_policy.proto
```diff
@@ -50,6 +50,104 @@ message TrafficPolicySpec {
     */
     repeated HttpMatcher http_request_matchers = 3;
 
+    TrafficPolicy policy = 4;
+
+    // Parameters for matching routes. All specified conditions must be satisfied for a match to occur.
+    message HttpMatcher {
```
joekelley, line 56: currently this is only used here, but in the future there may be value to having this elsewhere. let's extract it to common

harveyxia: Discussed offline, we'll wait to extract until the first instance of reuse. The change will not be breaking as long as we keep the message identical and the field name the same.

api/common/v1/selectors.proto
```diff
@@ -0,0 +1,112 @@
...
+        When used in a Gloo Mesh Role, a wildcard (`"*"`) must be specified to match any namespace.
+    */
+    repeated string namespaces = 2;
+
+    /*
+        If specified, match Kubernetes workloads if they exist in one of the specified clusters.
+        When used in a networking policy, omission matches any cluster.
+        When used in a Gloo Mesh Role, a wildcard (`"*"`) must be specified to match any cluster.
+    */
+    repeated string clusters = 3;
```
ilackarms, line 75: we should encapsulate these fields in a KubeWorkloadSelector

api/discovery/v1/destination.proto
```diff
@@ -0,0 +1,174 @@
...
+    .core.skv2.solo.io.ObjectRef mesh = 2;
+
+    // Describes a Kubernetes service.
+    message KubeService {
+
+        // Reference to the Kubernetes service object.
+        .core.skv2.solo.io.ClusterObjectRef ref = 1;
+
+        // Selectors for the set of pods targeted by the Kubernetes service.
+        map<string, string> workload_selector_labels = 2;
```
ilackarms, line 35: we should evaluate the use of workload labels here as not alll kube services are backed by pods

docs/content/guides/multicluster_communication.md
```diff
@@ -25,7 +25,7 @@ We will now perform a *multi-cluster traffic split*, splitting traffic from the
 
 {{< tabs >}}
 {{< tab name="YAML file" codelang="shell">}}
-apiVersion: networking.mesh.gloo.solo.io/v1alpha2
+apiVersion: networking.mesh.gloo.solo.io/v1
```
joekelley, line 28: we should ensure that all traffic policies in our guides are updated to reflect the new nested policy message

joekelley (approved):

PR #1305, opened 2021-02-26
docs-cleanup

ashleywang1:


docs/content/guides/federate_identity.md
```diff
@@ -23,6 +23,23 @@ To illustrate these concepts, we will assume that:
...
+## Enforce mTLS
+
+Apply the following yaml to both your management plane and remote cluster:
+
+```shell
+apiVersion: "security.istio.io/v1beta1"
+kind: "PeerAuthentication"
+metadata:
+  name: "default"
+  namespace: "istio-system"
```
harveyxia, line 35: let's include a note that we're assuming `istio-system` is the root namespace for the istio deployment

harveyxia (approved):

PR #1309, opened 2021-03-01
locality routing doc

EItanya:


docs/content/guides/locality_routing.md
```diff
@@ -0,0 +1,141 @@
...
+    name: bookinfo-federation
+    namespace: bookinfo
+EOF
+```
+
+{{% notice note %}}
+For demonstration purposes, we're setting `consecutiveErrors` to 1 and `maxEjectionPercent` to 100 to more easily trigger the failover. However, these should most likely not be used in production scenarios.
+{{% /notice %}}
+
+The `virtualMesh` field indicates the control planes that the VirtualDestination is visible to. It will be visible to all meshes in the `VirtualMesh`. A list of meshes can optionally be supplied here as well.
```
jenshu, line 65:
```suggestion
The `virtualMesh` field indicates the control planes that the VirtualDestination is visible to. It will be visible to all meshes in the `VirtualMesh`. Alternatively, a list of meshes can be supplied here instead. 
```
(to make it clear they can't provide both meshes and virtual mesh)

joekelley (approved):

Issue #1312, opened 2021-03-01
cannot re-init plugin manager


joekelley:
**Describe the bug**
If you install the plugin manager, uninstall the plugin manager, and then attempt to re-install the plugin manager, you will encounter an error. 

**To Reproduce**
1. meshctl init-plugin-manger
2. meshctl plugin uninstall plugin
3. 👇 
```
❯ meshctl init-plugin-manager
++ mktemp -d
+ cd /var/folders/66/5b1xj_gn1r7c64cfxgj7kmb40000gn/T/tmp.dZiH7ZAp
+ version=v1.0.0-beta5
++ uname
++ tr '[:upper:]' '[:lower:]'
+ os=darwin
+ arch=amd64
+ curl -o plugin -fsSL https://storage.googleapis.com/gloo-mesh-enterprise/meshctl-plugins/plugin/v1.0.0-beta5/meshctl-plugin-darwin-amd64
+ chmod u+x plugin
+ ./plugin index add default https://github.com/solo-io/meshctl-plugin-index.git
Error: index already exists
Error: exit status 1
```

**Expected behavior**
Either the meshctl-plugin install script should be able to handle the case that the index already exists, or we should clean up the plugin index when uninstalling `plugin` such that this error does not occur.

**Additional context**
meshctl 0.12.7

ryantking:
`init-plugin-manager` should require a `-f` to clear everything and reinit if it finds existing files.

PR #1319, opened 2021-03-04
Fix issues with init-plugin-manager

ryantking:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1312


pkg/meshctl/commands/initpluginmanager/initpluginmanager.go
```diff
@@ -20,58 +22,150 @@ func Command(ctx context.Context) *cobra.Command {
...
+	tempDir, err := ioutil.TempDir("", "")
+	if err != nil {
+		return nil, err
+	}
+	binPath := filepath.Join(tempDir, "plugin")
+
+	url := fmt.Sprintf(
+		"https://storage.googleapis.com/gloo-mesh-enterprise/meshctl-plugins/plugin/v1.0.0-beta7/meshctl-plugin-%s-%s",
+		runtime.GOOS, runtime.GOARCH,
+	)
```
joekelley, line 141: - we should consider continuing to expose a "latest" build here rather than hardcoding to a particular version, in order to pick up any potential bugfixes
- we don't support all possible GOOS or GOARCH values -- we should check for macos and fetch darwin, else get linux, and use amd64 in all cases

ryantking: Spoke offline about this, we're fine leaving the pinned version since the pinned version will install the latest version. We can bump the pinned version on breaking changes to the plugin manager, which we don't anticipate a lot of.

pkg/meshctl/commands/initpluginmanager/initpluginmanager.go
```diff
@@ -2,76 +2,170 @@ package initpluginmanager
...
-		b, err := ioutil.ReadAll(res.Body)
-		if err != nil {
-			logrus.WithError(err).Debug("could not read response body")
-		} else {
-			logrus.Debugf("response body: %s", string(b))
-		}
-		res.Body.Close()
-
+		logrus.Debug(string(b))
 		return nil, eris.Errorf("could not download script: %d %s", res.StatusCode, res.Status)
```
joekelley, line 158: nit: we're downloading the entire plugin-plugin now, no longer just the script

pkg/meshctl/commands/initpluginmanager/initpluginmanager.go
```diff
@@ -2,76 +2,170 @@ package initpluginmanager
...
 	}
+	if err := ioutil.WriteFile(binPath, b, 0755); err != nil {
+		return nil, err
+	}
+	return &pluginBinary{binPath, home}, nil
```
joekelley, line 163: nit: let's use named fields here for initialization

pkg/meshctl/commands/initpluginmanager/initpluginmanager.go
```diff
@@ -2,76 +2,170 @@ package initpluginmanager
...
+}
+
+type pluginBinary struct {
+	path string
+	home string
+}
+
+func checkExisting(home string, force bool) error {
+	dirty := false
+	for _, dir := range []string{"index", "receipts", "store"} {
```
joekelley, line 98: nit: let's extract the string slice here to a local variable for reuse below

joekelley (commented):
looks good, going to do some local testing

joekelley (approved):

PR #1320, opened 2021-03-04
Better meshctl output

ryantking:
Confiures logrus to use more CLI friendly output:

Screen Shot 2021-03-04 at 5 20 31 PM

Those are hooked up to `logrus.Debug`, `logrus.Info`, `logrus.Warn`, and `logrus.Error` so all existing locations that use those will work.

Also enhanced the logging in the install/register commands a bit.


pkg/meshctl/install/helm/installer.go
```diff
@@ -150,7 +150,10 @@ func output(release *release.Release, dryRun bool, isUpgrade bool) {
 		if isUpgrade {
 			verb = "upgrading"
 		}
-		logrus.Infof("finished %s chart %s as release", verb, release.Chart.Name())
+		logrus.Infof(
+			"Finished %s chart '%s' as release to %s:%s",
```
joekelley, line 154:
```suggestion
			"Finished %s chart '%s' as release %s:%s",
```

joekelley (approved):

PR #1329, opened 2021-03-10
Update meshctl dashboard command

ashleywang1:


pkg/meshctl/commands/dashboard/dashboard.go
```diff
@@ -78,7 +78,7 @@ func getStaticPort(ctx context.Context, kubeconfigPath, kubectx, namespace strin
 	if err != nil {
 		return "", err
 	}
-	dep, err := client.AppsV1().Deployments(namespace).Get(ctx, "gloo-mesh-apiserver", metav1.GetOptions{})
+	dep, err := client.AppsV1().Deployments(namespace).Get(ctx, "dashboard", metav1.GetOptions{})
```
jenshu, line 81: will we need to update the deployment name too? (in this same file)

joekelley:
We should also update our admin dashboard guide: https://docs.solo.io/gloo-mesh/latest/guides/accessing_enterprise_ui/

jenshu (approved):

PR #1331, opened 2021-03-11
improve install version selection

ryantking:


pkg/meshctl/commands/install/install.go
```diff
@@ -267,7 +265,14 @@ func installEnterprise(ctx context.Context, opts enterpriseOptions) error {
 		chartName = "gloo-mesh-enterprise"
 	)
 	if opts.version == "" {
-		version, err := helm.GetLatestChartVersion(repoURI, chartName)
+		cliVersion, err := version.NewVersion(cliversion.Version)
+		if err != nil {
+			return eris.Wrapf(err, "invalid CLI version: %s", cliversion.Version)
```
joekelley, line 270: We should be wary of erroring here in the event that we invoke this with a development version of the CLI, or via `go run` rather than a compiled binary. Could we instead only error in the event that the user didn't provide an override?

ryantking: We only get to this code path if the user did not provide an override

pkg/meshctl/install/helm/version.go
```diff
@@ -2,40 +2,92 @@ package helm
...
+func GetLatestChartMinorVersion(repoURI, chartName string, stable bool, major, minor int) (string, error) {
+	return getLatestChartVersion(repoURI, chartName, func(version version.Version) bool {
+		return version.Segments()[0] == major && version.Segments()[1] == minor &&
+			(!stable || version.Prerelease() == "")
+	})
+}
+
+func getLatestChartVersion(
+	repoURI, chartName string,
+	isVersionCopmatible func(version version.Version) bool,
```
joekelley, line 29: nit: copmatible -> compatible

ashleywang1:
The flags say:
```
      --version string             Version to install, defaults to latest if omitted
```
for both community and enterprise installs. Should this be updated?

docs/content/reference/cli/meshctl_install.md
```diff
@@ -29,7 +29,8 @@ difference between the editions.
       --kubecontext string         Name of the kubeconfig context to use for the management cluster
       --namespace string           namespace in which to install Gloo Mesh (default "gloo-mesh")
   -r, --register                   Also register the cluster
-      --version string             Version to install, defaults to latest if omitted
+      --version string             Version to install.
+                                   Community defaults to meshctl version, enterprise defaults to latest
```
ashleywang1, line 33: Or should it be 'latest compatible'?

ryantking: I think 'latest compatible' is vaguer than since meshctl version for community and latest matching minor for enterprise isn't obvious from taht.

joekelley: latest stable

pkg/meshctl/install/helm/version.go
```diff
@@ -2,40 +2,92 @@ package helm
...
+}
+
+func GetLatestChartMinorVersion(repoURI, chartName string, stable bool, major, minor int) (string, error) {
+	return getLatestChartVersion(repoURI, chartName, func(version version.Version) bool {
+		return version.Segments()[0] == major && version.Segments()[1] == minor &&
+			(!stable || version.Prerelease() == "")
+	})
+}
+
+func getLatestChartVersion(
```
harveyxia, line 27: this logic can be simplified if we return a sorted version list from `getChartVersions` using the [example logic here](https://github.com/hashicorp/go-version#version-sorting)

then we can just grab the last element in this function

pkg/meshctl/install/helm/version.go
```diff
@@ -2,40 +2,92 @@ package helm
...
 	if err != nil {
-		return "", err
+		return nil, err
 	}
 	defer res.Body.Close()
+	b, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return nil, eris.Wrap(err, "unable to read response body")
+	}
 	if res.StatusCode != http.StatusOK {
```
harveyxia, line 65: not sure if this will make any difference but it seems more natural to check for an ok response code before attempting to read the body

pkg/meshctl/install/helm/version.go
```diff
@@ -2,40 +2,92 @@ package helm
...
+	"github.com/hashicorp/go-version"
 	"github.com/rotisserie/eris"
+	"github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v2"
 )
 
-func GetLatestChartVersion(repoURI, chartName string) (string, error) {
+func GetLatestChartVersion(repoURI, chartName string, stable bool) (string, error) {
+	return getLatestChartVersion(repoURI, chartName, func(version version.Version) bool {
+		return !stable || version.Prerelease() == ""
```
harveyxia, line 16: Can we add a comment here and on L22-23 describing the logic of `isVersionCompatible` in English? it's a bit hard to parse from the conditional alone

pkg/meshctl/install/helm/version.go
ryantking: We still need to make sure the version is compatible. Mainly for enteprrise, where we don't want the latest version available, we want the latest stable version within the same minor as the community edition.

pkg/meshctl/install/helm/version.go
ryantking: If the request is successfully completed (as in hits a remote and get a response) you must handle the response body since there's an open connection. Regardless of the status code, we must read then close the body in order to close the connection.

harveyxia (approved):

PR #1332, opened 2021-03-11
auto-register enterprise

yuval-k:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/201


pkg/meshctl/enterprise/enterprise_utils.go
```diff
@@ -17,13 +19,139 @@ type RegistrationOptions struct {
...
 	}
+
+	values := map[string]string{
+		"relay.serverAddress": opts.RelayServerAddress,
+		"relay.authority":     "enterprise-networking.gloo-mesh",
+		"relay.insecure":      strconv.FormatBool(opts.RelayServerInsecure),
+		"relay.cluster":       opts.ClusterName,
+	}
+
+	if !opts.RelayServerInsecure {
```
joekelley, line 133: We dropped support for insecure mode in enterprise 1.0.0-beta6, so we can drop the insecure flag and any codepaths unique to insecure mode cc @ilackarms

pkg/meshctl/commands/cluster/register/cluster_register.go
```diff
@@ -119,7 +127,18 @@ type enterpriseOptions enterprise.RegistrationOptions
...
 func (o *enterpriseOptions) addToFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&o.RelayServerAddress, "relay-server-address", "", "The address via which the enterprise agent will communicate with the relay server.")
-	flags.BoolVar(&o.RelayServerInsecure, "relay-server-insecure", true, "Communicate with the relay server over an insecure connection.")
+	flags.BoolVar(&o.RelayServerInsecure, "relay-server-insecure", false, "Communicate with the relay server over an insecure connection.")
+
+	flags.StringVar(&o.RootCASecretName, "root-ca-secret-name", "", "Secret name for the root CA for communication with relay server.")
+	flags.StringVar(&o.RootCASecretNamespace, "root-ca-secret-namespace", "", "Secret namespace for the root CA for communication with relay server.")
+
+	flags.StringVar(&o.ClientCertSecretName, "client-cert-secret-name", "", "Secret name for the client cert for communication with relay server.")
+	flags.StringVar(&o.ClientCertSecretNamespace, "client-cert-secret-namespace", "", "Secret namespace for the client cert for communication with relay server.")
```
ilackarms, line 136: it might be helpful to add the info here that the client secret is not expected to be provided by the user; if not provided, it will be generated with this name / namespace

yuval-k:
added graphics:
image

pkg/meshctl/enterprise/enterprise_utils.go
```diff
@@ -56,6 +214,68 @@ func RegisterCluster(ctx context.Context, opts RegistrationOptions) error {
...
+		}
+		if err == nil {
+			break
+		}
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-timeout:
+			return eris.Errorf("timed out waiting for client cert")
+		case <-time.After(5 * time.Second):
```
ryantking, line 273: Can use a `time.Ticker` here

ryantking (commented):
CLI logic looks good from my end 👍🏻

pkg/meshctl/enterprise/enterprise_utils.go
yuval-k: Do you mean start a ticker outside the for loop?
i want to wait 5 seconds from the time of GetSecret returns, so i think time.After is more suitable

pkg/meshctl/enterprise/enterprise_utils.go
```diff
@@ -56,6 +214,68 @@ func RegisterCluster(ctx context.Context, opts RegistrationOptions) error {
...
+		if err != nil {
+			return err
+		}
+		if bootstrapTokenCreated {
+			logrus.Info("🗑 Removing bootstrap token")
+			key := client.ObjectKey{
+				Name:      opts.TokenSecretName,
+				Namespace: opts.TokenSecretNamespace,
+			}
+			err = remoteKubeSecretClient.DeleteSecret(ctx, key)
```
joekelley, line 240: if the user registers the management cluster, this will prevent future registrations from succeeding

joekelley: I think we just need to check if opts.RemoteContext == opts.MgmtContext

joekelley (approved):

PR #1334, opened 2021-03-12
add util function to get a destination from workload

jenshu:


pkg/mesh-discovery/utils/workloadutils/backing_workloads.go
```diff
@@ -18,6 +18,20 @@ func FindBackingWorkloads(
 	})
 }
 
+// Public to be used in enterprise
+func FindDestinationForWorkload(
+	workload *v1.Workload,
+	destinations v1alpha2sets.DestinationSet,
+) *v1.Destination {
```
harveyxia, line 25: this needs to be a slice of destinations, a single workload can back multiple destinations.

PR #1337, opened 2021-03-16
perf improvments

yuval-k:
do not merge
cc @ilackarms


harveyxia:
apimachinery's string set exposes an ["UnsortedList"](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/sets/string.go#L180) method, should we update [skv2 sets](https://github.com/solo-io/skv2/blob/master/contrib/pkg/sets/sets.go#L85) to use `UnsortedList` instead of `List`?

harveyxia (approved):

PR #1338, opened 2021-03-17
Don't regen reference docs during CI

ryantking:
Fixes the empty API docs by not regenerating the versioned control reference docs (CLI and API) during CI. Only the changelog will be generated at CI time.

### DIFF
--- a/docs/docsgen/generate_reference_docs.go
+++ b/docs/docsgen/generate_reference_docs.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"text/template"
 
@@ -104,7 +105,7 @@ type Options struct {
 
 func Execute(opts Options) error {
 	rootDir := filepath.Join(moduleRoot, opts.DocsRoot)
-	if os.Getenv("ONLY_CHANGELOG") == "" {
+	if !checkEnvVariable("ONLY_CHANGELOG") {
 		if err := generateCliReference(rootDir, opts.Cli); err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


docs/docsgen/generate_reference_docs.go
```diff
@@ -104,11 +104,13 @@ type Options struct {
 
 func Execute(opts Options) error {
 	rootDir := filepath.Join(moduleRoot, opts.DocsRoot)
-	if err := generateCliReference(rootDir, opts.Cli); err != nil {
-		return err
-	}
-	if err := generateApiDocs(rootDir, opts.Proto); err != nil {
-		return err
+	if os.Getenv("ONLY_CHANGELOG") == "" {
```
jenshu, line 107: what if ONLY_CHANGELOG was set to `false`?

docs/docsgen/generate_reference_docs.go
```diff
@@ -250,3 +253,12 @@ func getGitVersion() (string, error) {
...
 	return strings.TrimSpace(string(version)), nil
 }
+
+func checkEnvVariable(key string) bool {
+	val := os.Getenv(key)
+	if val == "" {
+		return false
+	}
+	b, err := strconv.ParseBool(val)
+	return err != nil || b // treat set env variables as true
```
jenshu, line 263:
```suggestion
	return err != nil && b // treat set env variables as true
```

ryantking: This is intentional, if err is non-nil, it means that the env variable was set to some non-empty gobbledygook, which we treat as `true`, only if the err is nil, which means the env variable was set to a valid bool value,  do we want to check the value of the bool.

harveyxia (approved):

Issue #1353, opened 2021-03-19
Remove Metrics svc from networking controller entirely


ashleywang1:
**Describe the bug**
It can confuse prometheus to have the metrics svc in both places.

PR #1354, opened 2021-03-19
Remove metrics svc from networking controller

ashleywang1:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1353

### DIFF
--- a/install/helm/gloo-mesh/templates/deployment.yaml
+++ b/install/helm/gloo-mesh/templates/deployment.yaml
@@ -126,7 +126,6 @@ spec:
       - image: {{ $networkingImage.registry }}/{{ $networkingImage.repository }}:{{ $networkingImage.tag }}
         args:
         - networking
-        - --metrics-port={{ $.Values.networking.ports.metrics }}
         - --settings-name={{ $.Values.glooMeshOperatorArgs.settingsRef.name }}
         - --settings-namespace={{ $.Values.glooMeshOperatorArgs.settingsRef.namespace }}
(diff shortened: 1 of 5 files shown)


install/helm/gloo-mesh/templates/deployment.yaml
```diff
@@ -126,7 +126,6 @@ spec:
       - image: {{ $networkingImage.registry }}/{{ $networkingImage.repository }}:{{ $networkingImage.tag }}
         args:
         - networking
-        - --metrics-port={{ $.Values.networking.ports.metrics }}
```
joekelley, line 129: would it be possible to still have this port running on the image / deployment, but not expose the service?

joekelley: That way we can still leverage it for debugging, but we don't have to worry about it being available publicly if the user uses a LoadBalancer service type.

ashleywang1: Not the way we currently have skv2 codegen set up

joekelley: In that case, could we add a new helm value for the metrics port? It can be a custom value that we just pass through here.

ashleywang1: There is a way to do it -> which is done in this commit https://github.com/solo-io/gloo-mesh/pull/1354/commits/cd23909a7725c5feeec25476ebc7c50a9568798e

joekelley (approved):

PR #1409, opened 2021-04-09
added new virtual-dest ingress api

EItanya:


api/discovery/v1/mesh.proto
```diff
@@ -61,10 +61,17 @@ message MeshSpec {
...
 
+
             // The externally-reachable address on which the gateway is listening for TLS connections.
             // This will be the address used for cross-cluster connectivity.
-            // Defaults to the LoadBalancer Address (or NodeIP) of the Kubernetes service (depending on its type).
-            string external_address = 2;
+                // Defaults to the LoadBalancer Address (or NodeIP) of the Kubernetes service (depending on its type).
+            oneof external_address_type {
+                // Will be populated if the LoadBalancer Address is a DNS name
+                string external_address = 2;
```
jenshu, line 70: can we name this `dns_name`?

EItanya: That would be a break, thus why I didn't change it

pkg/mesh-discovery/translation/mesh/detector/istio/istio_mesh_detector.go
```diff
@@ -209,20 +209,33 @@ func getIngressGateway(
...
 			return nil, eris.Errorf("no loadBalancer.ingress status reported for service")
 		}
 
-		externalAddress = ingress[0].Hostname
-		if externalAddress == "" {
-			externalAddress = ingress[0].IP
+		gatewayInfo.ExternalTlsPort = uint32(tlsPort.Port)
+
+		gatewayInfo.ExternalAddressType = &discoveryv1.MeshSpec_Istio_IngressGatewayInfo_DnsName{
+			DnsName: ingress[0].Hostname,
```
ilackarms, line 228: is external address always a dns name? iirc they are ips on GKE. i think we need to check

EItanya: Yes it is, in GKE it will show up in `ingress[0].IP` as shown below

api/discovery/v1/mesh.proto
```diff
@@ -61,10 +61,19 @@ message MeshSpec {
             // [Defaults to](https://github.com/istio/istio/blob/ab6cc48134a698d7ad218a83390fe27e8098919f/pkg/config/constants/constants.go#L73) `{"istio": "ingressgateway"}`.
             map<string, string> workload_labels = 1;
 
+            // Deprecated in favor of dns_name or external_ip
```
harveyxia, line 64:
```suggestion
            // DEPRECATED: in favor of dns_name or external_ip
```

api/discovery/v1/mesh.proto
```diff
@@ -61,10 +61,19 @@ message MeshSpec {
...
             // The externally-reachable address on which the gateway is listening for TLS connections.
             // This will be the address used for cross-cluster connectivity.
             // Defaults to the LoadBalancer Address (or NodeIP) of the Kubernetes service (depending on its type).
-            string external_address = 2;
+            oneof external_address_type {
+                // Will be populated if the LoadBalancer Address is a DNS name
+                string dns_name = 5;
+
+                // Will be populated if the LoadBalancer Address is an IP
+                string external_ip = 6;
```
harveyxia, line 75: nit: remove the `external` prefix

pkg/mesh-discovery/translation/mesh/detector/istio/istio_mesh_detector.go
harveyxia: Can we set the GKE IP address also in the new `external_ip` field so we also have it in a non-deprecated field ?

pkg/mesh-networking/translation/istio/mesh/access/access_policy_translator.go
```diff
@@ -128,5 +128,9 @@ func buildGlobalAuthPolicy(
 }
 
 func ingressGatewayAuthPolicyName(ingressGateway *discoveryv1.MeshSpec_Istio_IngressGatewayInfo) string {
-	return IngressGatewayAuthPolicyName + "-" + kubeutils.SanitizeNameV2(ingressGateway.ExternalAddress)
+	address := ingressGateway.GetExternalAddress()
```
harveyxia, line 131: can we move off the deprecated field and use `dnsName`?

pkg/mesh-discovery/translation/mesh/detector/istio/istio_mesh_detector.go
EItanya: That happens on line 233

ilackarms: i see, i was a bit confused due to some code duplication. think we could clean it up so we only need to set the fields in `gatewayInfo` one time?

ilackarms: the convention i follow when constructing a new struct in `go` is to initialize all the variables  which constitute a struct before initializing the struct itself, whenever possible. for example, prefer this:

```go
field1 := 1
field2 := 2
myStr := StructType{field1: field1, field2: field2}
```

over this:

```go

myStr := StructType{}
myStr.field1 = 1
myStr.field2 = 2

```

i believe it helps to clarify and organize object construction in a language with no constructors

EItanya: Normally I agree, but you can't really do that with `oneof`s since the type is private

EItanya: Resolved offline

ilackarms (approved):

PR #1413, opened 2021-04-12
Virtual dest port fixes api

EItanya:


api/discovery/v1/destination.proto
```diff
@@ -106,6 +114,19 @@ message DestinationSpec {
                 SubLocality sub_locality = 3;
             }
         }
+
+        // Describes the endpoints's ports. See [here](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/endpoints-v1/) for more information.
+        message EndpointPort {
+
+            uint32 port = 1;
```
joekelley, line 121: let's comment all these fields

api/enterprise/networking/v1beta1/virtual_destination.proto
```diff
@@ -51,13 +51,25 @@ message VirtualDestinationSpec {
...
     // The port on which the VirtualDestination listens.
+    // The target port of the VirtualDestination can be specified in 2 ways, in the following order of precedence:
+    // 1. If the name and the number are specified:
+    //    In this case the VirtualDestination will listen on the number specified, by number and target the ports specified by name.
+    // 2. The number is specified but not the name:
+    //    In this case the port number must match an existing port on all backends. If it does then the port matching the 
+    //    input port number will be used to find the target_port.
+    //
+    // Any other combination will be considered an error
+    // NOTE: Static VirtualDestinations only support number (2).
```
ashleywang1, line 62: Static vs Localized VirtualDestinations

ashleywang1: For more, see: https://docs.solo.io/gloo-mesh/latest/guides/locality_routing/

api/discovery/v1/destination.proto
```diff
@@ -63,12 +63,22 @@ message DestinationSpec {
...
             string name = 2;
-
+            
+            // Protocol on which this port serves traffic (HTTP, TCP, UDP, etc...)
             string protocol = 3;
 
             // Available in Kubernetes 1.18+, describes the application protocol.
             string app_protocol = 4;
+
+            // target_port in kubernetes may either be a string referring to the port_name, or a number referring to the number
```
harveyxia, line 75:
```suggestion
            // A Kubernetes Service's spec.targetPort can either be a string referring to the port name, or a number referring to the number. See [this reference](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/#ServiceSpec).
```

api/discovery/v1/destination.proto
```diff
@@ -63,12 +63,22 @@ message DestinationSpec {
...
+            oneof target_port {
+                // Name of the targer_port
```
jenshu, line 77:
```suggestion
                // Name of the target port
```

api/discovery/v1/destination.proto
```diff
@@ -63,12 +63,22 @@ message DestinationSpec {
...
+                string target_port_name = 5;
+                // Number of the target_port
```
jenshu, line 79:
```suggestion
                // Number of the target port
```

api/enterprise/networking/v1beta1/virtual_destination.proto
```diff
@@ -51,13 +51,25 @@ message VirtualDestinationSpec {
...
     message Port {
```
harveyxia, line 63: Discussed offline—agreed to explicitly distinguish the port on which the VirtualDestination will be exposed, from the target port that the VirtualDestination uses when targeting backing Destinations

api/enterprise/networking/v1beta1/virtual_destination.proto
```diff
@@ -50,26 +50,24 @@ message VirtualDestinationSpec {
...
         uint32 number = 1;
 
-        // Protocol of the requests sent to the VirtualDestination. Must be one of `HTTP`, `HTTPS`, `GRPC`, `HTTP2`, `MONGO`, `TCP`, `TLS`.
+        // OPTIONAL: Protocol of the requests sent to the VirtualDestination. Must be one of `HTTP`, `HTTPS`, `GRPC`, `HTTP2`, `MONGO`, `TCP`, `TLS`.
         string protocol = 2;
 
-        // Name of the port which will be used
-        string name = 3;
+        // REQUIRED: Name of the target_port which will be used
+        oneof target_port {
```
harveyxia, line 64: This API structure assumes that all backing Destinations for a given VirtualDestination will have a common target port name or number. Let's call out that assumption in this API.

api/discovery/v1/destination.proto
```diff
@@ -63,12 +63,22 @@ message DestinationSpec {
...
-
+            
+            // Protocol on which this port serves traffic (HTTP, TCP, UDP, etc...)
             string protocol = 3;
 
             // Available in Kubernetes 1.18+, describes the application protocol.
             string app_protocol = 4;
+
+            // A Kubernetes Service's spec.targetPort can either be a string referring to the port name, or a number referring to the number. See [this reference](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/#ServiceSpec).
+            oneof target_port {
```
harveyxia, line 76: We discussed this offline already, but I'm still not sure I understand why we need to capture the name/number distinction explicitly.

We use this field [here](https://github.com/solo-io/gloo-mesh-enterprise/pull/362/files#diff-57ddf20c8c1eda9d266ec5c97ffe021072b1529f1bf585558bd3bf154379dfe5R520-R531), and we could easily compare an `intOrString` field to both the endpointPort name and number, and safely assume that only one will match.

Unless our system needs to know whether the targeted port is being referenced by number or by name, we should be agnostic to that distinction and use a single string field to represent both. This will allow us to avoid some complexity in the code resulting from dealing with the explicit oneof.

EItanya: `IntOrString` functions exactly like a `oneof` in terms of code. It allows the user to distinguish the type of value specified without having to parse it themselves. Proto has the limitation that the fields have different names, but besides that it is the same.

We could potentially assume that, but I don't see how making that assumption is preferable to being explicit. The `VirtualDestination` as a concept is already complex enough, I think having a clearer interaction with the fields through code is desirable in this instance.

harveyxia: Makes sense

api/enterprise/networking/v1beta1/virtual_destination.proto
EItanya: Added

api/enterprise/networking/v1beta1/virtual_destination.proto
```diff
@@ -50,26 +50,24 @@ message VirtualDestinationSpec {
...
+    // VirtualDestination port infomation.
+    // Contains information about which port to listen on, as well as which backend port to target.
     message Port {
 
-        // Port number.
+        // REQUIRED: Port number which the VirutalDestination will listen on.
         uint32 number = 1;
 
-        // Protocol of the requests sent to the VirtualDestination. Must be one of `HTTP`, `HTTPS`, `GRPC`, `HTTP2`, `MONGO`, `TCP`, `TLS`.
+        // OPTIONAL: Protocol of the requests sent to the VirtualDestination. Must be one of `HTTP`, `HTTPS`, `GRPC`, `HTTP2`, `MONGO`, `TCP`, `TLS`.
```
ilackarms, line 60: i would mention the default in the comment

api/enterprise/networking/v1beta1/virtual_destination.proto
```diff
@@ -50,26 +50,24 @@ message VirtualDestinationSpec {
...
 
-        // Name of the port which will be used
-        string name = 3;
+        // REQUIRED: Name of the target_port which will be used
+        oneof target_port {
+            // If the target_name is specified, the VirtualDestination will attempt to find a port by this name on all backing services
+            string target_name = 3;
+
+            // If the target_number is specified, the VirtualDestination will attempt to find a port by this number on all backing services
+            string target_number = 4;
```
ilackarms, line 70: what is the reason for assuming all backing services will listen on the same port? not sure that restriction will be desirable for our users. if you look at our other apis where we specify destination services (https://github.com/solo-io/gloo-mesh/blob/main/api/networking/v1/traffic_policy.proto#L158), each destination gets its own port specifier. can we not permit the same flexibility here?

EItanya: We don't actually use BackingDestinations here, we use [selectors](https://github.com/solo-io/gloo-mesh/blob/3f1f4700dbf17c76fcce43fc531acd3ff3f242f0/api/enterprise/networking/v1beta1/virtual_destination.proto#L89). Selecting backends for the localized VirtualDestination is more akin to selecting targets for the `TrafficPolicy` We could potentially extend the selector with port information as well, but not sure what would be gained there. This system mirrors how kubernetes does target_port selection on `Services` so would most likely be intuitive. Also, this was the style of API I discussed with @djannot. Happy to discuss further if that doesn't work

ilackarms: this does make me inclined to think our selector should optionally support ports; but in the case the user is using a label selector to match backends, this style of api of "matching" ports makes sense. if @djannot expects users will be fine with it, i don't see a strong objection to moving forward with this api as-is

api/enterprise/networking/v1beta1/virtual_destination.proto
joekelley: Do we have data from users/the field that says this is an acceptable constraint?

api/enterprise/networking/v1beta1/virtual_destination.proto
EItanya: Should we maybe make an issue to discuss possible port selection?

api/enterprise/networking/v1beta1/virtual_destination.proto
EItanya: Discussed with @djannot offline and I believe he said yes, but I will leave it to him to respond

djannot:
The PR looks good to me

api/enterprise/networking/v1beta1/virtual_destination.proto
djannot: I think it's fine as we give them the name option.
If they have different port numbers, then they just need to make sure they have the same port name.

joekelley: lgtm from @djannot says yes 🙂

harveyxia (approved):

Issue #1433, opened 2021-04-15
Optimize Gloo Mesh reconciliation model


harveyxia:
Currently Gloo Mesh reconciles the state of the world on each loop. We can improve the performance and scalability of Gloo Mesh's reconciliation by moving to a model where only objects relevant to the k8s event are processed.

Part of this requires splitting out the FederationTranslator from the MeshTranslator so it can be invoked for specific federated Destinations without having to recompute the entire VirtualMesh.

harveyxia:
TODOs:

1. Garbage collect resources originating from extension servers

PR #1434, opened 2021-04-15
Restructure federation translator

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1433


api/discovery/v1/destination.proto
```diff
@@ -211,19 +211,22 @@ message DestinationStatus {
...
+        // Reference to the VirtualMesh object.
+        .core.skv2.solo.io.ObjectRef ref = 1;
+
         /*
             For any Workload that this Destination has been federated to (i.e., any Workload controlled by a Mesh whose reference appears in `federated_to_meshes`),
             that Workload will be able to reach this Destination using this DNS name.
             For Kubernetes Destinations this includes Workloads on clusters other than the one hosting this Destination.
         */
-        string federated_hostname = 1;
+        string federated_hostname = 2;
```
EItanya, line 222: When adding fields to a proto the numbers should remain consistent, or else it's "breaking"

harveyxia: We don't persist any of these protobufs in serialized format, so it should be okay

EItanya: It's not just about storage. If for instance a relay-agent were to have the old version, this would break. It could serialize this to an `Any` and then the server will fail to deserialize.

pkg/mesh-networking/translation/istio/destination/federation/federation_translator.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+type translator struct {
+	ctx                       context.Context
+	virtualServiceTranslator  virtualservice.Translator
+	destinationRuleTranslator destinationrule.Translator
+}
+
+func NewTranslator(
+	ctx context.Context,
```
EItanya, line 56: While we're refactoring can we pass the `ctx` in as an argument to the function, rather than hold a reference?

harveyxia: Will do 👍, I've been doing this in the follow up PR (that branches off this one). Do you mind if I make the changes there in order to minimize the need to rebase?

pkg/mesh-networking/translation/istio/mesh/federation/federation_translator.go
```diff
@@ -65,24 +54,13 @@ type Translator interface {
 }
 
 type translator struct {
-	ctx                       context.Context
-	destinations              discoveryv1sets.DestinationSet
-	virtualServiceTranslator  virtualservice.Translator
-	destinationRuleTranslator destinationrule.Translator
+	ctx context.Context
```
EItanya, line 57: Same comment about passing in this `ctx` rather than holding on to it

pkg/mesh-networking/translation/istio/mesh/federation/federation_translator.go
```diff
@@ -91,7 +69,7 @@ func (t *translator) Translate(
 	mesh *discoveryv1.Mesh,
 	virtualMesh *discoveryv1.MeshStatus_AppliedVirtualMesh,
 	outputs istio.Builder,
-	reporter reporting.Reporter,
+	_ reporting.Reporter,
```
EItanya, line 72: Can we remove this arg if it is no longer used?

harveyxia: All of our translator interfaces take an error reporter, so I'd prefer to keep this here, wdyt?

ilackarms (approved):

Issue #1431, opened 2021-04-15
Opt-out of sidecar injection in the gloo-mesh install/registration namespace


joekelley:
Some Istio installations will attempt to inject all namespaces if the `istio-injection=disabled` label isn't applied to the namespace. Gloo Mesh components should not be injected by the mesh by default. Meshctl should add the `istio-injection=disabled` label to the installation namespace for Gloo Mesh management and agent components (i.e. gloo-mesh).

PR #1443, opened 2021-04-15
Opt out of istio sidecar injection in GM namespace

ryantking:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1431


ashleywang1:
Would it also make sense to change the docs for this? We have Helm setup steps here: https://docs.solo.io/gloo-mesh/main/setup/installation/enterprise_installation/ that reference gloo-mesh namespace creation.

install/helm/cert-agent/templates/deployment.yaml
```diff
@@ -22,6 +22,7 @@ spec:
       labels:
         app: cert-agent
       annotations:
+        sidecar.istio.io/inject: "false"
```
harveyxia, line 25: Why do we need these annotations at the pod level if we're adding them to the namespace?

rvennam: To cover the cases where the user accidentally removes the label or overwrites it. Or creates the namespace with the label set to true

rvennam: istiod and ingressgateway pods do this as well

harveyxia: 👍

ryantking: If the user has an existing install before this or is using a custom namespace, the namespace level label will not be enforced.

@rvennam Can you add context to this if I'm missing anything?

jenshu (approved):

PR #1462, opened 2021-04-22
Event based reconcile

harveyxia:


pkg/mesh-networking/translation/networking_translator.go
```diff
@@ -65,18 +89,267 @@ func (t *translator) Translate(
...
 		Appmesh: appmeshOutputs,
 		Smi:     smiOutputs,
 		Local:   localOutputs,
-	}, nil
+	}
+}
+
+// Return true if the object should be garbage collected (i.e. deleted from k8s storage).
+// Specifically, garbage collect the object if any of its parents no longer exist.
+func shouldGarbageCollect(
```
harveyxia, line 280: Global garbage collection implemented here. It assumes that all output objects have been correctly labeled with their complete set of parent input objects, and garbage collects based on the absence of those parent annotations.

In certain cases, we need to explicitly mark an object for garbage collection using the `metautils.GarbageCollectDirective`. Such use cases currently include deleting an output because it conflicts with existing user output, and deleting an AuthorizationPolicy because it has an empty spec (and therefore it would cause all traffic to be denied, so it should not be output)

Issue #804, opened 2020-06-29
VirtualMesh Restrictive Federation


nea1:
**Is your feature request related to a problem? Please describe.**
SMH currently creates ServiceEntries, DestinationRules & Gateway resources for _all_ services in _all_ other clusters in the mesh.
In a mesh of say 6 clusters running 500 workloads each then this is 5 (number of other clusters) x 500 = 2500 serviceentries created in each cluster. 
Firstly are there any sort of resource/scaling/performance issues that would be hit with approach? Even if not, it does seem quite noisy. For example if only a handful of workloads are desired to be shared across clusters then this is a lot of noise when for example debugging a routing issue. 

**Describe the solution you'd like**
One approach would be to add annotations or labels on kubernetes `services` and/or `namespaces` to indicate that a service (or services if on a namespace) are indeed a `meshservice` and should be made available across clusters in the mesh. Any services/namespaces not labelled would not be federated. An option on `VirtualMesh` would be required to switch between this mode and the current standard 'federate everything' mode.

harveyxia:
Hey Neal, this issue should be addressed once we design and implement "restrictive" federation which will allow for, as you describe, a way to specify which services should be federated to which clusters.

PR #1484, opened 2021-04-28
Restrictive Federation

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/804


pkg/mesh-networking/apply/applier.go
```diff
@@ -680,6 +680,62 @@ func getAppliedVirtualMesh(
...
+				}
+			}
+			// federate to specified Meshes
+			for _, meshRef := range federationSelector.GetMeshes() {
+				federatedToMeshes.Insert(meshRef)
+			}
+		}
+	default:
+		// no federation applied
+		return nil
```
EItanya, line 721: Since the default behavior used to be `permissive`, do we want to maintain parity with that?

harveyxia: will raise this in parking lot tomorrow, I don't have strong opinions either way

harveyxia: In today's discussion we decided to go with this approach:

1. Deprecate the existing `mode` field.
2. Honor the deprecated `mode` field only if the new `selectors` field is not populated.
3. Permissive federation semantics are expressed with the following:

```yaml
spec:
    federation:
      selectors:
      - {}
```

api/networking/v1/virtual_mesh.proto
```diff
@@ -105,14 +105,16 @@ message VirtualMeshSpec {
     // originating from Workloads external to the Destination's Mesh.
     message Federation {
 
+        // Selectively federate Destinations to specific external meshes.
+        // If omitted, all Destinations will be federated to all Meshes in the VirtualMesh.
```
EItanya, line 109: I think we should be specific about the different states, eg.
1. nil = no federation
2. empty = everything
3. selectors

pkg/mesh-networking/apply/applier.go
```diff
@@ -688,37 +688,33 @@ func getFederatedToMeshes(
...
-	case *networkingv1.VirtualMeshSpec_Federation_Permissive:
-		// permissive federation exposes the Destination to all Meshes in the VirtualMesh
-		for _, groupedMeshRef := range virtualMesh.Spec.GetMeshes() {
-			federatedToMeshes.Insert(groupedMeshRef)
-		}
-	case *networkingv1.VirtualMeshSpec_Federation_Restrictive:
-		// omitted federation selectors have permissive semantics
-		if len(mode.Restrictive.GetFederationSelectors()) < 1 {
+	// respect deprecated `mode` field only if new federation selectors are not specified
+	if len(virtualMesh.Spec.GetFederation().GetSelectors()) < 1 {
```
EItanya, line 692: super nit: 
```suggestion
	if len(virtualMesh.Spec.GetFederation().GetSelectors()) == 0 {
```

api/networking/v1/virtual_mesh.proto
```diff
@@ -105,14 +105,16 @@ message VirtualMeshSpec {
...
+        repeated FederationSelector selectors = 4;
+
+        // DEPRECATED: Use `selectors` instead. Omission of the `selectors` field has permissive semantics.
```
EItanya, line 112: Will this show up in the docs?
If not we should deprecate the `permissive` field below

api/networking/v1/virtual_mesh.proto
ilackarms:
```yaml
# disable federation
selectors: null
```

```yaml
# restricted federation
selectors: 
- destinationSelectors:
  - kubeServiceMatcher:
      labels:  
        app: reviews
      clusters:
      - remote-cluster
  meshes:
  - name: istiod-istio-system-mgmt-cluster 
    namespace: gloo-mesh
```

```yaml
# permissive federation
selectors: 
- destinationSelectors: null
  meshes: null

# can also be expressed as
selectors: 
- {}
```

EItanya: Spoke with @harveyxia offline and this sounds ok

pkg/mesh-networking/apply/applier.go
```diff
@@ -317,6 +317,9 @@ func validateAndReturnApprovedFederation(
 	reporter *applyReporter,
 	destination *discoveryv1.Destination,
 ) *discoveryv1.DestinationStatus_AppliedFederation {
+	if destination.Status.AppliedFederation == nil {
```
ilackarms, line 320: jw, why the change here?

docs/content/guides/federate_identity.md
```diff
@@ -153,7 +153,10 @@ spec:
     shared:
       rootCertificateAuthority:
         generated: {}
-  federation: {}
+  federation:
+    # federate all Destinations to all external meshes
+    selectors:
+    - {}
```
ilackarms, line 159: just wondering, why the change here?

pkg/mesh-networking/apply/applier.go
harveyxia: without this clause, a Destination's `status.AppliedFederation` will be populated even though it's not actually being federated to any external meshes

docs/content/guides/federate_identity.md
harveyxia: updating our docs to use the non-deprecated field

pkg/mesh-networking/apply/applier.go
```diff
@@ -680,6 +680,58 @@ func getAppliedVirtualMesh(
...
+func getFederatedToMeshes(
+	destination *discoveryv1.Destination,
+	parentMesh *v1.ObjectRef,
+	virtualMesh *networkingv1.VirtualMesh,
+) []*v1.ObjectRef {
+	federatedToMeshes := sets.NewResourceSet()
+
+	// respect deprecated `mode` field only if new federation selectors are not specified
+	if len(virtualMesh.Spec.GetFederation().GetSelectors()) < 1 {
+		switch virtualMesh.Spec.GetFederation().GetMode().(type) {
```
ilackarms, line 693: i believe this can panic if `mode` is nil; can we test?

api/networking/v1/virtual_mesh.proto
harveyxia: good call, i forgot that `oneofs` don't show up in generated docs

ilackarms (approved):

PR #1495, opened 2021-04-29
Add diagnostic hint when mesh dashboard not found

jameshbarton:
Add diagnostic hint to check kubconfig.  This happened to me and could be common when users are managing multiple clusters in dev environment.


joekelley (approved):
thanks @jameshbarton!

PR #1502, opened 2021-05-03
Add auth settings

ryantking:


harveyxia (commented):
Looks like you may have used a protobuf formatter that changed a bunch of comments—could we limit the changes here to just those that are relevant for your PR?

api/settings/v1/settings.proto
```diff
@@ -98,6 +104,185 @@ message GrpcServer {
     bool reconnect_on_network_failures = 3;
 }
 
+message DashboardSettings {
+    message AuthConfig {
+        google.protobuf.StringValue name = 1;
```
ashleywang1, line 109: Could you add a comment here explaining what this names? Is it the pod or some other value?

api/settings/v1/settings.proto
```diff
@@ -98,6 +104,185 @@ message GrpcServer {
...
+
+        oneof auth_config {
+            OidcConfig oidc = 2;
+        }
+    }
+
+    repeated AuthConfig auth_configs = 1;
```
ashleywang1, line 116: The fact that this is repeated suggests that we could potentially route throughout multiple authenticators. Is that what happens?

ryantking: Yes, will add some clarifying comments, which will also address your above comment

api/settings/v1/settings.proto
ryantking: I'm actually going to remove this field since we don't have a use for it yet.

changelog/v1.1.0-beta10/auth-settings.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+- type: NEW_FEATURE
+  issueLink: https://github.com/solo-io/gloo-mesh-enterprise/issues/544
```
EItanya, line 3: Don't link to enterprise issues from changelog @joekelley

api/settings/v1/settings.proto
```diff
@@ -98,6 +104,199 @@ message GrpcServer {
...
+    // behavior can be changed by defining names for each config and defining `boolean_expr` below.
+    //
+    // State is shared between successful requests on the chain, i.e., the headers returned from each
+    // successful auth service get appended into the final auth response.
+    repeated AuthConfig auth_configs = 1;
+
+    // How to handle processing of named configs within an auth config chain.
+    // An example config might be: `( basic1 || basic2 || (oidc1 && !oidc2) )`
+    // The boolean expression is evaluated left to right but honors parenthesis and short-circuiting.
+    google.protobuf.StringValue boolean_expr = 10;
```
EItanya, line 130: Do users need this level of control here, why would a request ever need to go through multiple OIDC flows?

ryantking: I guess in theory users could have multiple OIDC configs?

I guess in theory we can just remove this field and `name` until we add in additional authentication mechanisms.

EItanya (approved):

PR #1512, opened 2021-05-05
Destination validation

nmnellis:
Traffic policies have to omit clusters for it to apply to all rather than '*'

### DIFF
--- a/pkg/mesh-networking/apply/configtarget/config_target_validator.go
+++ b/pkg/mesh-networking/apply/configtarget/config_target_validator.go
@@ -66,13 +66,33 @@ func (c *configTargetValidator) ValidateVirtualMeshes(virtualMeshes v1.VirtualMe
 
 func (c *configTargetValidator) ValidateTrafficPolicies(trafficPolicies v1.TrafficPolicySlice) {
 	for _, trafficPolicy := range trafficPolicies {
-		errs := c.validateDestinationReferences(trafficPolicy.Spec.DestinationSelector)
+		errs := c.validateDestinationReferencesForTrafficPolicy(trafficPolicy.Spec.DestinationSelector)
 		if len(errs) == 0 {
 			continue
 		}
 		trafficPolicy.Status.State = commonv1.ApprovalState_INVALID
 		trafficPolicy.Status.Errors = getErrStrings(errs)
 	}
+}
+
+func (c *configTargetValidator) validateDestinationReferencesForTrafficPolicy(serviceSelectors []*commonv1.DestinationSelector) []error {
+	var errs []error
+	for _, destinationSelector := range serviceSelectors {
+		kubeServiceRefs := destinationSelector.GetKubeServiceRefs()
+		// only validate Destinations selected by direct reference
+		if kubeServiceRefs == nil {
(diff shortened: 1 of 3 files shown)


pkg/mesh-networking/apply/configtarget/config_target_validator.go
```diff
@@ -75,6 +75,26 @@ func (c *configTargetValidator) ValidateTrafficPolicies(trafficPolicies v1.Traff
...
+		kubeServiceRefs := destinationSelector.GetKubeServiceRefs()
+		// only validate Destinations selected by direct reference
+		if kubeServiceRefs == nil {
+			continue
+		}
+		for _, ref := range kubeServiceRefs.Services {
+			if !c.kubeServiceExists(ref) {
+				errs = append(errs, eris.Errorf("Destination %s not found", sets.Key(ref)))
+			}
+			if ref.ClusterName == "*" {
```
harveyxia, line 90: If we add this special check, we should add it for reference selectors in all networking policies, i.e. add this logic [here](https://github.com/solo-io/gloo-mesh/blob/07626344b7c99178c2ab5944fe06258ef2eaf905/pkg/mesh-networking/apply/configtarget/config_target_validator.go#L122) instead of scoping it to just the TrafficPolicy

harveyxia (commented):
The current behavior will report an error message of the form `destination not found for cluster *`—the question is whether this error message is a sufficient signal to the user that asterisks do not carry "select all" semantics.

I can see an argument for this special error message given that asterisks are used in Gloo Edge.

@ilackarms @joekelley any thoughts?

pkg/mesh-networking/apply/configtarget/config_target_validator.go
```diff
@@ -75,6 +75,26 @@ func (c *configTargetValidator) ValidateTrafficPolicies(trafficPolicies v1.Traff
...
+				errs = append(errs, eris.Errorf("Cannot use '*' for all clusters, please omit clusters property."))
```
ilackarms, line 91: i'm not sure why we would expect a user to try using a `*` in their `clusters` field. why `*` and not `.*`?  I would be more inclined to ensure that any time a user specifies a `cluster` (or for that matter `namespace` or `name`) to validate that they use DNS-1123 compliant strings.

ilackarms: and that this should be generally applicable to any ref or selector provided anywhere in the GM api

nmnellis: the documentation specifically says so in certain cases. https://docs.solo.io/gloo-mesh/latest/reference/api/github.com.solo-io.gloo-mesh.api.common.v1.selectors/#common.mesh.gloo.solo.io.WorkloadSelector

harveyxia: Yeah, that's an incorrect statement on the documentation and we need to correct it. For reference based selectors, all fields are matched verbatim and omission does not carry wildcard semantics.

nmnellis:
closing for now

PR #1518, opened 2021-05-06
Fix reference selector doc

harveyxia:


api/common/v1/selectors.proto
```diff
@@ -45,8 +45,7 @@ message DestinationSelector {
     // Match Kubernetes Services by direct reference.
     message KubeServiceRefs {
         /*
-            Match Kubernetes Services by direct reference.
-            When used in a networking policy, omission of any field (name, namespace, or clusterName) allows matching any value for that field.
+            Match Kubernetes Services by direct reference. All fields are required.
             When used in a Gloo Mesh Role, a wildcard (`"*"`) must be specified to match any value for the given field.
```
jenshu, line 49: to confirm, the wildcard `*` only works for Roles and not networking policies?
is there a way to "select all" in a networking policy?

harveyxia: > to confirm, the wildcard * only works for Roles and not networking policies?

Correct

> is there a way to "select all" in a networking policy?

You use a matcher-based selector and omit all fields.

jenshu (approved):

PR #1526, opened 2021-05-10
Additional translate traffic policy utils

npolshakova:
See solo-io/gloo-mesh-enterprise#419 and https://github.com/solo-io/gloo-mesh-enterprise/pull/480


pkg/mesh-networking/translation/utils/trafficpolicyutils/translate_header_manipulation.go
```diff
@@ -1,7 +1,11 @@
 package trafficpolicyutils
```
ilackarms, line 1: we probably want to rename the file since it's no longer gonna be responsible solely for header manipulation

pkg/mesh-discovery/translation/mesh/detector/istio/istio_mesh_detector_test.go
```diff
@@ -277,6 +277,7 @@ var _ = Describe("IstioMeshDetector", func() {
 						TrustDomain:          trustDomain,
 						IstiodServiceAccount: serviceAccountName,
 						IngressGateways: []*discoveryv1.MeshSpec_Istio_IngressGatewayInfo{{
+							Namespace:      "namespace",
 							WorkloadLabels: workloadLabels,
```
Sodman, line 281: Name is now part of the discovered mesh's IngressGatewayInfo
```suggestion
							Namespace:      "namespace",
							Name:           "ingress-svc",
							WorkloadLabels: workloadLabels,
```

harveyxia:
Can we reuse this logic in the existing TrafficPolicy decorators?

ilackarms (changes requested):
looks like we still need to update the traffic policy decorators to use these util functions (rather than have duplicated logic)

harveyxia:
Also let's add unit tests

Sodman:
> Also let's add unit tests

I think the existing unit tests for the traffic policy functionality should cover this already, as it's not actually changing behavior, just moving the logic into a shared package. We'll want a few smoke tests on the Enterprise side to make sure that this shared logic is actually running in the translation loop though.

PR #1528, opened 2021-05-10
Smarter init plugin manager command

ryantking:
the `init-plugin-manager` command downloads a copy of the plugin manager binary, which it then uses to add the default index and install the plugin manager to its permanent location. Previously the temporary binary was a hardcoded version, this PR adds a bit of better logic that will parse the manifest definition in the index to find the latest compatible version (matching minor version).

### DIFF
--- a/pkg/meshctl/commands/initpluginmanager/initpluginmanager.go
+++ b/pkg/meshctl/commands/initpluginmanager/initpluginmanager.go
@@ -17,6 +17,7 @@ import (
 	pkgversion "github.com/solo-io/gloo-mesh/pkg/common/version"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"gopkg.in/yaml.v2"
 )
 
 func Command(ctx context.Context) *cobra.Command {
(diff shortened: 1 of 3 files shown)


pkg/meshctl/commands/initpluginmanager/initpluginmanager.go
```diff
@@ -167,3 +162,61 @@ func (binary pluginBinary) run(args ...string) (string, error) {
...
+func findLatestCompatibleBinary(ctx context.Context) (io.ReadCloser, error) {
+	v, err := version.NewVersion(pkgversion.Version)
+	if err != nil {
+		return nil, eris.Wrap(err, "unable to parse version")
+	}
+	major, minor, patch := v.Segments()[0], v.Segments()[1], v.Segments()[2]
+	if v.Prerelease() != "" {
+		minor -= 1
+		patch = 20
+	}
```
Sodman, line 176: What's the significance of vX.1.20 here? Or am I misreading this? Ie what happens when current gloo-mesh version is a beta release?

Sodman: Ah I see it's `-=`, missed the minus on my first pass! So this tries the previous version's stable release starting at patch 20 and working its way down?

Sodman: Could we maybe grab the latest stable release from the github releases page? You can get a list using 

```
curl  -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/solo-io/gloo-mesh/releases
```

and all we'd need to pull out is `tag_name`.

I'm just thinking because 20 feels like an arbitrary patch number (Gloo Edge 1.5.x and 1.6.x have > 20 patch releases, for example) which we might overshoot in Gloo-Mesh. It might be worth grabbing the tags from the API and iterating backwards over it to get the versions to try, instead of counting down from 20?

ryantking: The issue is the version is dependent on enterprise, which has its own patch numbering. One option that's less arbitrary is start a patch 0, then go up until we find the latest? It'll require more requests, but will guarantee the latest patch.

PR #1530, opened 2021-05-10
ServiceDependency API

harveyxia:


api/discovery/v1/workload.proto
```diff
@@ -68,6 +68,10 @@ message WorkloadStatus {
     // The set of WasmDeployments that have been applied to this Workload.
     repeated AppliedWasmDeployment applied_wasm_deployments = 3;
 
+    // Specifies the [ServiceDependencies]({{< versioned_link_path fromRoot="/reference/api/github.com.solo-io.gloo-mesh.api.enterprise.networking.v1alpha1.service_dependency/" >}})
+    // that apply to this Workload, and the resulting Destination hostnames that this Workload can send traffic to.
+    DestinationDependencies destination_dependencies = 4;
```
ilackarms, line 73: do you think it's confusing using both `DestinationDependencies` and `ServiceDependencies` ?

api/discovery/v1/workload.proto
```diff
@@ -93,4 +97,23 @@ message WorkloadStatus {
...
+        repeated AppliedServiceDependency applied_service_dependencies = 1;
+
+        // The set of Destination hostnames that this Workload depends on.
+        repeated string destination_hostnames = 2;
+
+        // Describes a [ServiceDependency]({{< versioned_link_path fromRoot="/reference/api/github.com.solo-io.gloo-mesh.api.enterprise.networking.v1alpha1.service_dependency/" >}})
+        // that applies to this workload.
+        message AppliedServiceDependency {
+            // Reference to the ServiceDependency object.
+            .core.skv2.solo.io.ObjectRef service_dependency_ref = 1;
```
ilackarms, line 113: do we want to store the spec here like we do with discovery / policy objects?

api/enterprise/networking/v1beta1/service_dependency.proto
```diff
@@ -0,0 +1,48 @@
...
+option (extproto.equal_all) = true;
+
+/*
+    A ServiceDependency specifies explicit dependencies between traffic sources and destinations.
+    Depending on the underlying service mesh, explicitly describing dependencies can improve the performance
+    of the data plane by pruning away any unneeded networking configuration from the relevant proxies.
+
+    The complete set of destination dependencies for a given traffic source is the aggregation of all
+    unique Destinations selected by any applicable ServiceDependencies. If a traffic source has no applied
+    ServiceDependencies, its destination dependency configuration defaults to the behavior of the underlying service mesh.
```
ilackarms, line 20: `s/destination dependenc/service dependenc/g`

ilackarms: i might want to add some context here to describe the context of a service graph. explain a little bit the functionality here to a user who may not understand the benefit of "pruning away unneeded configuration from relevant proxies". for example, that this is a declarative mechanism for explicitly defining the dependencies between services, effectively blocking communication between services outside of these explicit definitions.

api/discovery/v1/workload.proto
harveyxia: discussed offline: there's a larger story about preserving last good config, which we currently don't implement consistently across our networking policies. We'll handle this concern when/if we prioritize this

api/discovery/v1/workload.proto
harveyxia: renaming to `ServiceDependencies`

ilackarms (approved):

PR #1531, opened 2021-05-11
External ca code ext

EItanya:


pkg/certificates/agent/reconciliation/pod-bouncer/pod_bouncer.go
```diff
@@ -0,0 +1,212 @@
+package pod_bouncer
```
ilackarms, line 1: nit: underscores are discouraged in package names according to official go style: https://blog.golang.org/package-names

PR #1546, opened 2021-05-12
Sort helm versions docsgen

ashleywang1:


docs/docsgen/copy_helm_values_docs.go
```diff
@@ -136,8 +136,8 @@ func copyHelmValuesDocsForComponent(
 		}
 		versions = append(versions, version)
 	}
-	sort.Reverse(semver.Collection(versions))
-	versions = getLatestPerMinorVersion(versions)
+	sort.Sort(sort.Reverse(semver.Collection(versions)))
```
joekelley, line 139: do we need to reverse then sort? could we just sort?

ashleywang1: Yep! We expect the results to be sorted from highest semver to lowest, wheras sort will return the versions from lowest to highest.

ryantking (approved):

docs/docsgen/copy_helm_values_docs.go
harveyxia: Elaborating on Ashley's explanation a bit—`sort.Reverse(...)` returns a `sort.Interface`, which requires an invocation of `sort.Sort` to actually perform the sorting mutation

PR #1556, opened 2021-05-17
External ca agent ext

EItanya:
Create code extension point for certificate agent reconciler. This reconciler has a few responsibilities:
1. Generate csr and create `CertificateRequest` with said csr.
2. Get signed cert, and signing cert from `CertificateRequest, and then save it to the target `Secret`
3. Delete pods so that mesh can pick up the new certificate. 

These responsibilities mirror the different statuses of the `IssuedCertificate` which the `CertificateRequest` is generated from. Therefore, the new translator has multiple entrypoints, each one corresponding to a different section of the workflow.
```
// These functions correspond to issued certiticate statuses
// PENDING
// REQUESTED
// ISSUED
// FINISHED
type Translator interface {

	// Should the reconciler process this resource
	ShouldProcess(ctx context.Context, issuedCertificate *certificatesv1.IssuedCertificate) bool

	// This function is called when the IssuedCertiticate is first created, it is meant to create the CSR
	// and return the bytes directly
	// If both return values are nil, it signals to the reconciler to ignore this resource.
	IssuedCertiticatePending(
		ctx context.Context,
		issuedCertificate *certificatesv1.IssuedCertificate,
		inputs input.Snapshot,
		outputs certagent.Builder,
	) ([]byte, error)

	// This function is called when the IssuedCertiticate has been REQUESTED, this means that the
	// CertificateRequest has been successfully created, and is passed in along with the
	// IssuedCertificate
	// This function may be called multiple times in a row, as long as the IssuedCertificate is in a
	// REQUESTED state
	// If err == nil, and continueIterating == false it signals to the reconciler to ignore this resource.
	IssuedCertificateRequested(
		ctx context.Context,
		issuedCertificate *certificatesv1.IssuedCertificate,
		certificateRequest *certificatesv1.CertificateRequest,
		inputs input.Snapshot,
		outputs certagent.Builder,
	) error

	// This function is called when the IssuedCertiticate has been ISSUED, this means that the
	// CertificateRequest has been fulfilled by the relevant party. By this stage the intermediate
	// cert should already have made it to it's destination.
	IssuedCertificateIssued(
		ctx context.Context,
		issuedCertificate *certificatesv1.IssuedCertificate,
		inputs input.Snapshot,
		outputs certagent.Builder,
	) error
}
```


pkg/certificates/agent/reconciliation/cert_agent_reconciler.go
```diff
@@ -120,18 +125,21 @@ func (r *certAgentReconciler) reconcile(_ ezkube.ResourceId) (bool, error) {
...
-	inputCertificateRequests v1alpha2sets.CertificateRequestSet,
-	podBounceDirectives v1alpha2sets.PodBounceDirectiveSet,
+// Exposed for testing
+func (r *certAgentReconciler) ReconcileIssuedCertificate(
+	issuedCertificate *certificatesv1.IssuedCertificate,
+	inputSnap input.Snapshot,
 	outputs certagent.Builder,
 ) error {
+
+	// If this translator doesn't own the resource, return early
```
harveyxia, line 135: Let's make this comment more specific as to why the cert agent isn't responsible for processing this IssuedCertificate. Something to the effect of,

```suggestion
	// If the IssuedCertificate's IssuedCertificateSecret is nil, the cert-agent is not responsible for issuing the certificate.
```

pkg/certificates/agent/translation/cert_agent_translator.go
```diff
@@ -0,0 +1,217 @@
...
+// FINISHED
+type Translator interface {
+
+	// Should the reconciler process this resource
+	ShouldProcess(ctx context.Context, issuedCertificate *certificatesv1.IssuedCertificate) bool
+
+	// This function is called when the IssuedCertiticate is first created, it is meant to create the CSR
+	// and return the bytes directly
+	// If both return values are nil, it signals to the reconciler to ignore this resource.
+	IssuedCertiticatePending(
```
harveyxia, line 59: `IssuedCertiticatePending` -> `IssuedCertificatePending`

pkg/certificates/agent/translation/cert_agent_translator.go
```diff
@@ -0,0 +1,217 @@
...
+type certAgentTranslator struct{}
+
+func (c *certAgentTranslator) ShouldProcess(ctx context.Context, issuedCertificate *certificatesv1.IssuedCertificate) bool {
+	return issuedCertificate.Spec.GetIssuedCertificateSecret() != nil
+}
+
+func (c *certAgentTranslator) IssuedCertiticatePending(
+	ctx context.Context,
+	issuedCertificate *certificatesv1.IssuedCertificate,
+	inputs input.Snapshot,
```
harveyxia, line 104: not used

pkg/certificates/agent/reconciliation/cert_agent_reconciler.go
```diff
@@ -140,135 +148,92 @@ func (r *certAgentReconciler) reconcileIssuedCertificate(
...
-			issuedCertificate.Spec.Hosts,
-			issuedCertificate.Spec.Org,
-			privateKey,
-		)
-		if err != nil {
-			return err
+
+		// If translator does not return bytes or an error, it means the workflow should NOT
+		// continue any further
+		if csrBytes == nil {
```
harveyxia, line 178: this block isn't needed because it's handled on L136

pkg/certificates/agent/reconciliation/pod-bouncer/pod_bouncer.go
```diff
@@ -0,0 +1,211 @@
...
+type RootCertMatcher interface {
+	MatchesRootCert(
+		ctx context.Context,
+		rootCert []byte,
+		selector *certificatesv1.PodBounceDirectiveSpec_PodSelector,
+		allSecrets corev1sets.SecretSet,
+	) (matches bool, err error)
+}
+
+// bounce (delete) the listed pods
```
harveyxia, line 32: orthogonal concern, but maybe worth adding a TODO to investigate bouncing these pods without downtime (which might be more involved than simply using deployment rollout vs. deleting pods, since we're swapping out the certs)

pkg/certificates/agent/reconciliation/pod-bouncer/pod_bouncer.go
```diff
@@ -0,0 +1,211 @@
...
+// returns true if we need to wait before proceeding to process the podBounceDirective.
+// we must wait for the following conditions:
+// 1. istiod control plane has come back online after it has been restarted
+// 2. istio's root cert has been propagated to all istio-controlled namespaces for consumption by the data plane.
+// this should cause the reconcile to end early and persist the IssuedCertificate in the Issued state
+type PodBouncer interface {
+	BouncePods(
+		ctx context.Context,
+		podBounceDirective *certificatesv1.PodBounceDirective,
+		allPods corev1sets.PodSet,
```
harveyxia, line 42: nit: our convention so far is to assume all pods when interpreting an entity set as an input arg, so we don't need the `all` prefix here

pkg/certificates/agent/start_extended.go
```diff
@@ -0,0 +1,47 @@
+package agent
```
harveyxia, line 1: nit: makes it slightly easier to read if this file is merged with `pkg/certificates/agent/start.go`

pkg/certificates/agent/start_extended.go
```diff
@@ -0,0 +1,47 @@
...
+	opts.CertAgentReconciler.initDefaults(parameters)
+}
+
+// Options for overriding functionality of the Networking Reconciler
+type CertAgentReconcilerExtensionOpts struct {
+
+	// Hook to override Translator used by Networking Reconciler
+	MakeTranslator func(translator translation.Translator) translation.Translator
+	// Pod Bouncer to be used by translator, allows overriding the dependency
+	PodBouncer pod_bouncer.PodBouncer
```
harveyxia, line 29: do we have an alternative implementation of the pod bouncer in the related PRs?

harveyxia (commented):
Thanks for the additional unit test coverage!

pkg/certificates/agent/translation/cert_agent_translator.go
EItanya: lol

pkg/certificates/agent/translation/cert_agent_translator.go
EItanya: Used in enterprise

pkg/certificates/agent/start_extended.go
EItanya: We have an alternative implementation of the `RootCertMatcher`, a subcomponent of the `PodBouncer`, so I can replace this with the swappable component

pkg/certificates/agent/reconciliation/cert_agent_reconciler.go
```diff
@@ -38,43 +28,62 @@ var (
...
 }
 
 func Start(
 	ctx context.Context,
 	builder input.Builder,
 	mgr manager.Manager,
+	podBouncer pod_bouncer.PodBouncer,
+	translator translation.Translator,
 ) error {
 	d := &certAgentReconciler{
```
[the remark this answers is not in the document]
EItanya: So I set it up like this because I only really wanted to test the `ReconcileIssuedCertificate` function. The rest of it is just bootstrap, code similar to many other places. If you'd like I can test the `reconcile` function instead, it just requires testing many more interfaces and functions, without a lot of benefit as I see it. Also, the following line isn't really mockable:
`_, err := input.RegisterSingleClusterReconciler(ctx, mgr, d.reconcile, time.Second/2, reconcile.Options{})`
So I was trying to figure out how to test while ignoring that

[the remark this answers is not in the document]
ilackarms: my preference is to leave code private and simply put the test inside the original package. this allows you to keep packages clean for the code which consumes it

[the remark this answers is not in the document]
EItanya: I think in this case I'd be fine with that because of how the reconcile func needs to be called. I can make that change

ilackarms (approved):

PR #1559, opened 2021-05-17
Use istio with custom port for older istios in e2e

Sodman:
Setup istio with a custom port for ingress (8081) to fully test that path.

We were already doing this for Istio 1.9.x, this PR just adds it for our regression tests on Istio 1.8.x and 1.7.x.


npolshakova (approved):
Nice!

npolshakova (approved):

Issue #1560, opened 2021-05-19
Istio Gateway AUTO_PASSTHROUGH breaks Mesh federation


harveyxia:
**Describe the bug**
In 1.9.5 Istio changed their Gateway `AUTO_PASSTHROUGH` [implementation](https://github.com/istio/istio/blob/b3bdc0c71bfd6152e645b8a12074f560efa42882/pilot/pkg/networking/core/v1alpha3/gateway.go#L642) due to a [security vulnerability](https://istio.io/latest/news/security/istio-security-2021-006/). They no longer use an Envoy `sni_cluster` so our current approach of using an Envoy `tcp_cluster_rewrite` to modify the remote cluster name to the local cluster name on the server side no longer works.

The new Istio implementation instead looks at the hosts specified on the Gateway CR, and creates routing config (i.e. a filter chain) per known Service, but these filter chains match on the “DNSSrvSubsetKey” cluster name (e.g. “outbound_.9080_._.reviews.bookinfo.svc.cluster.local”), whereas our ServiceEntries representing remote services on the client side use the federated FQDN as the SNI (e.g. “reviews.bookinfo.svc.remote-cluster.soloio”).

**Workaround:**

Create ServiceEntries on the server side in order to produce mappings from the SNI that the client sends (which is qualified with the cluster name) to the correct destination. For example,

```yaml
apiVersion: networking.istio.io/v1alpha3
kind: ServiceEntry
metadata:
  name: reviews-remote
  namespace: istio-system # must be in same namespace as gateway
spec:
  resolution: DNS
  location: MESH_INTERNAL
  ports:
  - name: http
    number: 9080
    protocol: HTTP
  exportTo:
  - .
  hosts:
  - reviews.bookinfo.svc.remote-cluster.soloio
  endpoints:
  - address: reviews.bookinfo.svc.cluster.local
```

Additionally, a DestinationRule is needed to configure the ServiceEntry with ISTIO_MUTUAL tls so that this ServiceEntry shows up in the gateway Envoy config, since the new auto_passthrough implementation filters out any plaintext destinations. For example,

```yaml
apiVersion: networking.istio.io/v1beta1
kind: DestinationRule
metadata:
  name: ratings-bookinfo-federated
  namespace: istio-system
spec:
  host: reviews.bookinfo.svc.remote-cluster.soloio
  trafficPolicy:
    tls:
      mode: ISTIO_MUTUAL
```

**Additional context**
[...]

PR #1567, opened 2021-05-20
Fix for Istio Gateway `AUTO_PASSTHROUGH`

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1560


pkg/mesh-networking/translation/istio/destination/destinationrule/destination_rule_translator.go
```diff
@@ -133,14 +133,6 @@ func (t *translator) Translate(
 		}
 	}
 
-	// TODO need a more robust implementation of determining whether a DestinationRule has any effect
```
harveyxia, line 136: Some context here: this behavior is incorrect, we'll always need to output a DR in order to enforce the Destination-specific mTLS settings declared in Settings, or overridden by a TrafficPolicy, in order to override Istio's [global mTLS settings](https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/) (under `enableAutoMtls`)

pkg/mesh-networking/translation/istio/destination/federation/federation_translator.go
```diff
@@ -120,26 +127,40 @@ func (t *translator) Translate(
...
 
 		serviceEntries = append(serviceEntries, serviceEntry)
 		virtualServices = append(virtualServices, virtualService)
 		destinationRules = append(destinationRules, destinationRule)
+
+		remoteDestinationRule = destinationRule
 	}
 
+	// translate local resources
+	localServiceEntry, localDestinationRule := t.translateForLocalMesh(
```
EItanya, line 147: Do we want to version gate this feature? There isn't much reason to have it before this new AUTO_PASSTHROUGH implementation. Same thing with creating the remote `EnvoyFilter`

EItanya: Same thing if the env var is set to maintain the old behavior

harveyxia: Discussed offline—since this new federation implementation works for both new and old Istio `AUTO_PASSTHROUGH` implementations, we'll just go with this one approach to reduce complexity in our code base

pkg/mesh-networking/translation/istio/destination/federation/federation_translator.go
```diff
@@ -120,26 +127,40 @@ func (t *translator) Translate(
...
-		// TODO(harveyxia) append Destination as parent once new GC is being implemented
 		metautils.AppendParent(t.ctx, serviceEntry, destination.Status.AppliedFederation.GetVirtualMeshRef(), networkingv1.VirtualMesh{}.GVK())
 		metautils.AppendParent(t.ctx, virtualService, destination.Status.AppliedFederation.GetVirtualMeshRef(), networkingv1.VirtualMesh{}.GVK())
 		metautils.AppendParent(t.ctx, destinationRule, destination.Status.AppliedFederation.GetVirtualMeshRef(), networkingv1.VirtualMesh{}.GVK())
 
 		serviceEntries = append(serviceEntries, serviceEntry)
 		virtualServices = append(virtualServices, virtualService)
 		destinationRules = append(destinationRules, destinationRule)
+
+		remoteDestinationRule = destinationRule
```
EItanya, line 143: Can you explain what this variable does and why it's ok that it can equal any of the `destinationRule` values since this gets set in a loop

.github/workflows/pull_request.yaml
```diff
@@ -68,9 +68,9 @@ jobs:
       fail-fast: false
       matrix:
         istioctl-binary: [
-          'https://github.com/istio/istio/releases/download/1.7.7/istio-1.7.7-linux-amd64.tar.gz',
-          'https://github.com/istio/istio/releases/download/1.8.3/istio-1.8.3-linux-amd64.tar.gz',
-          'https://github.com/istio/istio/releases/download/1.9.1/istio-1.9.1-linux-amd64.tar.gz'
+            'https://github.com/istio/istio/releases/download/1.7.7/istio-1.7.7-linux-amd64.tar.gz',
+            'https://github.com/istio/istio/releases/download/1.8.6/istio-1.8.6-linux-amd64.tar.gz',
+            'https://github.com/istio/istio/releases/download/1.9.5/istio-1.9.5-linux-amd64.tar.gz'
```
EItanya, line 73: Add `1.10`

pkg/mesh-networking/translation/istio/destination/federation/federation_translator.go
harveyxia: discussed offline: the DestinationRule config for the federated FQDN is the same as that of the remote DestinationRules, so we could either re-translate or copy over the necessary fields from the translated DR's. I opted to do the latter to save some compute. Adding a comment to clarify this logic

EItanya (approved):

Issue #1564, opened 2021-05-20
Add support for discovering `Service` `ExternalIP`


EItanya:
Currently mesh ingress discovery only checks `LoadBalancer` `Status` or `NodePort` for an external address, however, some users are using `ExternalIPs`. Gloo Mesh should be able to discover those as well.

ashleywang1:
Needs backport

ashleywang1:
Backport was done here: https://github.com/solo-io/gloo-mesh-enterprise/pull/568

ashleywang1:
Needs a Gloo Mesh Enterprise release

ashleywang1:
Hi, this has been released and tested in Gloo Mesh v1.0.11.

The following service (in a kind cluster):
```
apiVersion: v1
kind: Service
metadata:
  name: istio-ingressgateway
  namespace: istio-system
spec:
  externalIPs:
  - 80.11.12.10
...
  type: LoadBalancer
status:
  loadBalancer: {}
```

generates the following Mesh discovery object:
```
apiVersion: discovery.mesh.gloo.solo.io/v1
kind: Mesh
metadata:
  name: istiod-istio-system-mgmt-cluster
  namespace: gloo-mesh
spec:
  agentInfo:
    agentNamespace: gloo-mesh
  istio:
    ingressGateways:
    - externalAddress: 80.11.12.12
```

PR #1568, opened 2021-05-20
Load balancer external ip

ashleywang1:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1564


pkg/mesh-discovery/translation/mesh/detector/istio/istio_mesh_detector.go
```diff
@@ -217,13 +217,22 @@ func getIngressGateway(
 		gatewayInfo.ExternalTlsPort = uint32(tlsPort.NodePort)
```
ashleywang1, line 218: Under what circumstances will NodePort services have an externalIP set? Is this even a possibility?

joekelley: Seems like it's possible, so we may as well support it IMO

pkg/mesh-discovery/translation/mesh/detector/istio/istio_mesh_detector.go
```diff
@@ -207,23 +208,41 @@ func getIngressGateway(
...
 		if len(ingress) == 0 {
-			return nil, eris.Errorf("no loadBalancer.ingress status reported for service")
+			// Check for user-set external IPs
+			externalIPs := svc.Spec.ExternalIPs
+			if len(externalIPs) != 0 {
+				gatewayInfo.ExternalAddressType = &discoveryv1.MeshSpec_Istio_IngressGatewayInfo_Ip{
+					Ip: svc.Spec.ExternalIPs[0],
+				}
+				// Continue to set deprecated field until it is removed
+				gatewayInfo.ExternalAddress = svc.Spec.ExternalIPs[0]
```
joekelley, line 240: rather than copy/paste this block, could we instead either:
- extract a util function
- run this check outside the switch statement so it'll be run in all service types?

ashleywang1: It doesn't quite make sense to pull it out into a util function - I've refactored the code to make it a bit more readable.

pkg/mesh-discovery/translation/mesh/detector/istio/istio_mesh_detector.go
```diff
@@ -207,39 +208,47 @@ func getIngressGateway(
...
-			return nil, eris.Errorf("no loadBalancer.ingress status reported for service")
-		}
-
 		gatewayInfo.ExternalTlsPort = uint32(tlsPort.Port)
-
-		// If the Ip address is set in the ingress, use that
-		if ingress[0].IP != "" {
+		ingress := svc.Status.LoadBalancer.Ingress
+		var addr string
+		if len(ingress) == 0 { // Check for user-set external IPs
```
joekelley, line 228: nit: can you move this comment and the one on 238 to the following line?

pkg/mesh-discovery/translation/mesh/detector/istio/istio_mesh_detector.go
```diff
@@ -207,39 +208,47 @@ func getIngressGateway(
...
+			externalIPs := svc.Spec.ExternalIPs
+			if len(externalIPs) != 0 {
+				addr = svc.Spec.ExternalIPs[0]
+				gatewayInfo.ExternalAddressType = &discoveryv1.MeshSpec_Istio_IngressGatewayInfo_Ip{
+					Ip: addr,
+				}
+			} else {
+				return nil, eris.Errorf("no loadBalancer.ingress status reported for service. Please set an external IP on the service as a user if you are using a non-kubernetes load balancer.")
```
joekelley, line 236:
```suggestion
				return nil, eris.Errorf("no loadBalancer.ingress status reported for service. Please set an external IP on the service if you are using a non-kubernetes load balancer.")
```

pkg/test/apps/gloo-mesh/kube.go
```diff
@@ -251,15 +251,21 @@ func serviceIngressToAddress(svc *v1.Service) (string, error) {
...
-
-	// If the Ip address is set in the ingress, use that
-	if ingress[0].IP != "" {
-		address = ingress[0].IP
+		// Check for user-set external IPs
+		externalIPs := svc.Spec.ExternalIPs
+		if len(externalIPs) != 0 {
+			address = svc.Spec.ExternalIPs[0]
+		} else {
+			return "", fmt.Errorf("no loadBalancer.ingress status reported for service. Please set an external IP on the service as a user if you are using a non-kubernetes load balancer.")
```
joekelley, line 259:
```suggestion
			return "", fmt.Errorf("no loadBalancer.ingress status reported for service. Please set an external IP on the service if you are using a non-kubernetes load balancer.")
```

joekelley (approved):

Issue #1386, opened 2021-04-04
Python 2 is required to install meshctl


murphye:
**Describe the bug**

I already have Python 3 installed in Ubuntu. However, when I try to install `meshctl` I get this error:

```
curl -sL https://run.solo.io/meshctl/install | sh
Python is required to install meshctl
```

I have to run `sudo apt install python` to install Python 2 to proceed with the `meshctl` installation.

Any modern tool should be using Python 3 at this point.

joekelley:
Gloo recently resolved this: https://github.com/solo-io/gloo/pull/4751/files

PR #1593, opened 2021-05-27
python3 in install script

mlholland:
Modifies the python-checking login in the meshctl install script to check for both python 2 and 3, and use either (preferring 3).
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1386

### DIFF
--- a/hack/install.sh
+++ b/hack/install.sh
@@ -3,11 +3,11 @@
 set -eu
 
 if [ -x "$(command -v python3)" ]; then
-  echo Using Python3 to install meshctl
   alias any_python='python3'
 elif [ -x "$(command -v python)" ]; then
-  echo Using Python2 to install meshctl
   alias any_python='python'    
+elif [ -x "$(command -v python2)" ]; then
+  alias any_python='python2'    
 else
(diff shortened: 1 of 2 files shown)


hack/install.sh
```diff
@@ -2,13 +2,19 @@
...
 set -eu
 
-if ! [ -x "$(command -v python)" ]; then
-     echo Python is required to install meshctl
-     exit 1
- fi
+if [ -x "$(command -v python3)" ]; then
+  echo Using Python3 to install meshctl
+  alias any_python='python3'
+elif [ -x "$(command -v python)" ]; then
```
harveyxia, line 8: Should we also check for the `python2` binary [the way Gloo does](https://github.com/solo-io/gloo/pull/4751/files#diff-39e8107dedd87bfb05ca4868464520708c89b7d9610ed4d10e6a44c4981607a7R5-R25)?

mlholland: Is 'python2' ever used? I'll certainly add it, but I've only ever seen 'python' and 'python3' referenced.

harveyxia (approved):

PR #1598, opened 2021-05-28
add guide for ServiceDependency API

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1579


docs/content/guides/service_dependencies.md
```diff
@@ -0,0 +1,99 @@
...
+## Declaring Service Dependencies
+
+#### Motivation
+
+Istio's [default behavior](https://istio.io/latest/docs/reference/config/networking/sidecar/)
+is to configure all sidecar proxies in the mesh with all
+necessary information required to send traffic to any destination in the mesh. While useful
+for day one operations, this comes with a performance tradeoff that scales with the number of workloads and destinations—
+proxies may be maintaining unnecessary routing information in memory, which in aggregate may lead
+to a large memory footprint for the entire data plane.
```
linsun, line 31:
```suggestion
necessary information required to send traffic to any destination in the mesh. ```suggestion
necessary information required to send traffic to any destination in the mesh. While this is helpful when getting started with Istio, this is not recommended in production when you have more than a few services.  As the number of workloads and destinations changes, proxies may be maintaining unnecessary information in memory, which in aggregate will lead to a large memory footprint for the entire data plane.
```

linsun: I removed routing because if you say routing, you'll need to spell out routes, endpoints, listeners and clusters.

docs/content/guides/service_dependencies.md
```diff
@@ -0,0 +1,99 @@
...
+
+If the operator of a service mesh has a priori knowledge of which destinations a particular workload
```
linsun, line 33:
```suggestion
If the operator of a service mesh has a prior knowledge of which destinations a particular workload
```

linsun: A bit concerned on operator here as user of service dependency...

docs/content/guides/service_dependencies.md
```diff
@@ -0,0 +1,99 @@
...
+      app: productpage
+      version: v1
+```
+
+#### Semantics
+
+1. An important property to note about the ServiceDependency API is that objects are *additive*. 
+   If multiple ServiceDependencies target the same workload, that workload's egress dependencies will 
+   consist of the set union of all destinations declared on ServiceDependencies.
+
```
linsun, line 96: Is this true?  if service consumer (client) are in different ns, it should be fine, I think, no need to union.

docs/content/guides/service_dependencies.md
```diff
@@ -0,0 +1,99 @@
...
+2. Depending on the underlying mesh (such as Istio), declaring a ServiceDependency for a 
+   workload may prevent that workload from sending traffic to any destination other than those 
+   explicitly declared. Keep this in mind as you create these objects.
```
linsun, line 99: We should note this is not a security feature.  It just doesn't provide the configuration of sidecar to reach from client to server but users can still hack into the sidecar and configure it manually.

docs/content/guides/service_dependencies.md
```diff
@@ -0,0 +1,99 @@
+---
+title: Service Dependencies
+menuTitle: Service Dependencies
+weight: 30
+---
+
+{{% notice note %}} Gloo Mesh Enterprise is required for this feature. {{% /notice %}}
```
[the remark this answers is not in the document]
linsun: Since this doesn't configure exportTo, I think this API is for service consumers.

docs/content/guides/service_dependencies.md
harveyxia: This is true—we aggregate all egress dependencies for a given workload

docs/content/guides/service_dependencies.md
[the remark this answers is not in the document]
harveyxia: We'll have a clearer picture of the personas and use cases for the ServiceDependency API once we get some initial user feedback, after which we'll update this guide with some use case examples

docs/content/guides/service_dependencies.md
linsun: This is still confusing to me, could be further clarified in a follow up PR.  would be good to clarify workload is client or server (destination workload) and what the sidecar egress looks like in this case when two clients target the same destination workload. My understanding is that it would only impact client's egress, not server's.

linsun:
LGTM for an initial pass to get more feedback: 1) persons, 2) two clients depend on same server service.

docs/content/guides/service_dependencies.md
harveyxia: > My understanding is that it would only impact client's egress (since we have two clients thus no union), not server's.

This is correct. I've added a line to clarify this important point

Issue #1584, opened 2021-05-25
meshctl configuration file


ashleywang1:
meshctl should be made to support a configuration file with default values. this is especially useful so that a user can predefine a mapping of clusters to kubeconfig entries like so:
```yaml
clusters:
  managementPlane: #reserved special name for the management cluster
    kubeConfig: /Users/ilackarms/.kube/config
    kubeContext: kind-mgmt-cluster
  remote-cluster: # data plane cluster
    kubeConfig: /Users/ilackarms/.kube/config
    kubeContext: kind-remote-cluster
  mgmt-cluster: # mgmt cluster is also registered as data plane
    kubeConfig: /Users/ilackarms/.kube/config
    kubeContext: kind-mgmt-cluster
```
We can store this file in the user's  home dir, e.g.`${HOME}/.meshctl/config.yaml` via a `meshctl configure` command (interactively configures meshctl like `aws configure`). We can then reference this file in subsequent `meshctl` commands which automate cluster interaction rather than requiring the user to always pass their kubeconfig/context for each cluster individually. it also makes it much easier to support commands that interact with multiple clusters at once.

ashleywang1:
The configuration file should go under `.gloo-mesh/meshctl-config.yaml`

ashleywang1:
The design doc for this file is here: https://soloio.slab.com/posts/meshctl-config-file-design-doc-2qwo134j

This file should also be updated automatically upon `meschtl install` and `meshctl cluster register`.

Sodman:
Needs user facing docs.

PR #1609, opened 2021-06-02
Meshctl cluster configure

ashleywang1:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1584


pkg/meshctl/commands/cluster/configure/cluster_configure.go
```diff
@@ -0,0 +1,94 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return configure(opts)
+		},
+	}
+	opts.addToFlags(cmd.PersistentFlags())
+	return cmd
+}
+
+type options struct {
+	MeshctlConfigPath string
```
ilackarms, line 29: any reason to not make this private?

pkg/meshctl/commands/cluster/configure/cluster_configure.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func (o *options) addToFlags(flags *pflag.FlagSet) {
+	flags.StringVarP(&o.MeshctlConfigPath, "meshctl-config-file", "f", "", "path to the meshctl config file")
+}
+
+func configure(opts *options) error {
+	var err error
+	if opts.MeshctlConfigPath == "" {
```
ilackarms, line 38: nit: instead of reading and writing to the `opts` that get passed in, i'd suggest just capturing `opts.MeshctlConfigPath` in a variable and setting/getting from that

pkg/meshctl/commands/cluster/configure/cluster_configure.go
```diff
@@ -0,0 +1,94 @@
...
+type options struct {
+	MeshctlConfigPath string
+}
+
+func (o *options) addToFlags(flags *pflag.FlagSet) {
+	flags.StringVarP(&o.MeshctlConfigPath, "meshctl-config-file", "f", "", "path to the meshctl config file")
+}
+
+func configure(opts *options) error {
+	var err error
```
ilackarms, line 37: nit: this error can be scoped to the `if` block below

pkg/meshctl/commands/cluster/configure/config.go
```diff
@@ -0,0 +1,44 @@
...
+	ApiVersion string                 `json:"apiVersion"`
+	Clusters   map[string]KubeCluster `json:"clusters"`
+}
+
+type KubeCluster struct {
+	KubeConfig  string `json:"kubeConfig"`
+	KubeContext string `json:"kubeContext"`
+}
+
+func ParseMeshctlConfig(fileName string) (Config, error) {
```
ilackarms, line 23: any reason this is public?

ashleywang1: I think it'll be used elsewhere - for example, in the `meshctl debug bug-report` command.

EItanya: I agree with @ilackarms, make it public when needed, it's harder to switch back

pkg/meshctl/commands/cluster/configure/cluster_configure.go
```diff
@@ -0,0 +1,225 @@
...
+
+func configureInteractive(meshctlConfigPath string) error {
+	config, err := utils.ParseMeshctlConfig(meshctlConfigPath)
+	if err != nil {
+		return err
+	}
+
+	keepGoing := "Yes"
+	for keepGoing == "Yes" {
+		answer, err := selectValueInteractive("Are you configuring a management cluster or a data plane cluster?",
```
EItanya, line 104: Can you configure a cluster to be both, or would they have to add it as both seperately?

pkg/meshctl/commands/cluster/list/cluster_list.go
```diff
@@ -0,0 +1,97 @@
...
+
+func Command(ctx context.Context) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "List all Kubernetes cluster registered with Gloo Mesh",
+	}
+
+	cmd.AddCommand(
+		communityCommand(ctx),
+		enterpriseCommand(ctx),
```
EItanya, line 35: Can we infer this from the networking deployment type? seems overly verbose

pkg/meshctl/commands/cluster/configure/cluster_configure.go
ashleywang1: Currently they would have to add it as both separately

pkg/meshctl/commands/cluster/configure/config.go
ashleywang1: I've moved it to a `utils` package - it will need to be public there.

pkg/meshctl/commands/cluster/list/cluster_list.go
```diff
@@ -0,0 +1,97 @@
...
+				return errors.Wrapf(err, "Failed to list clusters.")
+			}
+			for _, c := range kubernetesClusterList.Items {
+				fmt.Printf("%s\n", c.GetName())
+			}
+			return nil
+		},
+	}
+	opts.addToFlags(cmd.Flags())
+	cmd.SilenceUsage = true
```
EItanya, line 89: ?

pkg/meshctl/commands/cluster/list/cluster_list.go
```diff
@@ -0,0 +1,97 @@
...
+				if string(s.Type) == clusterSecretType {
+					fmt.Printf("%s\n", s.GetName())
+				}
+			}
+			return nil
+		},
+	}
+
+	opts.addToFlags(cmd.Flags())
+	cmd.SilenceUsage = true
```
EItanya, line 63: ?

pkg/meshctl/commands/cluster/list/cluster_list.go
ashleywang1: I agree that this is silly, we should just scan for both community and enterprise clusters and display them if they exist.

ashleywang1:
Output of `meshctl cluster list` looks like:
```
+----------------+-----------+
|    CLUSTER     | NAMESPACE |
+----------------+-----------+
| mgmt-cluster   | gloo-mesh |
+----------------+-----------+
| remote-cluster | gloo-mesh |
+----------------+-----------+
```

pkg/meshctl/commands/cluster/configure/cluster_configure.go
```diff
@@ -0,0 +1,227 @@
...
+		Long: `Create a mapping of clusters to kubeconfig entries in ${HOME}/.gloo-mesh/meshctl-config.yaml.
+
+There are two modes for this - interactive and non-interactive. Each data plane cluster should be configured with
+a cluster name. Note that if a cluster is both a management and data plane cluster, it will need to be configured twice.`,
+		Example: " meshctl cluster configure --disable-prompt --kubeconfig ${HOME}/.kube/config --kubecontext cluster1 ## Registers a management plane cluster\n" +
+			" meshctl cluster configure --disable-prompt --cluster-name cluster2 --kubeconfig ${HOME}/.kube/config --kubecontext cluster2 ## Registers a data plane cluster",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if opts.disablePrompt {
+				if opts.kubeconfig == "" || opts.kubecontext == "" {
+					return eris.Errorf("must pass in additional flags when configuring in non-interactive mode")
```
Sodman, line 32: Nit: Could we be a little more explicit here so the user knows which missing flags are required? Something like "Must pass in either --kubeconfig or --kubecontext when configuring in non-interactive mode"?

Sodman (commented):
LGTM, left one nit comment, let me know what you think. Otherwise ready to approve.

Sodman (approved):

Issue #1450, opened 2021-04-16
Expose helm options for setting extra load balancer annotations


jameshbarton:
**Is your feature request related to a problem? Please describe.**
A Gloo Mesh Enterprise customer who also uses Gloo Edge with Amazon EKS wants to be able to specify extra annotations in helm charts just as they can with [Edge](https://docs.solo.io/gloo-edge/latest/guides/integrations/aws/#how-to).  Specifically, they want to be able to specify AWS load balancer annotations like these:
```
service.beta.kubernetes.io/aws-load-balancer-type: nlb
service.beta.kubernetes.io/aws-load-balancer-internal: "true"
```

**Describe the solution you'd like**
Use `extraAnnotations` support in Gloo Mesh similar to what exists in Edge today.

**Additional context**
Internal reference: https://solo-io.slack.com/archives/G01CB20BFR8/p1618495032029000

ashleywang1:
This should now be possible in Gloo Mesh Enterprise v1.1.0-beta7, e.g.
```
enterpriseNetworking:
  service:
    extraAnnotations:
      - service.beta.kubernetes.io/aws-load-balancer-type: nlb
```

ashleywang1:
This has been released in Gloo Mesh oss v1.0.11

harveyxia:
@ashleywang1 can you cut an enterprise release for this? thanks

harveyxia:
@EItanya confirm that this is available in enterprise

harveyxia:
@ashleywang1 

This change exists on `skv2@v0.17.19`, which needs to be brought into GM and GME, on both v1.0.x and v1.1.x branches

ashleywang1:
GM v1.1.0-beta11 branch PR: https://github.com/solo-io/gloo-mesh/pull/1612

ashleywang1:
A generalized way of doing this was done in skv2@v0.17.20. The following helm override file will allow you to override any operator-generated deployment or service for the gloo mesh components, e.g.
```
discovery:
  deploymentOverrides:
    spec:
      replicas: 5
      template:
        metadata:
          annotations:
            test: new-annotation
            app.kubernetes.io/name: test-replacing-annotation
```

harveyxia:
Implemented in https://github.com/solo-io/gloo-mesh/pull/1612 for OSS

Implemented in https://github.com/solo-io/gloo-mesh-enterprise/pull/644 for GME

ashleywang1:
[...]

PR #1612, opened 2021-06-02
Bump skv2 to v0.17.20

ashleywang1:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1450


install/helm/gloo-mesh/templates/deployment.yaml
```diff
@@ -101,12 +130,25 @@ spec:
 # Deployment manifest for networking
 {{- $networking := $.Values.networking}}
 {{- $networkingImage := $networking.image }}
+{{- $networkingExtraPodLabels := $networking.extraPodLabels -}}
+{{- $networkingExtraPodAnnotations := $networking.extraPodAnnotations -}}
+{{- $networkingExtraDeploymentLabels := $networking.extraDeploymentLabels -}}
+{{- $networkingExtraDeploymentAnnotations := $networking.extraDeploymentAnnotations -}}
```
harveyxia, line 136: When you run `helm template` on the packaged chart, you'll notice the following:

```yaml
# Source: gloo-mesh/templates/deployment.yaml
# Deployment manifest for networkingapiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: networking
  name: networking
  namespace: default
```

It looks like the nil `networking.extraDeploymentAnnotations` is causing a newline to be absent, we need to make sure the newline is preserved. Maybe try remove the trailing dash in `-}}`?

install/helm/gloo-mesh/templates/_helpers.tpl
```diff
@@ -0,0 +1,54 @@
+# Code generated by skv2. DO NOT EDIT.
```
mlholland, line 1: What is causing this helper do be generated? I didn't see anything that looked like it would prompt this.

ashleywang1: It's a new template in skv2, and is run when we run the generate.go file. (See: https://github.com/solo-io/skv2/blob/master/codegen/templates/chart/_helpers.tpl)

mlholland (approved):

PR #1626, opened 2021-06-08
[Backport] bump skv2

ashleywang1:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1450


docs/content/reference/helm/gloo_mesh/latest/_index.md
```diff
@@ -1,6 +1,6 @@
 
 ---
-title: "v1.1.0-beta10"
+title: "v1.1.0-beta11"
```
mlholland, line 3: I'm not quite sure how we approach docs changes for backports, but should this be changed to match the version this backport will produce, or does it not matter both because this is the 'latest' page? It might also not matter because I assume the 'latest' docs are produced by a more recent version than this.

mlholland (approved):

PR #1636, opened 2021-06-10
added security tests and cleanup of crds on clusters

nmnellis:
more integration tests. no impact on core codebase


harveyxia (commented):
I'm not familiar with the test framework being used here—do we need to add all of the CRDs to the repo?

harveyxia (approved):

Issue #1550, opened 2021-05-13
VirtualDestination 503 error


nmnellis:
**Describe the bug**
2 cluster setup where MP and agent are on cluster-0 and another agent on cluster-1. Deployed an application called "backend" to cluster-1 and created a [VirtualDestination](https://github.com/solo-io/gloo-mesh/blob/istio-e2e-demo/pkg/test/manifests/gloo-mesh/in-mesh/virtual-destination-single-cluster-http.yaml) to expose the application on http-backend.solo.io. Calling from a client in cluster-0 that has a mesh sidecar. i get a 503.
```
root@nginx-client:/# curl -v http://http-backend.solo.io:8090/info
*   Trying 252.68.23.155...
* TCP_NODELAY set
* Expire in 200 ms for 4 (transfer 0x556ad4ac5fb0)
* Connected to http-backend.solo.io (252.68.23.155) port 8090 (#0)
> GET /info HTTP/1.1
> Host: http-backend.solo.io:8090
> User-Agent: curl/7.64.0
> Accept: */*
>
< HTTP/1.1 503 Service Unavailable
< content-length: 84
< content-type: text/plain
< date: Thu, 13 May 2021 15:13:51 GMT
< server: envoy
<
* Connection #0 to host http-backend.solo.io left intact
upstream connect error or disconnect/reset before headers. reset reason: local reset
```

* cluster-1 gateway logs
```
[2021-05-13T14:32:28.236Z] "- - -" 0 NR - - "-" 0 0 0 - "-" "-" "-" "-" "-" - - 10.42.0.8:15443 10.42.0.1:30871 - -
[2021-05-13T14:32:30.601Z] "- - -" 0 NR - - "-" 0 0 0 - "-" "-" "-" "-" "-" - - 10.42.0.8:15443 10.42.0.1:40474 - -
[2021-05-13T14:32:30.601Z] "- - -" 0 NR - - "-" 0 0 0 - "-" "-" "-" "-" "-" - - 10.42.0.8:15443 10.42.0.1:40474 - **-**
```

* Istiod logs
```json
    "pilot_conflict_inbound_listener": {
        "backend-v1-7899c8f85c-jttc8.app-1-42782": {
            "proxy": "backend-v1-7899c8f85c-jttc8.app-1-42782",
            "message": "Conflicting inbound listener:8090. existing: backend.app-1-42782.svc.cluster.local, incoming: http-backend.solo.io"
        }
    }
}
```

Use case
[...]

PR #1640, opened 2021-06-11
default federation selectors

nmnellis:
If no federation values selected, default federated services to all.

trying to remove this block from being required. it should just be defaulted on
```
  federation:
    selectors:
      - { }
```
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1550


Sodman (changes requested):
We should talk about this as it's a breaking change for current behavior.

Elsewhere for selectors, the standard has been:
- If selectors not specified / nil - select none
- When selector is set to empty object - select all
- Specific selectors defined - select only those

I'm open to changing it if we want a different behavior, but we'll need to change it in a few places to be consistent and communicate it as a breaking change next release if we do.

Issue #1611, opened 2021-06-02
Support multiple Istio ingress gateway workloads


harveyxia:
- [x] **Discovery:** Support discovering multiple Istio ingress gateway workload IPs
  - [x] Update discovery to capture all externally addressable k8s Services (implemented in https://github.com/solo-io/gloo-mesh/pull/1643)
  - [x] Allow discovering ingress gateway workloads in namespaces other than the hardcoded `istio-system`, [issue here] (https://github.com/solo-io/gloo-mesh-enterprise/issues/664) (implemented in https://github.com/solo-io/gloo-mesh/pull/1643)
  - [x] Mark as deprecated istio ingress gateway discovery settings (implemented in https://github.com/solo-io/gloo-mesh/pull/1643)
- [x] **Networking:** Where relevant in translated outputs, support default behavior of load balancing across those ingress gateway workload IPs, else implement the desired APIs described below:
  - [x] **Virtual Mesh:** [allow selection of desired ingress gateway workloads for a given VM](https://github.com/solo-io/gloo-mesh-enterprise/issues/659)
  - [x] **VirtualGateway:** allow selection of desired ingress gateway for N/S traffic
 - [x] Deprecate usage of old `Mesh.Spec.IngressGatewayInfo` and replace with `Mesh.Status.AppliedEastWestIngressGateway`
- [x] **Docs**:
  - [x] Add docs for new VirtualMesh ingress gateway selector

harveyxia:
Discovery requirements from @christian-posta:

1. We can't assume that all ingress gateway workloads have a common label pair, so we'll need OR semantics between sets of labels
2. We also need to expose ability to configure ingress gateway namespace, currently it's hardcoded to `istio-system`

harveyxia:
When routing to externally accessible services, the `<IP:Port>` pair values depend on the Service type, which has some nuance [described here](https://stackoverflow.com/questions/41509439/whats-the-difference-between-clusterip-nodeport-and-loadbalancer-service-types/41510604#41510604)

harveyxia:
The remaining work is on the VirtualGateway, tracked in this issue, https://github.com/solo-io/gloo-mesh-enterprise/issues/857

Sodman:
All tasks in this list have been completed & merged into main - will be part of 1.1 release.

PR #1643, opened 2021-06-14
Update Discovery to capture all externally addressable k8s services

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1611
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/664


pkg/mesh-discovery/translation/destination/detector/destination_detector.go
```diff
@@ -323,3 +326,144 @@ func convertPorts(service *corev1.Service) (ports []*v1.DestinationSpec_KubeServ
...
+	}
+
+	return externalNodeIps
+}
+
+// return true is the Node is provisioned by Kind
+func isKindNode(node *corev1.Node) bool {
+	for _, image := range node.Status.Images {
+		for _, name := range image.Names {
+			if strings.Contains(name, "kindnetd") {
```
EItanya, line 447: So when installing with flat-networking we actually don't install this component, so I'm wondering if there's a more consistent way to check if it's a `kind` cluster?

harveyxia: Interesting—this is the same as how we're [currently doing it](https://github.com/solo-io/gloo-mesh/blob/c4c3a3688cbbed4f94b782c1d0c3a16a29a1b853/pkg/mesh-discovery/translation/mesh/detector/istio/istio_mesh_detector.go#L326-L335). The flat networking e2e tests work fine, so perhaps it's not an issue?

EItanya: If it's how we're currently doing it, let's keep it. flat-networking doesn't use the gateway, so it probably works because it doesn't need the gateway.

api/discovery/v1/destination.proto
```diff
@@ -57,6 +57,23 @@ message DestinationSpec {
...
+        repeated ExternalAddress external_addresses = 8;
+
+        // Describes the address data for Kubernetes Services exposed to external traffic (i.e. for non ClusterIP type Services).
+        message ExternalAddress {
+
+            // The externally-reachable address for this Kubernetes Service, either an IP address or DNS name.
+            oneof external_address_type {
+
+                // Populated for LoadBalancer Services with specified DNS name.
+                string dns_name = 5;
```
EItanya, line 70: I think we may need more information here than just this. For instance, the port on which this is accessible. A `NodeIP` is only accessible on the `NodePort` whereas a `LoadBalancerIP` is not accessible on the `NodePort`

harveyxia: Good point—so we just need to also capture the service type. Any other data you think we need?

EItanya (approved):

EItanya (approved):

Issue #1562, opened 2021-05-19
meshctl check falsely reports that relay pull clients are not connected


joekelley:
`meshctl check` will incorrectly indicate that relay pull clients aren't connected when the dashboard is installed in cluster. This is because the dashboard creates a pull client for every registered cluster, causing this check to fail: https://github.com/solo-io/gloo-mesh/blob/f32de9660773a1ba1d0f53ede44021057d0f5ec9/pkg/meshctl/checks/enterprise_registration_check.go#L277

We should resolve this in the relay server code where we generate the connected client stat. There, we can differentiate between clients connected to the relay pull server used for multicluster connectivity and that used for the dashboard.

PR #1651, opened 2021-06-16
meshctl bug-report

nmnellis:
Copied over the libraries from Istio so that i could modify them to add gloo components and multi-cluster support!

added
* support for multiple kubeconfigs and multiple contexts
* gloo mesh container logs
* gloo mesh replicaset describe
* gloo mesh versions

Usage 
```
meshctl bug-report --kubeconfig /Users/nick/.kube/mp,/Users/nick/.kube/cp-us-east
```
Screen Shot 2021-06-15 at 9 32 36 PM

BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1562


pkg/meshctl/bug-report/pkg/testdata/output/multi_line_entries.log
```diff
@@ -0,0 +1,20 @@
+2020-06-29T23:37:27.287895Z	info	Effective config: binaryPath: /usr/local/bin/envoy
```
ashleywang1, line 1: Delete all the testdata/ directory files

rvennam:
this is amazing...we need this asap.

does this capture all the gloo mesh cr's? meshes, kubernetesclusters, destinations, settings, workloads, etc

nmnellis:
> this is amazing...we need this asap.
> 
> does this capture all the gloo mesh cr's? meshes, kubernetesclusters, destinations, settings, workloads, etc

yes it gathers al CRs in a crs file

EItanya:
Let's combine this with @ashleywang1's work on the meshctl config file

pkg/meshctl/bug-report/pkg/kubeclient/kubeclient.go
```diff
@@ -0,0 +1,76 @@
+// Copyright Istio Authors
```
harveyxia, line 1: We already have logic to build kube clients from [this pkg](https://github.com/solo-io/gloo-mesh/blob/580e9211778cc11ce8900f9e3e62919d36aad50d/pkg/meshctl/utils/utils.go#L16), can we remove this pkg and use that one instead?

harveyxia (commented):
I was going to suggest selectively importing this code and extending with custom logic, but reading through the copied istio code, it's written in a way that makes that difficult.

Given how it's implemented, long term this may be difficult to extend—we should discuss our options and approach for this

ashleywang1:
> Let's combine this with @ashleywang1's work on the meshctl config file

I'll push updates to this PR to add the following changes:
- adding support for gloo mesh oss logs
- moving the command to meshctl debug report (this way, it's under our debug suite)
- adding support for the meshctl config file

nmnellis:
> Given how it's implemented, long term this may be difficult to extend—we should discuss our options and approach for this

you are right that it might be hard to maintain long term but this functionality is desperately needed on the field. we can always iterate

harveyxia:
> you are right that it might be hard to maintain long term but this functionality is desperately needed on the field. we can always iterate

That's a fair point. In that case, if we're going to wholesale copy, could we move all the copied pkgs into an `internal` pkg so that the copied public functions aren't accidentally used elsewhere in our code base?

nmnellis:
> That's a fair point. In that case, if we're going to wholesale copy, could we move all the copied pkgs into an `internal` pkg so that the copied public functions aren't accidentally used elsewhere in our code base?

that seems like a good idea

pkg/meshctl/commands/debug/report/internal/bugreport/bugreport.go
```diff
@@ -0,0 +1,687 @@
...
+		DryRun: config.DryRun,
+	}
+	common.LogAndPrintf("\nFetching Istio control plane information from cluster.\n\n")
+	getFromCluster(content.GetK8sResources, params, clusterDir, &mandatoryWg)
+	getFromCluster(content.GetCRs, params, clusterDir, &mandatoryWg)
+	getFromCluster(content.GetEvents, params, clusterDir, &mandatoryWg)
+	getFromCluster(content.GetClusterInfo, params, clusterDir, &mandatoryWg)
+	getFromCluster(content.GetSecrets, params.SetVerbose(config.FullSecrets), clusterDir, &mandatoryWg)
+	getFromCluster(content.GetDescribePods, params.SetIstioNamespace(config.IstioNamespace), clusterDir, &mandatoryWg)
+	// Gloo mesh describe pods
```
ashleywang1, line 373: This is the place where Gloo Mesh extensions are happening

pkg/meshctl/commands/debug/report/internal/bugreport/bugreport.go
```diff
@@ -0,0 +1,687 @@
...
+type kubeCombo struct {
+	context  string
+	filepath string
+}
+
+// this is kinda tricky because you can provide the following combinations
+// no kubeconfig no context (use default for both)
+// no kubeconfig, list of contexts (meaning use default kubeconfig)
+// all kubeconfigs, no contexts (meaning use default context in each kubeconfig)
+// and equal list of kubeconfigs and contexts (use each cooresponding context per kubeconfig)
```
ashleywang1, line 207: fix typo

pkg/meshctl/commands/debug/report/internal/bugreport/bugreport.go
```diff
@@ -0,0 +1,708 @@
...
+		if err != nil {
+			return err
+		}
+		resources, err := cluster2.GetClusterResources(context.Background(), clientset)
+		if err != nil {
+			return err
+		}
+
+		dumpGlooMeshVersions(kubeConfigPath, kubeContext, config.GlooMeshNamespace)
+		dumpMeshctlCheck(kubeConfigPath, kubeContext, config.GlooMeshNamespace)
```
ashleywang1, line 151: This is another place where Gloo Mesh extensions are happening

ashleywang1 (approved):

PR #1666, opened 2021-06-18
set ServiceEntry resolution to STATIC if any endpoint IPs are ipv6

harveyxia:


pkg/mesh-networking/translation/istio/destination/federation/federation_translator.go
```diff
@@ -374,3 +375,19 @@ func getHostnameSuffix(hostname string) string {
...
+// Workaround for Istio issue where ipv6 addresses are supplied to Envoy incorrectly: https://github.com/envoyproxy/envoy/issues/10489#issuecomment-606290733.
+// If any endpoints have an ipv6 address, set the resolution to STATIC,
+// else, set the resolution to DNS.
+// exported for use in enterprise
+func ResolutionForEndpointIpVersions(
+	workloadEntries []*networkingv1alpha3spec.WorkloadEntry,
+) networkingv1alpha3spec.ServiceEntry_Resolution {
+	for _, workloadEntry := range workloadEntries {
+		// if endpoint addr is an IP (instead of hostname) and ipv6 format, return STATIC
+		if ip := net.ParseIP(workloadEntry.Address); ip != nil && ip.To4() == nil {
```
harveyxia, line 388: logic based on this [stackoverflow answer](https://stackoverflow.com/a/22752227/2255272)

EItanya: This comment claims that function can have false positives https://stackoverflow.com/a/48519490

EItanya: Not sure of the validity, just want to be sure

harveyxia: discussed offline—this logic does not handle ipv4-mapped ipv6 addresses so we'll change the logic to simply detect ipv6 if there are two or more colons present in the string

EItanya (approved):

PR #1673, opened 2021-06-22
report error if SE endpoints have both ipv6 and hostnames

harveyxia:


pkg/mesh-networking/translation/istio/destination/federation/federation_translator.go
```diff
@@ -381,13 +397,27 @@ func getHostnameSuffix(hostname string) string {
...
 		if strings.Count(workloadEntry.Address, ":") >= 2 {
-			return networkingv1alpha3spec.ServiceEntry_STATIC
+			foundIpv6 = true
 		}
+		if ip := net.ParseIP(workloadEntry.Address); ip == nil {
+			foundHostname = true
+		}
+	}
+
+	if !foundHostname {
```
EItanya, line 413: This conditional is a bit hard to read, can we switch it to:
```
if (foundHostname && foundIpv6) {
    return networkingv1alpha3spec.ServiceEntry_NONE, eris.New("endpoints contain both ipv6 and hostname addresses")
} else if (foundHostname && !foundIpv6) {
  return networkingv1alpha3spec.ServiceEntry_DNS, nil
}
return networkingv1alpha3spec.ServiceEntry_STATIC, nil 

```

pkg/mesh-networking/translation/istio/destination/federation/federation_translator.go
```diff
@@ -225,6 +231,11 @@ func (t *translator) translateRemoteServiceEntryTemplate(
 	federatedHostname := destination.Status.AppliedFederation.GetFederatedHostname()
 
 	// ObjectMeta's Namespace and ClusterName will be populated when translating federated outputs
```
EItanya, line 233: This comment should be moved down as it has to do with the `ServiceEntry`

pkg/mesh-networking/translation/istio/destination/federation/federation_translator_test.go
```diff
@@ -354,21 +354,32 @@ var _ = Describe("FederationTranslator", func() {
...
 				Address: "192.168.21.2",
 			},
-			{
-				Address: "hostname",
-			},
 		}
-		Expect(federation.ResolutionForEndpointIpVersions(workloadEntries)).To(Equal(networkingv1alpha3spec.ServiceEntry_DNS))
+		res, err := federation.ResolutionForEndpointIpVersions(workloadEntries)
+		Expect(err).To(BeNil())
+		Expect(res).To(Equal(networkingv1alpha3spec.ServiceEntry_STATIC))
```
EItanya, line 360: Can we test `DNS` since that default was removed

EItanya (approved):

PR #1676, opened 2021-06-22
Cert Rotation Api

EItanya:


api/discovery/v1/mesh.proto
```diff
@@ -206,4 +207,17 @@ message MeshStatus {
...
         repeated string errors = 3;
     }
+
+    message AppliedCA {
+
+        // Certificate which is currently applied to the mesh
+        .networking.mesh.gloo.solo.io.SharedTrust applied_shared_trust = 1;
+
+        
+        // Reserve for limited trust when it's implemented
```
mlholland, line 217: What's the merit for reserving an index like this as opposed to setting cert_rotation_conditions as value 2, and implementing limited trust as value 3 when its time comes?

pkg/mesh-networking/translation/istio/mesh/mtls/mtls_translator.go
```diff
@@ -119,6 +119,24 @@ func (t *translator) Translate(
...
+		// applied is different from VM spec, time to rotate
+
+		// Check conditions to see what needs to be done.
+		if len(vmStatus.GetCertRotationConditions()) == 0 {
+			// No conditions yet, begin rotation
+			// Build issuedCert with current + next CA
+			// Phase: ADDING_NEW_ROOT
+		} else {
+			// Check previous rotation condition to see what to do.
+
```
mlholland, line 136: should this have a TODO, or are you planning on filling this in

pkg/mesh-networking/reconciliation/networking_reconciler.go
```diff
@@ -229,7 +229,7 @@ func (r *networkingReconciler) reconcile(obj ezkube.ResourceId) (bool, error) {
 			Verifier: r.remoteResourceVerifier,
 		}
 		userSupplied, err = r.remoteBuilder.BuildSnapshot(ctx, "mesh-networking-istio-inputs", input.RemoteBuildOptions{
-			IssuedCertificates:    resourceBuildOptions,
+			// IssuedCertificates:    resourceBuildOptions,
```
EItanya, line 232: This is a temporary hack to get remote IssuedCertificates into the snapshot. @ilackarms @harveyxia we need to come up with a more permanent way to do it

api/certificates/v1/ca_options.proto
```diff
@@ -33,4 +34,83 @@ message IntermediateCertificateAuthority  {
...
   
-}
\ No newline at end of file
+}
+
+// State of Certificate Rotation
+// Possible states in which a CertificateRotation can exist.
+enum CertificateRotationState {
+  // No Certificate rotation is currently happening
+  NOT_APPLICABLE = 0;
```
ilackarms, line 43: wdyt `PENDING`?

api/certificates/v1/ca_options.proto
```diff
@@ -33,4 +34,83 @@ message IntermediateCertificateAuthority  {
...
+
+  // The CertificateRotation is underway, both roots are set, and the new root is being propogated
```
ilackarms, line 45: `The CertificateRotation is underway` -> `The CertificateRotation has begun`

api/certificates/v1/ca_options.proto
```diff
@@ -33,4 +34,83 @@ message IntermediateCertificateAuthority  {
...
+
+message CertificateRotationVerificationMethod {
+  oneof method {
+    // Verification not enabled. NOTE: This setting is only recommended for testing.
+    // When enabled rotation will continue from step to step without any kind of verification.
+    google.protobuf.Empty none = 1;
+
+    // Verification must be completed manually. This involves using our certificate verification 
+    // endpoint when the certificates are in a VERIFYING state
+    google.protobuf.Empty manual = 2;
```
ilackarms, line 84: why are these empties? why not use an enum?

api/certificates/v1/ca_options.proto
```diff
@@ -33,4 +34,83 @@ message IntermediateCertificateAuthority  {
...
+  }
+}
+
+enum CertificateRotationMethod {
+  MULTI_ROOT = 0;
+
+  NONE = 1;
+
+  // TODO: Implement
+  CROSS_SIGNED = 2;
```
ilackarms, line 94: comments

api/certificates/v1/ca_options.proto
EItanya: I opted for NOT_APPLICABLE, because it's not really PENDING. It actually signals that there is no rotation, and there won't be any happening. Definitely open to a different name though.

api/certificates/v1/ca_options.proto
```diff
@@ -33,4 +34,99 @@ message IntermediateCertificateAuthority  {
...
+    // When enabled rotation will continue from step to step without any kind of verification.
+    google.protobuf.Empty none = 1;
+
+    // Verification must be completed manually. This involves using our certificate verification 
+    // endpoint when the certificates are in a VERIFYING state
+    google.protobuf.Empty manual = 2;
+  }
+}
+
+enum CertificateRotationStrategy {
```
harveyxia, line 92: Are there any other rotation strategies? If not, we can make this a boolean and name it accordingly, like `useMultirootRotation` or something better

api/certificates/v1/ca_options.proto
```diff
@@ -33,4 +34,99 @@ message IntermediateCertificateAuthority  {
...
+
+  // The rotation has finished, the new root has been propagated to all data-plane clusters, and traffic has
+  // been verified successfully.
+  FINISHED = 8;
+
+  // Processing the certificate rotation workflow failed.
+  FAILED = 9;
+}
+
+message CertificateRotationVerificationMethod {
```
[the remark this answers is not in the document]
EItanya: Stats based verification. So awaiting/listening to certain stats which the user indicates as as healthy traffic

api/certificates/v1/ca_options.proto
EItanya: Yes there are, there is a method called [cross-signing](https://www.consul.io/docs/connect/ca#root-certificate-rotation). Only certain certificates support this method though, and the `MULTI_ROOT` is more universal, so we implemented it first. Cross signing is only 2 steps, so it is better to do when possible

EItanya:
blocked by: https://github.com/solo-io/gloo-mesh/pull/1739

harveyxia (approved):

PR #1677, opened 2021-06-22
Virtualmesh keepalive

mlholland:
Adds a TCPKeepalive value to virtual mesh configs, which is then applied to all istio destinations in under this virtual mesh.
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1678


pkg/mesh-networking/translation/istio/destination/destinationrule/destination_rule_translator.go
```diff
@@ -150,6 +151,39 @@ func (t *translator) Translate(
 		return nil
 	}
 
+	// retrieve the virtual mesh to use its keepalive value, if it exists.
```
EItanya, line 154: Maybe I misunderstand/misunderstood, but I thought we talked earlier about only adding the KeepAlive for our generated multi-cluster `DestinationRules`. Those `DestinationRules` are created in the `federation_translator`. @harveyxia please correct me if I'm wrong

mlholland: @EItanya  That is correct, `DestinationRules` are created in the `federation_translator`. The key detail is that they're made by calling a [separate translator](https://github.com/solo-io/gloo-mesh/blob/459be861fef9d44f0257309ea0995650b4debbec/pkg/mesh-networking/translation/istio/destination/federation/federation_translator.go#L370)

EItanya: But we only want TCP keep alive on the DestinationRules in that other translator, so I think we should just add the config there

Sodman: My understanding was that we only needed the `DestinationRule` on the actual ingress gateways that handled the cross-cluster traffic, and not the final remote destination that represents the target service itself.

Ie, example request flow for a requesting service calling a target service in another mesh is:
requesting service (mesh 1) -> ingress gateway (mesh 1) -> ingress gateway (mesh 2) -> target service (mesh 2)

I believe we only need the KeepAlive on the DestinationRule targeting the two ingress gateways (as this is where folks are seeing connections being dropped most frequently, between clusters), but not the requesting or target services, right?

More broadly, we want the ability to add destination-specific keepalive config to any given service in a mesh, but that's going to be out of scope for this pr, and will be handled by destinationPolicy.

api/networking/v1/virtual_mesh.proto
```diff
@@ -119,6 +123,20 @@ message VirtualMeshSpec {
...
+        // (unless overridden, Linux defaults to 9.)
+        uint32 probes = 1;
+        // The time duration a connection needs to be idle before keep-alive
+        // probes start being sent. Default is to use the OS level configuration
+        // (unless overridden, Linux defaults to 7200s (ie 2 hours.)
+        google.protobuf.Duration time = 2;
+        // The time duration between keep-alive probes.
+        // Default is to use the OS level configuration
+        // (unless overridden, Linux defaults to 75s.)
+        google.protobuf.Duration  interval   = 3;
```
Sodman, line 137: nit: whitespace
```suggestion
        google.protobuf.Duration interval   = 3;
```

api/networking/v1/virtual_mesh.proto
```diff
@@ -104,6 +108,17 @@ message VirtualMeshSpec {
...
+        // Specify a keepalive rule for all destinations associated with this VirtualMesh
+        message TCPKeepalive {
+            uint32 probes = 1;
+            // The time duration a connection needs to be idle before keep-alive
+            // probes start being sent. Default is to use the OS level configuration.
+            google.protobuf.Duration time = 2;
+            // The time duration between keep-alive probes.
+            // Default is to use the OS level configuration.
+            google.protobuf.Duration  interval   = 3;
+        }
```
harveyxia, line 121:
```suggestion
        // Configure TCP keepalive for the ingress gateways of all meshes in this VirtualMesh.
        message TcpKeepalive {
        
            // Maximum number of TCP keepalive probes to send before determining that connection is dead.
            uint32 probes = 1;
            
            // The time duration a connection needs to be idle before keep-alive probes start being sent.
            google.protobuf.Duration time = 2;
            
            // The time duration between keep-alive probes.
            google.protobuf.Duration  interval   = 3;
        }
```

harveyxia: Let's hoist this into `common` and reuse it in the Destination status and VirtualMesh

changelog/v1.1.0-beta16/virtualmesh-keepalive.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+  - type: NEW_FEATURE
+    description: Add TCPKeepalive value to virtual mesh config. This sets the corresponding value in all istio destinations associated with this virtualmesh.
```
harveyxia, line 3:
```suggestion
    description: Extend VirtualMesh API with TCP keepalive config for ingress gateways on all meshes in the VirtualMesh.
```

api/networking/v1/virtual_mesh.proto
mlholland: moved to keepalive.proto in the common directory.

changelog/v1.1.0-beta16/virtualmesh-keepalive.yaml
mlholland: Applied

mlholland:
Switching to ready for review now that I've smoke tested this. I can set virtualmesh keepalive configs and see them reach istio destinationRule configs.

api/networking/v1/virtual_mesh.proto
mlholland: Done

pkg/mesh-networking/translation/istio/destination/destinationrule/destination_rule_translator.go
```diff
@@ -133,6 +134,35 @@ func (t *translator) Translate(
...
+		}
+		tcp := destinationRule.Spec.GetTrafficPolicy().GetConnectionPool().GetTcp()
+		if tcp == nil {
+			destinationRule.Spec.TrafficPolicy.ConnectionPool.Tcp = &networkingv1alpha3spec.ConnectionPoolSettings_TCPSettings{}
+		}
+
+		// Istio uses gogo duration structs. Since we don't use gogo in our protos, we have to convert durations during runtime.
+		gogoTime := gogoutils.DurationProtoToGogo(keepalive.GetTime())
+		gogoInterval := gogoutils.DurationProtoToGogo(keepalive.GetInterval())
+		destinationRule.Spec.GetTrafficPolicy().GetConnectionPool().GetTcp().TcpKeepalive = &networkingv1alpha3spec.ConnectionPoolSettings_TCPSettings_TcpKeepalive{
```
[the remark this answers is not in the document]
mlholland: changed

pkg/mesh-networking/translation/istio/destination/destinationrule/destination_rule_translator.go
```diff
@@ -133,6 +134,35 @@ func (t *translator) Translate(
 		}
 	}
 
+	// if Destination is in a different mesh than the sourceMeshInstallation, than it's a federated Destination
```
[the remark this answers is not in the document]
mlholland: applied in its new location.

api/discovery/v1/destination.proto
```diff
@@ -296,6 +297,9 @@ message DestinationStatus {
 
         // Reference to the VirtualMesh object.
         .core.skv2.solo.io.ObjectRef virtual_mesh_ref = 4;
+
+        // Specify a keepalive rule for all destinations associated with this VirtualMesh
```
Sodman, line 301: This is only for all cross-cluster communications within the VirtualMesh, right? Not necessarily all destinations. ServiceA calling ServiceB within a mesh only gets this TCP-Keepalive if they live on different clusters. Also this applies to requests that go externalService type destinations.

Maybe update to something like:
> Specify a keepalive rule for all requests made within the VirtualMesh which cross clusters within that VirtualMesh, as well as any requests to externalService type destinations.

Functionally I think we're in the right place, just a nit on the comment!

api/networking/v1/virtual_mesh.proto
```diff
@@ -93,6 +94,9 @@ message VirtualMeshSpec {
         // If omitted, the hostname suffix defaults to "global".
         string hostname_suffix = 3;
 
+        // Specify a keepalive rule for all destinations associated with this VirtualMesh
```
Sodman, line 97: Same nit comment as above re: all destinations / vs cross cluster traffic.

pkg/mesh-networking/translation/istio/destination/destinationrule/destination_rule_translator.go
```diff
@@ -247,3 +251,36 @@ func conflictsWithUserDestinationRule(
...
 }
+
+// If this is a federated (AKA cross-cluster) destination, add keepalive values if they're present.
+// Possible Todo: refactor this code into a decorator if keepalive conflicts become possible.
+func addKeepaliveToDestinationRule(destination *discoveryv1.Destination, sourceMeshInstallation *discoveryv1.MeshSpec_MeshInstallation, destinationRule *networkingv1alpha3.DestinationRule) {
+	keepalive := destination.Status.AppliedFederation.GetTcpKeepalive()
+	// If we also have a non-nil keepalive and this is a federated destination, then extract and apply the keepalive value
+	// to the resulting destination rule.
+	// If the destination is in a different mesh than the sourceMeshInstallation, then it's a federated destination.
+	if sourceMeshInstallation != nil && destination.Spec.GetKubeService().GetRef().GetClusterName() != sourceMeshInstallation.GetCluster() && keepalive != nil {
```
Sodman, line 262: I believe this will also apply to any external service destinations, as `destination.Spec.GetKubeService()` would always be nil for those destinations. I think this is ok/expected though, we just need to make sure we call it out in the documentation (which is addressed by my top comment on the proto above!)

Sodman (commented):
The code LGTM! I'm happy to approve, just wanted a quick update to the 2 comments in protos to make it clear that requests between services in the _same_ mesh won't be affected by this config! (And also call out externalServices!)

Sodman (approved):
LGTM 🚀

PR #1689, opened 2021-06-25
Use go.mod/tools.go to lock generate deps

ryantking:


Makefile
```diff
@@ -56,17 +56,24 @@ clear-vendor-any:
...
-	mkdir -p $(DEPSGOBIN)
-	go install istio.io/tools/cmd/protoc-gen-jsonshim
-	go install github.com/golang/protobuf/protoc-gen-go
-	go install github.com/solo-io/protoc-gen-ext
-	go install github.com/golang/mock/mockgen
-	go install golang.org/x/tools/cmd/goimports
-	go install github.com/onsi/ginkgo/ginkgo
-	go install github.com/gobuffalo/packr/packr
-	go mod tidy
+mockgen: $(DEPSGOBIN)/mockgen
```
ilackarms, line 59: nit: make `PHONY` target

ilackarms: same for below

ryantking: What is the advantage to making them PHONY? Since the dependency is the binary, if the binary is present then the routine won't be run. Won't making the routines PHONY cause these binaries to always be built even if they're already present?

ryantking:
Back burnering

PR #1691, opened 2021-06-28
Multiple ingress gateways

ashleywang1:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/659


pkg/mesh-networking/apply/configtarget/config_target_validator.go
```diff
@@ -187,6 +189,56 @@ func validateOneVirtualMeshPerMesh(virtualMeshes []*v1.VirtualMesh) {
 			}
 			invalidVirtualMeshes.Insert(vMesh)
 		}
+
+		if len(vMesh.Spec.GetFederation().GetIngressGatewayServiceSelectors()) > 0 {
+			// Validate Ingress gateway service selectors
+			matchedDestinations := 0
+			for _, ingressGatewayServiceSelector := range vMesh.Spec.GetFederation().GetIngressGatewayServiceSelectors() {
+				gatewayTlsPortName := "tls"
```
ashleywang1, line 197: I'm not sure if this code can be DRY-ed, but if so, I'll use the federation default instead of making a new constant

pkg/mesh-networking/translation/istio/mesh/federation/federation_translator.go
```diff
@@ -89,33 +91,89 @@ func (t *translator) Translate(
...
-				},
-				Hosts: []string{"*." + federatedHostnameSuffix},
-				Tls: &networkingv1alpha3spec.ServerTLSSettings{
-					Mode: networkingv1alpha3spec.ServerTLSSettings_AUTO_PASSTHROUGH,
-				},
-			}},
-			Selector: ingressGateway.WorkloadLabels,
-		},
+	// one gatweay for each destination, port tuple
+	var selectedIngressGatewayList []*ingressGatewayInfo
```
ashleywang1, line 95: We create a list of (destination, tls port) tuples and each results in a gateway.

pkg/mesh-networking/translation/istio/mesh/federation/federation_translator.go
```diff
@@ -89,33 +91,89 @@ func (t *translator) Translate(
...
-	// Append the virtual mesh as a parent to each output resource
-	metautils.AppendParent(t.ctx, gw, virtualMesh.GetRef(), networkingv1.VirtualMesh{}.GVK())
-	outputs.AddGateways(gw)
+	// Each federated mesh ingress gateway will only responsible for east-west traffic in between meshes in
+	// the same virtual mesh.
+	for _, federatedMeshIngressGateway := range selectedIngressGatewayList {
+		// currently, in non flat-network mode, only TLS-secured cross cluster traffic is supported
+		gw := &networkingv1alpha3.Gateway{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:        gwName,
```
ashleywang1, line 152: Ideas on how to name this gateway if there are multiple? Maybe add the gatewayTlsPortName at the end, e.g. virtual-mesh-gloo-mesh-tls gateway?

api/common/v1/selectors.proto
```diff
@@ -118,3 +118,19 @@ message IdentitySelector {
         repeated .core.skv2.solo.io.ClusterObjectRef service_accounts = 1;
     }
 }
+
+// Select a set of Destinations with tls ports to use as ingress gateway services for the referenced Meshes.
+message IngressGatewayServiceSelector {
```
harveyxia, line 123:
```suggestion
message IngressGatewaySelector {
```

api/common/v1/selectors.proto
```diff
@@ -118,3 +118,19 @@ message IdentitySelector {
...
+
+    // The set of Destinations that will be used as ingress gateway services for the external Meshes.
+    // If omitted, the default ingressgateway destination will be used.
```
harveyxia, line 126:
```suggestion
    // If omitted, a mesh-specific default ingress gateway destination will be used.
```

api/common/v1/selectors.proto
```diff
@@ -118,3 +118,19 @@ message IdentitySelector {
...
+    repeated DestinationSelector destination_selectors = 1;
+
+    // The name of the TLS port on the service used as an ingress gateway. Kubernetes services must have a port with this name
+    // in order to be recognized as an ingress gateway. If not specified, will default to `tls`.
```
harveyxia, line 130:
```suggestion
    // Specify by name the TLS port on the ingress gateway destination. If not specified, will default to "tls".
```

api/common/v1/selectors.proto
```diff
@@ -118,3 +118,19 @@ message IdentitySelector {
...
+    string gateway_tls_port_name = 2;
+
+    // The Meshes for which the selected Destinations will be federated. All referenced Meshes must exist in this VirtualMesh.
+    // If omitted, the selected Destinations will be federated to all Meshes in the VirtualMesh.
```
harveyxia, line 134:
```suggestion
    // The Meshes for which the selected Destinations will be configured as ingress gateways. All referenced Meshes must exist in this VirtualMesh.
    // If omitted, the selected Destinations will be configured as ingress gateways for all Meshes in the VirtualMesh.
```

api/discovery/v1/mesh.proto
```diff
@@ -180,6 +180,19 @@ message MeshStatus {
     // The VirtualDestinations, if any, which apply to this mesh.
     repeated AppliedVirtualDestination applied_virtual_destinations = 3;
 
+    // The list of ingress gateway destinations used to access services in this mesh in non-flat-networking mode.
+    // Each individual IngressGatewayDestination should translate to a single ingress gateway on this mesh.
+    repeated IngressGatewayDestination ingress_destinations = 4;
```
harveyxia, line 185:
```suggestion
    // The Destination(s) acting as ingress gateways for east west traffic.
    repeated IngressGateway east_west_ingress_gateways = 4;
```

api/discovery/v1/mesh.proto
```diff
@@ -180,6 +180,19 @@ message MeshStatus {
...
+
+    message IngressGatewayDestination {
+        // The destination on the mesh that acts as the ingress gateway for the mesh.
+        .core.skv2.solo.io.ObjectRef destination_ref = 1;
+
+        // The name of the TLS port on the service used as an ingress gateway. Kubernetes services must have a port with this name
+        // in order to be recognized as an ingress gateway. If not specified, will default to `tls`.
+        string gateway_tls_port_name = 2;
+    }
```
harveyxia, line 194:
```suggestion
    message IngressGateway {
        // The Destination on the mesh that acts as an ingress gateway for the mesh.
        .core.skv2.solo.io.ObjectRef destination_ref = 1;

        // The name of the TLS port on the ingress gateway Destination.
        string tls_port_name = 2;
    }
```

pkg/mesh-networking/translation/istio/mesh/federation/federation_translator.go
```diff
@@ -24,10 +25,13 @@ import (
...
 	defaultGatewayProtocol = "TLS"
-	DefaultGatewayPortName = "tls"
+	defaultGatewayPortName = "tls"
+	defaultGatewayPort     = 15443
+)
 
-	envoySniClusterFilterName        = "envoy.filters.network.sni_cluster"
-	envoyTcpClusterRewriteFilterName = "envoy.filters.network.tcp_cluster_rewrite"
+var (
+	// [Defaults to](https://github.com/istio/istio/blob/ab6cc48134a698d7ad218a83390fe27e8098919f/pkg/config/constants/constants.go#L73) `{"istio": "ingressgateway"}`.
```
harveyxia, line 33: no need to use markdown in Go comments since they don't get rendered

pkg/mesh-networking/apply/configtarget/config_target_validator.go
```diff
@@ -187,6 +189,56 @@ func validateOneVirtualMeshPerMesh(virtualMeshes []*v1.VirtualMesh) {
...
+							continue
+						}
+					}
+					matchedDestinations++
+				}
+				if matchedDestinations == 0 {
+					vMesh.Status.State = commonv1.ApprovalState_INVALID
+					vMesh.Status.Errors = append(
+						vMesh.Status.Errors,
+						fmt.Sprintf("One or more of your ingress gateway destination selectors did not match any destinations"),
```
harveyxia, line 236:
```suggestion
						fmt.Sprintf("No Destinations selected as ingress gateway. At least one must be selected."),
```

pkg/mesh-networking/apply/configtarget/config_target_validator.go
```diff
@@ -187,6 +189,56 @@ func validateOneVirtualMeshPerMesh(virtualMeshes []*v1.VirtualMesh) {
...
+		if len(vMesh.Spec.GetFederation().GetIngressGatewayServiceSelectors()) > 0 {
+			// Validate Ingress gateway service selectors
+			matchedDestinations := 0
+			for _, ingressGatewayServiceSelector := range vMesh.Spec.GetFederation().GetIngressGatewayServiceSelectors() {
+				gatewayTlsPortName := "tls"
+				if ingressGatewayServiceSelector.GetGatewayTlsPortName() != "" {
+					gatewayTlsPortName = ingressGatewayServiceSelector.GetGatewayTlsPortName()
+				}
+				for _, destination := range destinations.List() {
+					if selectorutils.SelectorMatchesDestination(ingressGatewayServiceSelector.GetDestinationSelectors(), destination) {
```
harveyxia, line 202: We also need to validate that Destinations selected by direct reference exist

pkg/mesh-networking/apply/configtarget/config_target_validator.go
```diff
@@ -187,6 +189,56 @@ func validateOneVirtualMeshPerMesh(virtualMeshes []*v1.VirtualMesh) {
 			}
 			invalidVirtualMeshes.Insert(vMesh)
 		}
+
+		if len(vMesh.Spec.GetFederation().GetIngressGatewayServiceSelectors()) > 0 {
```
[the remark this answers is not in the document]
harveyxia: Can we place the ingress gateway selection logic into its own function? It's confusing to colocate it in a function doing something unrelated (validating the one VM per Mesh constraint)

pkg/mesh-networking/translation/istio/mesh/federation/federation_translator.go
```diff
@@ -89,33 +91,89 @@ func (t *translator) Translate(
...
-				Tls: &networkingv1alpha3spec.ServerTLSSettings{
-					Mode: networkingv1alpha3spec.ServerTLSSettings_AUTO_PASSTHROUGH,
-				},
-			}},
-			Selector: ingressGateway.WorkloadLabels,
-		},
+	// one gatweay for each destination, port tuple
+	var selectedIngressGatewayList []*ingressGatewayInfo
+
+	for _, ingressGatewayServiceSelector := range virtualMesh.GetSpec().GetFederation().GetIngressGatewayServiceSelectors() {
```
harveyxia, line 97: discussed offline—we'll want to use the applier convention here, add a [line here](https://github.com/solo-io/gloo-mesh/blob/558c3c8dbe22fa02ad28a9add14b18abd4f6d403/pkg/mesh-networking/apply/applier.go#L127):

```golang
for _, mesh := range input.Meshes().List() {
	mesh.Status.AppliedVirtualMesh = getAppliedVirtualMesh(input.VirtualMeshes().List(), mesh)
	mesh.Status.EastWestIngressGateways = getEastWestIngressGateways(input.VirtualMeshes().List(), mesh, input.Destinations().List())
}
```

that resolves the selected Destinations for ingress gateways on the mesh. Then here, the translator reads directly from the Mesh status for the ingress gateway destinations

pkg/mesh-networking/translation/istio/mesh/federation/federation_translator.go
harveyxia: Since the Gateway CRD will map 1-1 with respect to a (VirtualMesh, Destination) tuple, we can modify the `BuildGatewayName` function to construct a unique name from the VirtualMesh and Destination object meta

pkg/mesh-networking/apply/configtarget/config_target_validator.go
ashleywang1: That should be checked in the `SelectorMatchesDestination` function

harveyxia: to clarify, we should report an error if any direct references don't exist, [we can reuse this function](https://github.com/solo-io/gloo-mesh/blob/8d4fbb67b918dbee810d449b3ee797ca40393d8e/pkg/mesh-networking/apply/configtarget/config_target_validator.go#L102)

pkg/mesh-networking/translation/istio/mesh/federation/federation_translator.go
ashleywang1: This is resulting in names like
```
istio-system   virtual-mesh-gloo-mesh-istio-in-01fe4f88d1b6407050af9b7967b4957   2m49s
istio-system   virtual-mesh-gloo-mesh-istio-in-ea0fbf7b9febcbec1fd78df0b653eb9   2m48s
```

pkg/mesh-networking/apply/applier.go
```diff
@@ -752,6 +755,63 @@ func getAppliedVirtualMesh(
...
+	mesh *discoveryv1.Mesh,
+	destinations discoveryv1.DestinationSlice,
+) []*discoveryv1.MeshStatus_IngressGateway {
+	var selectedIngressGatewayList []*discoveryv1.MeshStatus_IngressGateway
+	for _, virtualMesh := range virtualMeshes {
+		for _, ingressGatewayServiceSelector := range virtualMesh.Spec.GetFederation().GetEastWestIngressGatewaySelectors() {
+			// Check if this selector applies to this mesh
+			var applies bool
+			for _, meshRef := range ingressGatewayServiceSelector.GetMeshes() {
+				if meshRef.GetName() == mesh.GetName() && meshRef.GetNamespace() == mesh.GetNamespace() {
```
harveyxia, line 769: we can use skv2's `ezkube.RefsMatch` for this

pkg/mesh-networking/apply/applier.go
```diff
@@ -752,6 +755,63 @@ func getAppliedVirtualMesh(
...
+					applies = true
+					break
+				}
+			}
+			if !applies {
+				continue
+			}
+			for _, destination := range destinations {
+				// Currently, we can only create gateways out of kube service destination types.
+				if destination.Spec.GetExternalService() != nil {
```
harveyxia, line 779:
```suggestion
				if destination.Spec.GetKubeService() == nil {
```

pkg/mesh-networking/apply/applier.go
```diff
@@ -752,6 +755,63 @@ func getAppliedVirtualMesh(
...
+					continue
```
harveyxia, line 780: we'll want to throw an error here if the user selects a non-k8s destination, and propagate that to the status

pkg/mesh-networking/apply/configtarget/config_target_validator.go
harveyxia: use the constant defined [here](https://github.com/solo-io/gloo-mesh/blob/dcdbfbbb98a8babf29fc2c55d02e4cdf5760d785/pkg/mesh-networking/translation/istio/mesh/federation/federation_translator.go#L27). also include a comment describing that it's the default, hardcoded fallback value

pkg/mesh-networking/apply/configtarget/config_target_validator.go
```diff
@@ -190,6 +193,98 @@ func validateOneVirtualMeshPerMesh(virtualMeshes []*v1.VirtualMesh) {
...
+					continue
+				}
+
+				gatewayTlsPortName := "tls"
+				if ingressGatewayServiceSelector.GetGatewayTlsPortName() != "" {
+					gatewayTlsPortName = ingressGatewayServiceSelector.GetGatewayTlsPortName()
+				}
+				for _, destination := range c.destinations.List() {
+					if selectorutils.SelectorMatchesDestination(ingressGatewayServiceSelector.GetDestinationSelectors(), destination) {
+						if len(destination.Spec.GetKubeService().GetWorkloadSelectorLabels()) != 0 {
```
harveyxia, line 226: code style nit:

you can express required conditions by continuing if they don't hold to avoid nested levels of indentation:

```golang
if !selectorutils.SelectorMatchesDestination(ingressGatewayServiceSelector.GetDestinationSelectors(), destination) {
    continue
}
// logic here
```

pkg/mesh-networking/apply/configtarget/config_target_validator.go
```diff
@@ -190,6 +193,98 @@ func validateOneVirtualMeshPerMesh(virtualMeshes []*v1.VirtualMesh) {
...
+				if matchedDestinations == 0 {
+					vMesh.Status.State = commonv1.ApprovalState_INVALID
+					vMesh.Status.Errors = append(
+						vMesh.Status.Errors,
+						fmt.Sprintf("No Destinations selected as ingress gateway. At least one must be selected."),
+					)
+				}
+			}
+		} else {
+			// Check the mesh specific default exists
```
harveyxia, line 264: We should have the following fallback behavior, in order of precedence:

1. Use the deprecated ingress gateway info on the Mesh object
2. If #1 doesn't exist, use default fallback values

pkg/mesh-networking/apply/configtarget/config_target_validator.go
```diff
@@ -190,6 +193,98 @@ func validateOneVirtualMeshPerMesh(virtualMeshes []*v1.VirtualMesh) {
...
+			defaultExists := false
+			if c.destinations != nil {
+				for _, destination := range c.destinations.List() {
+					if destination.Spec.GetKubeService().GetWorkloadSelectorLabels()["istio"] == "ingressgateway" {
```
harveyxia, line 268: use the [constant defined here](https://github.com/solo-io/gloo-mesh/blob/2fa37161ba5ef92097b3f28a49a32f3ece78849e/pkg/common/defaults/istio.go#L6)

pkg/mesh-networking/apply/configtarget/config_target_validator.go
```diff
@@ -190,6 +193,98 @@ func validateOneVirtualMeshPerMesh(virtualMeshes []*v1.VirtualMesh) {
...
+						for _, ports := range destination.Spec.GetKubeService().GetPorts() {
+							if ports.GetName() == federation.DefaultGatewayPortName && ports.GetPort() != 0 {
+								defaultExists = true
+							}
+						}
+					}
+				}
+			}
+			if !defaultExists {
```
harveyxia, line 277: we can consolidate this logic into the `matchedDestinations` count above, my reasoning being that we don't need a separate error message for not finding the default ingress, we can just use the same error above (`"No Destinations selected as ingress gateway. At least one must be selected."`)

harveyxia (changes requested):
There's some more work to do for the Destination federation translation, [see this TODO](https://github.com/solo-io/gloo-mesh/blob/main/pkg/mesh-networking/translation/istio/destination/federation/federation_translator.go#L183-L185)

api/common/v1/selectors.proto
```diff
@@ -118,3 +118,18 @@ message IdentitySelector {
         repeated .core.skv2.solo.io.ClusterObjectRef service_accounts = 1;
     }
 }
+
+// Select a set of Destinations with tls ports to use as ingress gateway services for the referenced Meshes.
+message IngressGatewaySelector {
+
+    // The set of Destinations that will be used as ingress gateway services for the external Meshes.
+    // If omitted, a mesh-specific default ingress gateway destination will be used.
```
harveyxia, line 126:
```suggestion
    // If omitted, a mesh-specific default ingress gateway destination will be used.
    // For Istio, any Kubernetes Service(s) with the label pair `{"istio": "ingressgateway"}` will be selected.
```

pkg/mesh-networking/apply/applier.go
ashleywang1: We can just check it in the config target validator and put it in the status directly there.

harveyxia:
Can we add a unit test for the new applier logic that applies east west gateways to the Mesh status?

pkg/mesh-networking/translation/istio/mesh/federation/federation_translator.go
harveyxia: discussed offline, we'll use destination name + destination namespace (no need to include VM name/namespace)

pkg/mesh-networking/apply/configtarget/config_target_validator.go
```diff
@@ -190,6 +196,139 @@ func validateOneVirtualMeshPerMesh(virtualMeshes []*v1.VirtualMesh) {
...
+		}
+	}
+}
+
+func (c *configTargetValidator) validateMeshIngressGatewaysForEachVirtualMesh(virtualMeshes []*v1.VirtualMesh) {
+	for _, vMesh := range virtualMeshes {
+		if vMesh.Status.State != commonv1.ApprovalState_ACCEPTED {
+			continue
+		}
+		vMesh := vMesh
```
harveyxia, line 228: I don't think this is needed

pkg/mesh-networking/apply/configtarget/config_target_validator.go
```diff
@@ -190,6 +196,139 @@ func validateOneVirtualMeshPerMesh(virtualMeshes []*v1.VirtualMesh) {
...
+
+		// Check that an ingress gateway exists for each mesh in the virtual mesh
+		eachMeshIsAllSet := make(map[string]bool)
```
harveyxia, line 231: We can make this code more readable by using a string set (import from `"k8s.io/apimachinery/pkg/util/sets"`, and use `sets.NewString`)) named `meshesWithSelectedIngressDestination` keyed on the resource key of the mesh, where existence means the mesh has a selected ingress gateway

pkg/mesh-networking/apply/applier.go
```diff
@@ -752,6 +755,55 @@ func getAppliedVirtualMesh(
...
+) []*discoveryv1.MeshStatus_IngressGateway {
+	var selectedIngressGatewayList []*discoveryv1.MeshStatus_IngressGateway
+	for _, virtualMesh := range virtualMeshes {
+		for _, ingressGatewayServiceSelector := range virtualMesh.Spec.GetFederation().GetEastWestIngressGatewaySelectors() {
+			// Check if this selector applies to this mesh
+			if !selectorutils.IngressGatewaySelectorMatchesMesh(ingressGatewayServiceSelector, mesh) {
+				continue
+			}
+			for _, destination := range destinations {
+				// Currently, we can only create gateways out of kube service destination types.
```
harveyxia, line 771: we need to ignore destinations don't belong to the mesh

pkg/mesh-networking/apply/applier.go
```diff
@@ -752,6 +755,55 @@ func getAppliedVirtualMesh(
...
+						},
+						TlsPortName: ingressGatewayServiceSelector.GetGatewayTlsPortName(),
+					})
+				}
+			}
+		}
+	}
+	if len(selectedIngressGatewayList) == 0 {
+		for _, destination := range destinations {
+			if destination.Spec.GetKubeService().GetWorkloadSelectorLabels()["istio"] == "ingressgateway" {
```
harveyxia, line 789: Can we hoist the label key and value into constants? We can define them in `federation_translator.go`:

```
// Defaults to {"istio": "ingressgateway"} based on https://github.com/istio/istio/blob/ab6cc48134a698d7ad218a83390fe27e8098919f/pkg/config/constants/constants.go#L73
IstioGatewayLabelKey = "istio"
IstioIngressGatewayLabelValue = "ingressgateway"
DefaultGatewayWorkloadLabels = map[string]string{IstioGatewayLabelKey: IstioIngressGatewayLabelValue}
```

and reuse them here

pkg/mesh-networking/apply/configtarget/config_target_validator.go
```diff
@@ -190,6 +196,139 @@ func validateOneVirtualMeshPerMesh(virtualMeshes []*v1.VirtualMesh) {
...
+						vMesh.Status.Errors,
+						fmt.Sprintf("Attempting to select ingress gateway destination %v with no workload labels", sets.Key(destination)),
+					)
+					continue
+				}
+				if err := destinationHasPortNamed(destination, ingressGatewayServiceSelector.GetGatewayTlsPortName()); err != nil {
+					vMesh.Status.State = commonv1.ApprovalState_INVALID
+					vMesh.Status.Errors = append(
+						vMesh.Status.Errors,
+						fmt.Sprintf("Attempting to select ingress gateway destination: %v", err))
```
harveyxia, line 272: this error message needs to be corrected

pkg/mesh-networking/apply/configtarget/config_target_validator.go
```diff
@@ -190,6 +196,139 @@ func validateOneVirtualMeshPerMesh(virtualMeshes []*v1.VirtualMesh) {
...
+			}
+		}
+		// Use the deprecated ingress gateway info on the Mesh object
+		if c.meshes != nil {
+			for _, mesh := range c.meshes.List() {
+				if _, ok := eachMeshIsAllSet[sets.Key(mesh)]; ok {
+					if len(mesh.Spec.GetIstio().GetIngressGateways()) > 0 {
+						eachMeshIsAllSet[sets.Key(mesh)] = true
+						// Add a warning, but don't mark as invalid.
+						vMesh.Status.Errors = append(
```
harveyxia, line 297: I don't think we need to signal this to the user—it's a discovery field so the deprecation is internal to our system

pkg/mesh-networking/apply/configtarget/config_target_validator.go
```diff
@@ -190,6 +196,139 @@ func validateOneVirtualMeshPerMesh(virtualMeshes []*v1.VirtualMesh) {
...
+						if ports.GetName() == defaults.DefaultGatewayPortName && ports.GetPort() != 0 {
+							eachMeshIsAllSet[meshRefKey] = true
+							break
+						}
+					}
+				}
+			}
+		}
+		// Use the deprecated ingress gateway info on the Mesh object
+		if c.meshes != nil {
```
harveyxia, line 291: we don't need this conditional, it's handled by `c.meshes.List()`

pkg/mesh-networking/apply/configtarget/config_target_validator.go
```diff
@@ -190,6 +196,139 @@ func validateOneVirtualMeshPerMesh(virtualMeshes []*v1.VirtualMesh) {
...
+			for _, mesh := range c.meshes.List() {
+				if _, ok := eachMeshIsAllSet[sets.Key(mesh)]; ok {
```
harveyxia, line 293: this should be `!ok`

pkg/mesh-networking/translation/istio/mesh/federation/federation_translator.go
```diff
@@ -24,10 +27,12 @@ import (
...
 	defaultGatewayProtocol = "TLS"
-	DefaultGatewayPortName = "tls"
+	defaultGatewayPort     = 15443
+)
 
-	envoySniClusterFilterName        = "envoy.filters.network.sni_cluster"
-	envoyTcpClusterRewriteFilterName = "envoy.filters.network.tcp_cluster_rewrite"
+var (
+	// Defaults to {"istio": "ingressgateway"} based on https://github.com/istio/istio/blob/ab6cc48134a698d7ad218a83390fe27e8098919f/pkg/config/constants/constants.go#L73
+	defaultGatewayWorkloadLabels = map[string]string{"istio": "ingressgateway"}
```
harveyxia, line 35: this already exists [here](https://github.com/solo-io/gloo-mesh/blob/2fa37161ba5ef92097b3f28a49a32f3ece78849e/pkg/common/defaults/istio.go#L6)

pkg/mesh-networking/translation/istio/mesh/federation/federation_translator.go
```diff
@@ -73,53 +83,91 @@ func (t *translator) Translate(
...
 	istioCluster := istioMesh.Installation.Cluster
 	istioNamespace := istioMesh.Installation.Namespace
 	federatedHostnameSuffix := hostutils.GetFederatedHostnameSuffix(virtualMesh.Spec)
 
-	// istio gateway names must be DNS-1123 labels
-	// hyphens are legal, dots are not, so we convert here
-	gwName := BuildGatewayName(virtualMesh)
+	// create the east west ingress gateways
+	if len(mesh.Status.GetEastWestIngressGateways()) == 0 {
+		// fall back to old behavior
```
harveyxia, line 93: we should encapsulate the fallback behavior in the applier—the applier should select ingress gateway services based on the fallback behavior and apply them to the mesh status. Then this translator can just fetch the ingress gateway data from the mesh status without knowing anything about fallback behavior

pkg/mesh-networking/apply/applier.go
```diff
@@ -752,6 +755,55 @@ func getAppliedVirtualMesh(
...
+							Name:      destination.GetName(),
+							Namespace: destination.GetNamespace(),
+						},
+						TlsPortName: ingressGatewayServiceSelector.GetGatewayTlsPortName(),
+					})
+				}
+			}
+		}
+	}
+	if len(selectedIngressGatewayList) == 0 {
```
harveyxia, line 787: discussed offline for fallback behavior:

// 1. look at mesh ingress gateway info (if it's populated) for the ingress Destination
// 2. look for a Destination in the mesh with the default hardcoded workload label pair + tls port name

pkg/mesh-networking/apply/configtarget/config_target_validator.go
```diff
@@ -63,6 +67,8 @@ func (c *configTargetValidator) ValidateVirtualMeshes(virtualMeshes v1.VirtualMe
 	}
 
 	validateOneVirtualMeshPerMesh(virtualMeshes)
+	c.validateVirtualMeshIngressGatewaySelectors(virtualMeshes)
+	c.validateMeshIngressGatewaysForEachVirtualMesh(virtualMeshes)
```
harveyxia, line 71: discussed offline, we can move all of this logic into the applier

pkg/mesh-networking/apply/configtarget/config_target_validator.go
ashleywang1: It'll look like this: ```Attempting to select ingress gateway destination: destination x has no port named x"

pkg/mesh-networking/apply/configtarget/config_target_validator.go
ashleywang1: Some tests ended up hitting this case because they set c.meshes to nil

harveyxia: if `c.meshes` is nil, `c.meshes.List()` will return nil, [see here](https://github.com/solo-io/gloo-mesh/blob/073a2202604aaf1a981f4ca94cfb53810e916bb2/pkg/api/discovery.mesh.gloo.solo.io/v1/sets/sets.go#L501-L503)

EItanya (changes requested):
Need to have further discussion about move to circle ci

harveyxia:
> Need to have further discussion about move to circle ci

Discussed offline, we'll only deploy the second ingress gateway in closed source. The open source e2e test will only test the ability to explicitly select the default istio ingress gateway (and not the ability to select and load balance across 2 ingress gateways).

ci/setup-funcs.sh
```diff
@@ -13,19 +13,21 @@ function create_kind_cluster() {
   # https://github.com/solo-io/gloo-mesh/issues/700
   kindImage=kindest/node:v1.17.5
 
+  # gloo mesh cluster name
   cluster=$1
-  port=$2
+
+  # ingress ports
+  port1=$2
```
EItanya, line 20: Are we exposing these args through the e2e golang code?

harveyxia: they can be set through environment variables, see the changes in `ci/setup-kind.sh`

ci/setup-funcs.sh
```diff
@@ -460,7 +461,8 @@ EOF
 # Operator spec for istio 1.9.x
 function install_istio_1_9() {
   cluster=$1
-  port=$2
+  ingressPort1=$2
+  ingressPort2=$3
```
EItanya, line 465: I don't think we need this second port as an arg to the istio functions anymore since we are only creating a single Gateway

harveyxia: yeah i cleaned up the scripts, see commit `1e39e30`

ci/setup-funcs.sh
```diff
@@ -313,7 +315,7 @@ EOF
 # Operator spec for istio 1.7.x
 function install_istio_1_7() {
   cluster=$1
-  port=$2
+  ingressPort1=$2
```
EItanya, line 318: Can we rename these to `ingressPort` since there is only 1 of them

pkg/common/defaults/istio.go
```diff
@@ -4,10 +4,13 @@ package defaults
 // their ingress gateway is hardcoded in their own implementation to have this label
 // https://github.com/istio/istio/blob/4e27ddc64f6a12e622c4cd5c836f5d7edf94e971/istioctl/cmd/describe.go#L1138
```
EItanya, line 5: No longer relevant

test/e2e/istio/pkg/tests/federation.go
```diff
@@ -343,4 +343,32 @@ func FederationTest() {
...
+
+		By("able to select the default ingress gateway destination explicitly", func() {
+			defaultFederatedIngressGatewayVm := VirtualMesh.DeepCopy()
+			defaultFederatedIngressGatewayVm.Spec.Federation.EastWestIngressGatewaySelectors = []*commonv1.IngressGatewaySelector{
+				{
+					DestinationSelectors: []*commonv1.DestinationSelector{
+						{
+							KubeServiceMatcher: &commonv1.DestinationSelector_KubeServiceMatcher{
+								Labels: map[string]string{
+									"istio": "ingressgateway",
```
EItanya, line 357: `traffic: east-west`?

pkg/mesh-networking/translation/istio/mesh/federation/federation_translator.go
```diff
@@ -4,10 +4,14 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/rotisserie/eris"
+	"github.com/solo-io/skv2/pkg/ezkube"
+
```
EItanya, line 9: Nit: unecessary newline

test/e2e/istio/pkg/tests/federation.go
harveyxia: the other ingress gateway has label `istio: ingressgateway2` so this label pair alone should be sufficient to distinctly select the first ingress gateway

pkg/mesh-networking/apply/applier.go
```diff
@@ -755,6 +761,234 @@ func getAppliedVirtualMesh(
...
+		// only consider kubernetes services
+		if kubeService == nil {
+			continue
+		}
+
+		if kubeService.GetWorkloadSelectorLabels()[defaults.IstioGatewayLabelKey] != defaults.IstioIngressGatewayLabelValue {
+			continue
+		}
+
+		if appliedIngressGateway, err := buildAppliedIngressGateway(destination, defaults.IstioGatewayTlsPortName); err != nil {
```
EItanya, line 900: Are we planning on allowing users to change this port name? This [issue](https://github.com/solo-io/gloo-mesh/issues/1705) is not quite the same, but just wondering if that's something we care about

pkg/mesh-networking/apply/applier.go
```diff
@@ -755,6 +761,234 @@ func getAppliedVirtualMesh(
...
+		return nil
+	}
+	virtualMesh, err := virtualMeshes.Find(mesh.Status.GetAppliedVirtualMesh().GetRef())
+	if err != nil {
+		// should never happen
+		return nil
+	}
+
+	var selectedIngressGatewayList []*discoveryv1.MeshStatus_AppliedIngressGateway
+	// Resolve all of the VM’s ingress gateway selectors to a list of (Destination, tls port) tuples that belong to the mesh
```
EItanya, line 780: Super Nit, I was very confused why we were talking about Virtual Machines 😆 
```suggestion
	// Resolve all of the VirtualMesh’s ingress gateway selectors to a list of (Destination, tls port) tuples that belong to the mesh
```

pkg/mesh-networking/apply/applier.go
harveyxia: This is the default gateway tls port name that we fall back on in the case that 1) we didn't find any ingress destinations from the user provided ingress selector and 2) we didn't find any ingress gateways in the deprecated Mesh ingress gateways field

Because we allow users to [configure the tls port name ](https://github.com/solo-io/gloo-mesh/pull/1691/files#diff-01a5bfa7395892a4c2a33dfcf287029adc8ab84eb3df83d2bceec1feaec06983R131) through the ingress gateway selector, I feel comfortable making the fallback default non-configurable. What do you think?

EItanya: Nvm, I see that this is the default function, sorry

api/common/v1/selectors.proto
```diff
@@ -117,4 +117,20 @@ message IdentitySelector {
...
+    // For Istio, any Kubernetes Service(s) with the label pair `{"istio": "ingressgateway"}` will be selected.
+    repeated DestinationSelector destination_selectors = 1;
+
+    // Specify by name the TLS port on the ingress gateway destination. If not specified, will default to "tls".
+    string gateway_tls_port_name = 2;
+
+    // The Meshes for which the selected Destinations will be configured as ingress gateways. All referenced Meshes must exist in this VirtualMesh.
+    // If omitted, the selected Destinations will be configured as ingress gateways for all Meshes in the VirtualMesh.
+    repeated .core.skv2.solo.io.ObjectRef meshes = 3;
```
EItanya, line 135: Is there any way we can make this a label set rather than a list. A list can become invalid very quickly, and is potentially very verbose. I suppose we have the same issue for the VirtualMesh Mesh selection now as well.

api/common/v1/selectors.proto
```diff
@@ -117,4 +117,20 @@ message IdentitySelector {
...
+// Select a set of Destinations with tls ports to use as ingress gateway services for the referenced Meshes.
+message IngressGatewaySelector {
+
+    // The set of Destinations that will be used as ingress gateway services for the external Meshes.
+    // If omitted, a mesh-specific default ingress gateway destination will be used.
+    // For Istio, any Kubernetes Service(s) with the label pair `{"istio": "ingressgateway"}` will be selected.
+    repeated DestinationSelector destination_selectors = 1;
+
+    // Specify by name the TLS port on the ingress gateway destination. If not specified, will default to "tls".
+    string gateway_tls_port_name = 2;
```
EItanya, line 131: Can we also specify by number in a `oneof`?

api/discovery/v1/mesh.proto
```diff
@@ -186,6 +186,23 @@ message MeshStatus {
...
     repeated AppliedVirtualDestination applied_virtual_destinations = 3;
 
+    // The Destination(s) acting as ingress gateways for east west traffic.
+    repeated AppliedIngressGateway applied_east_west_ingress_gateways = 4;
+
+    message AppliedIngressGateway {
+        // The Destination on the mesh that acts as an ingress gateway for the mesh.
+        .core.skv2.solo.io.ObjectRef destination_ref = 1;
+
+        // The externally accessible IP addresses.
```
EItanya, line 196: Only IPs or can these also be hostnames, and does that make a difference?

api/discovery/v1/mesh.proto
```diff
@@ -186,6 +186,23 @@ message MeshStatus {
...
+        repeated string external_addresses = 2;
+
+        // The port on the ingress gateway Destination receiving TLS traffic.
+        uint32 tls_destination_port = 3;
+
+        // The port on the ingress gateway workload receiving TLS traffic.
+        uint32 tls_container_port = 4;
```
EItanya, line 203: Wdyt about removing `tls` from these since the user can use a different name other than `tls` for the port name, and this feels like a bit of a relic. Or is this more to do with the type of traffic? Still feels a bit prescriptive.

pkg/mesh-networking/apply/applier.go
```diff
@@ -755,6 +761,234 @@ func getAppliedVirtualMesh(
...
+		for _, ingressGatewayServiceSelector := range virtualMesh.Spec.GetFederation().GetEastWestIngressGatewaySelectors() {
+			// Check if this selector applies to this mesh
+			if !ingressGatewaySelectorMatchesMesh(ingressGatewayServiceSelector, mesh) {
+				continue
+			}
+			if !selectorutils.SelectorMatchesDestination(ingressGatewayServiceSelector.GetDestinationSelectors(), destination) {
+				continue
+			}
+
+			if len(kubeService.GetWorkloadSelectorLabels()) == 0 {
```
EItanya, line 801: Just wondering why this is automatically an error? What does Kubernetes do if there is a service with no workload selector labels?

pkg/mesh-networking/apply/applier.go
```diff
@@ -755,6 +761,234 @@ func getAppliedVirtualMesh(
...
+				virtualMesh.Status.State = commonv1.ApprovalState_INVALID
+				virtualMesh.Status.Errors = append(
+					virtualMesh.Status.Errors,
+					fmt.Sprintf("attempting to select ingress gateway destination %v with no workload labels", sets.Key(destination)),
```
EItanya, line 805:
```suggestion
					fmt.Sprintf("attempting to select ingress gateway destination %v with no workload selector labels", sets.Key(destination)),
```

pkg/mesh-networking/apply/applier.go
harveyxia: It's possible for k8s services to have no selectors, see [here](https://kubernetes.io/docs/concepts/services-networking/service/#services-without-selectors). So this is a guard rail in case the user accidentally selects an externally addressable service that has no selector

api/discovery/v1/mesh.proto
harveyxia: Interesting point—we currently prescribe TLS-secured traffic for cross cluster communication because we set tls `AUTO_PASSTHROUGH` on our Gateway. If we want to relax that constraint we'll have to make some changes to the way we configure the Gateway, and possibly some API changes as well.

api/discovery/v1/mesh.proto
harveyxia: good call, these can be either IPs or hostnames. It affects the federated ServiceEntry resolution mode which handles the different scenarios (e.g. only IPs, only hostnames, mix of ips + hostnames, etc)

I'll clarify the docstring

pkg/mesh-networking/apply/applier.go
EItanya: Gotcha, I think this is a ok guard rail for now, but this could be a legitimate use case I suppose

api/discovery/v1/mesh.proto
EItanya: Just to clarify, the traffic does need to be TLS. But the name I think had to do with the port name, not the traffic type. Might be overthinking it though

test/e2e/istio/pkg/tests/federation.go
EItanya: I just recommended this because it's not the default, so it will ensure we're not falling back to the default for some unknown reason.

api/common/v1/selectors.proto
harveyxia: very good question—I couldn't come up with a better way to represent these semantics in a structured way.

We could couple any per-mesh config with the [`meshes` field](https://github.com/solo-io/gloo-mesh/blob/f43f69c716c5be105dc229030e592dd2f0e6bc2e/api/networking/v1/virtual_mesh.proto#L21), but that also complicates the UX. Let's discuss offline

api/discovery/v1/mesh.proto
harveyxia: discussed offline—we'll remove the `tls` suffix since the port can be any port and k8s itself has no notion of tls secured ports

api/common/v1/selectors.proto
EItanya: On a separate note: How does this field interact with the federation selectors? Do we need both layers?

harveyxia: They are separate—federation selectors affect what destinations get federated, and ingress gateway selectors affect which externally accessible k8s services are used for remote traffic (i.e. the address used to populate SE's for federated destinations on remote clusters)

harveyxia: They are separate—federation selectors affect what destinations get federated, and ingress gateway selectors affect which externally accessible k8s services are used for remote traffic (i.e. the address used to populate SE's for federated destinations on remote clusters)

EItanya: Discussed offline. The core of my argument here is that this field is potentially more complicated than the value it offers. Because of how API deprecation works we can never remove fields only add, so my recommendation would be to leave this field out for the first version, and add it in later if users need it.

The main source of the complication being the list of meshes. Mesh reference lists are cumbersome for the user, and we should revisit this when we have label selection for meshes.

EItanya (approved):

harveyxia (approved):

Issue #1639, opened 2021-06-11
Traffic policy request timeout using wrong port


nmnellis:
**Describe the bug**
trying to set a request timeout on all backend services results in a 503 error. It appears that its trying to rewrite my port from 8090 to 9000

```
curl "backend.app-1-6787.svc.cluster.local:8090"
upstream connect error or disconnect/reset before headers. reset reason: connection termination

[2021-06-11T00:41:33.441Z] "GET / HTTP/1.1" 503 UC upstream_reset_before_response_started{connection_termination} - "-" 0 95 3 - "-" "curl/7.58.0" "9b73951a-b501-464f-a61c-66ff4ba35f9b" "backend.app-1-6787.svc.cluster.local:8090" "10.42.0.50:9000" outbound|9000||backend.app-1-6787.svc.cluster.local 10.42.0.49:59810 10.43.243.25:8090 10.42.0.49:34420 - -
```

**To Reproduce**
```yaml
apiVersion: networking.mesh.gloo.solo.io/v1
kind: VirtualMesh
metadata:
  name: virtual-mesh
  namespace: gloo-mesh
spec:
  federation:
    selectors:
      - { }
  meshes:
    - name: istiod-istio-system-cluster-0
      namespace: gloo-mesh
    - name: istiod-istio-system-cluster-1
      namespace: gloo-mesh
---
apiVersion: networking.mesh.gloo.solo.io/v1
kind: TrafficPolicy
metadata:
  namespace: gloo-mesh
  name: backend-timeout
spec:
  destinationSelector:
    - kubeServiceMatcher:
        labels:
          app: backend
  policy:
    requestTimeout: 1s
```
**Expected behavior**

**Additional context**

```
▶ kg vs -n app-1-6787 backend -o yaml
+ kubectl get vs -n app-1-6787 backend -o yaml
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  annotations:
    federation.networking.mesh.gloo.solo.io: backend.app-1-6787.cluster-0
    parents.networking.mesh.gloo.solo.io: '{"discovery.mesh.gloo.solo.io/v1, Kind=Destination":[{"name":"backend-app-1-6787-cluster-0","namespace":"gloo-mesh"}],"networking.mesh.gloo.solo.io/v1,
      Kind=VirtualMesh":[{"name":"virtual-mesh","namespace":"gloo-mesh"}]}'
  creationTimestamp: "2021-06-11T00:40:04Z"
[...]

PR #1692, opened 2021-06-28
Fix generated istio virtual service in case of delegation to k8s service with several ports

kdorosh:
# Change

Fix a potential bug in the generated istio virtual service when using a traffic policy that selects a kubernetes service that itself has several ports. Before this fix, the generated virtual service did not have an istio `HttpMatchRequest` for each port, so HTTP requests would always match the first route on the virtual service (which might have the wrong port) rather than the correct route.

For example:
```yaml
apiVersion: networking.mesh.gloo.solo.io/v1
kind: VirtualMesh
metadata:
  name: virtual-mesh
  namespace: gloo-mesh
spec:
  federation:
    selectors:
      - { }
  meshes:
    - name: istiod-istio-system-cluster-0
      namespace: gloo-mesh
    - name: istiod-istio-system-cluster-1
      namespace: gloo-mesh
---
apiVersion: networking.mesh.gloo.solo.io/v1
kind: TrafficPolicy
metadata:
  namespace: gloo-mesh
  name: backend-timeout
spec:
  destinationSelector:
    - kubeServiceMatcher:
        labels:
          app: backend
  policy:
    requestTimeout: 1s
```

generates

```yaml
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  labels:
    cluster.multicluster.solo.io: ""
    owner.networking.mesh.gloo.solo.io: gloo-mesh
    relay-agent: cluster-1
  name: backend
  namespace: app-1-6787
spec:
  hosts:
  - backend.app-1-6787.svc.cluster.local
  http:
  - route:
    - destination:
        host: backend.app-1-6787.svc.cluster.local
        port:
          number: 9000
    timeout: 1s
  - route:
    - destination:
        host: backend.app-1-6787.svc.cluster.local
        port:
          number: 9443
    timeout: 1s
  - route:
    - destination:
        host: backend.app-1-6787.svc.cluster.local
        port:
          number: 8090
    timeout: 1s
  - route:
    - destination:
        host: backend.app-1-6787.svc.cluster.local
        port:
          number: 7070
    timeout: 1s
```

rather than

```yaml
spec:
  hosts:
  - backend.app-1-6787.svc.cluster.local
  http:
  - route:
    - destination:
        host: backend.app-1-6787.svc.cluster.local
        port:
          number: 9000
    match:
      - port: 9000
    timeout: 1s
  - route:
    - destination:
        host: backend.app-1-6787.svc.cluster.local
        port:
          number: 9443
    match:
      - port: 9443
    timeout: 1s
...
```

# Context

The crux of the bug here was that [this code](https://github.com/solo-io/gloo-mesh/blob/da73eaec6419f106dd24170291185ea2059298a7/pkg/mesh-networking/translation/istio/destination/virtualservice/virtual_service_translator.go#L363-L367) was not being run in the case of `baseRoute` not being initialized with any matchers after the decorators had run, thus the port would never be copied over.
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1639


pkg/mesh-networking/translation/istio/destination/virtualservice/virtual_service_translator.go
```diff
@@ -333,6 +334,22 @@ func (t *translator) setDefaultDestination(
 			Host: destinationFQDN,
 		},
 	}}
+
+	// need a default matcher for the destination, which gets populated later
+	// in duplicateRouteForEachPort()
+	if len(baseRoute.Match) == 0 {
```
harveyxia, line 340: we'll need to make sure this logic executes even when `baseRoute.Route != nil`, we can either change L329 to not return early and rename this function to `setDefaultDestinationAndPortMatchers` or hoist this new logic into its own function

pkg/mesh-networking/translation/istio/destination/virtualservice/virtual_service_translator.go
```diff
@@ -333,6 +334,22 @@ func (t *translator) setDefaultDestination(
...
+		var defaultMatchers []*networkingv1alpha3spec.HTTPMatchRequest
+		for _, p := range ports {
+			port := p.GetPort()
+			if port == 0 {
+				port = p.GetNodePort()
```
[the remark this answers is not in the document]
kdorosh: it was a todo to myself (since this was a draft pr) to check whether or not we wanted to fallback to the exposed nodeport if the kube service is using that instead of the cluster port. seemed like it might be sane behavior if users wanted flat network behavior cross cluster: https://www.bmc.com/blogs/kubernetes-port-targetport-nodeport/

looks like we don't use the nodeport today though https://github.com/solo-io/gloo-mesh/blob/1e12809f95b84ad63f35ac953607230f488e8ebc/pkg/mesh-networking/translation/istio/destination/virtualservice/virtual_service_translator.go#L358

and to my knowledge we have flat networking in GM, still have to read / see how we did that to resolve this todo

pkg/mesh-networking/translation/istio/destination/virtualservice/virtual_service_translator.go
kdorosh: addressed in https://github.com/solo-io/gloo-mesh/pull/1692/commits/46409f20bbfdab6d052cdf5072cdaa0598b8d20d and

pkg/mesh-networking/translation/istio/destination/virtualservice/virtual_service_translator.go
```diff
@@ -344,10 +350,6 @@ func duplicateRouteForEachPort(
 	baseRoute *networkingv1alpha3spec.HTTPRoute,
 	ports []*discoveryv1.DestinationSpec_KubeService_KubeServicePort,
 ) []*networkingv1alpha3spec.HTTPRoute {
-	if len(ports) == 1 {
-		// no need to specify port for single-port service
-		return []*networkingv1alpha3spec.HTTPRoute{baseRoute}
-	}
```
kdorosh, line 350: mostly looking for feedback on this change. opted to be explicit to prevent future kinds of overlap / short-circuit matching, at the expense of verbosity in the generated istio config

EItanya: Decided in parking lot that it's worth being more verbose here

harveyxia (approved):

PR #1697, opened 2021-06-30
Deterministic helm codegen fix

kdorosh:
picking up https://github.com/solo-io/k8s-utils/pull/13


kdorosh:
codegen uses skv2 which already has this fix

PR #1700, opened 2021-07-01
Fix generated istio virtual service in case of delegation to k8s service with several ports

kdorosh:
backport of https://github.com/solo-io/gloo-mesh/pull/1692

# Change

Fix a potential bug in the generated istio virtual service when using a traffic policy that selects a kubernetes service that itself has several ports. Before this fix, the generated virtual service did not have an istio `HttpMatchRequest` for each port, so HTTP requests would always match the first route on the virtual service (which might have the wrong port) rather than the correct route.

For example:
```yaml
apiVersion: networking.mesh.gloo.solo.io/v1
kind: VirtualMesh
metadata:
  name: virtual-mesh
  namespace: gloo-mesh
spec:
  federation:
    selectors:
      - { }
  meshes:
    - name: istiod-istio-system-cluster-0
      namespace: gloo-mesh
    - name: istiod-istio-system-cluster-1
      namespace: gloo-mesh
---
apiVersion: networking.mesh.gloo.solo.io/v1
kind: TrafficPolicy
metadata:
  namespace: gloo-mesh
  name: backend-timeout
spec:
  destinationSelector:
    - kubeServiceMatcher:
        labels:
          app: backend
  policy:
    requestTimeout: 1s
```

generates

```yaml
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  labels:
    cluster.multicluster.solo.io: ""
    owner.networking.mesh.gloo.solo.io: gloo-mesh
    relay-agent: cluster-1
  name: backend
  namespace: app-1-6787
spec:
  hosts:
  - backend.app-1-6787.svc.cluster.local
  http:
  - route:
    - destination:
        host: backend.app-1-6787.svc.cluster.local
        port:
          number: 9000
    timeout: 1s
  - route:
    - destination:
        host: backend.app-1-6787.svc.cluster.local
        port:
          number: 9443
    timeout: 1s
  - route:
    - destination:
        host: backend.app-1-6787.svc.cluster.local
        port:
          number: 8090
    timeout: 1s
  - route:
    - destination:
        host: backend.app-1-6787.svc.cluster.local
        port:
          number: 7070
    timeout: 1s
```

rather than

```yaml
spec:
  hosts:
  - backend.app-1-6787.svc.cluster.local
  http:
  - route:
    - destination:
        host: backend.app-1-6787.svc.cluster.local
        port:
          number: 9000
    match:
      - port: 9000
    timeout: 1s
  - route:
    - destination:
        host: backend.app-1-6787.svc.cluster.local
        port:
          number: 9443
    match:
      - port: 9443
    timeout: 1s
...
```

# Context

The crux of the bug here was that [this code](https://github.com/solo-io/gloo-mesh/blob/da73eaec6419f106dd24170291185ea2059298a7/pkg/mesh-networking/translation/istio/destination/virtualservice/virtual_service_translator.go#L363-L367) was not being run in the case of `baseRoute` not being initialized with any matchers after the decorators had run, thus the port would never be copied over.
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1639


kdorosh:
bug fix would be needed for test to pass, removed the e2e conflict detection test: https://github.com/solo-io/gloo-mesh/pull/1543

harveyxia (approved):

PR #1703, opened 2021-07-02
TrafficPolicy gateway selectors API

Sodman:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/579


api/enterprise/networking/v1beta1/route.proto
```diff
@@ -55,6 +55,10 @@ message Route {
     // RouteOption behavior will be inherited by delegated routes which do not specify their own `options`
     .networking.mesh.gloo.solo.io.TrafficPolicySpec.Policy options = 7;
 
+    // Route level labels. Can be used by other resources (eg a TrafficPolicy) to
+    // select specific routes within a given gateway object.
```
harveyxia, line 59:
```suggestion
    // Specify labels for this route, which are used by other resources (e.g. TrafficPolicy) to
    // select specific routes within a given gateway object.
```

api/enterprise/networking/v1beta1/route_table.proto
```diff
@@ -40,6 +43,10 @@ message RouteTableStatus {
 
     // List of child RouteTables that this RouteTable delegates to
     repeated .core.skv2.solo.io.ObjectRef selected_route_tables = 5;
+
+    // The set of TrafficPolicies that have been applied to this Destination.
+    // {{/* Note: validation of this field disabled because it slows down cue tremendously*/}}
```
harveyxia, line 48: This gets rendered in our docs [like so](https://gloo-mesh--pr1703-tp-gateway-selectors-6uyfkkdd.web.app/gloo-mesh/latest/reference/api/github.com.solo-io.gloo-mesh.api.enterprise.networking.v1beta1.route_table/#routetablestatus). I'm not sure if there's a way to hide docstrings from the rendered docs.

api/networking/v1/traffic_policy.proto
```diff
@@ -238,6 +253,32 @@ message TrafficPolicySpec {
...
+
+    // Specify selected gateway traffic by specifying which gateway
+    // resources (virtualHosts or routeTables) to select. You can optionally further
+    // filter by using route labels to only select a subset of routes within those resources.
+    // If no virtualHost or routeTable selectors (or references) are explicitly
+    // set, all virtualHosts and routeTables will be selected by default.
+    // If no route label matcher is specified, all routes on the
+    // selected resources are matched.
+    message RouteSelector {
+        // Select Virtual Hosts by Reference
```
harveyxia, line 265:
```suggestion
        // Select VirtualHosts by reference.
```

api/networking/v1/traffic_policy.proto
```diff
@@ -238,6 +253,32 @@ message TrafficPolicySpec {
...
+        repeated .core.skv2.solo.io.ObjectRef virtual_host_refs = 1;
+
+        // Select Virtual Hosts by label/namespace
```
harveyxia, line 268:
```suggestion
        // Select VirtualHosts by label and/or namespace.
```

api/networking/v1/traffic_policy.proto
```diff
@@ -238,6 +253,32 @@ message TrafficPolicySpec {
...
+        .core.skv2.solo.io.ObjectSelector virtual_host_selector = 2;
+
+        // Select Route Tables by Reference
```
harveyxia, line 271:
```suggestion
        // Select RouteTables by reference.
```

api/networking/v1/traffic_policy.proto
```diff
@@ -238,6 +253,32 @@ message TrafficPolicySpec {
...
+        repeated .core.skv2.solo.io.ObjectRef route_table_refs = 3;
+
+        // Select Route Tables by label/namespace
```
harveyxia, line 274:
```suggestion
        // Select RouteTables by label and/or namespace.
```

api/networking/v1/traffic_policy.proto
```diff
@@ -238,6 +253,32 @@ message TrafficPolicySpec {
...
+        .core.skv2.solo.io.ObjectSelector route_table_selector = 4;
+
+        // Map of route labels to match. All labels listed here must be
```
harveyxia, line 277:
```suggestion
        // Specify a set of labels for selecting Routes. All labels listed here must be
```

harveyxia (commented):
Have we implemented DestinationPolicy yet? Or is that in a follow up?

api/enterprise/networking/v1beta1/route_table.proto
npolshakova: Hm, we could have the note as a comment like we do in the route proto? https://github.com/solo-io/gloo-mesh/blob/16daf13b820786bd807974bee691b8ae856c4dec/api/enterprise/networking/v1beta1/route.proto#L68 I don't know if that's any better though for public facing docs though. Is there no way of hiding comments from rendered docs?

api/networking/v1/applied_policies.proto
```diff
@@ -0,0 +1,27 @@
...
+
+	// The observed generation of the accepted TrafficPolicy.
+	int64 observedGeneration = 2;
+
+	// The spec of the last known valid TrafficPolicy.
+	TrafficPolicySpec spec = 3;
+
+	// The list of routes to which the TrafficPolicy applies, as selected by their labels.
+	// Value is "*" if the TrafficPolicy applies to all routes.
+	repeated string routes = 4;
```
EItanya, line 26: How are these string ids going to be constructed for the routes. Just so a user/us can know how properly read this list

npolshakova: We can change the docs to be more clear here. The list of routes will be indicated by the route name here.

api/enterprise/networking/v1beta1/route_table.proto
harveyxia: unfortunately, we don't have a way of doing this right now

api/networking/v1/applied_policies.proto
ilackarms: since route name is optional we can either document that the route name will be omitted, or we can substitute the path matcher

api/networking/v1/traffic_policy.proto
```diff
@@ -260,5 +297,16 @@ message TrafficPolicyStatus {
 
     // Any errors found while processing this generation of the resource.
     repeated string errors = 5;
+
+    // The Gateway resoures to which this traffic policy has been applied.
+    // The resource names are in the format `name.namespace (resourceType)`
```
ilackarms, line 302: nit: note that the keys in the map are the resource names referred to here

ilackarms (commented):
a few comments, nothing earth shattering

npolshakova (approved):
Looks good!

PR #1710, opened 2021-07-08
Gloo Mesh Gateway external auth API

kdorosh:
## Change

Adds first iteration of extauth API for Gloo Mesh Gateway.

## Context

There will be follow ups to add more features in an attempt to bring closer to feature parity with Gloo Edge


kdorosh (commented):
note that this pr does not have the extauth equivalent for https://github.com/solo-io/gloo-mesh/blob/16daf13b820786bd807974bee691b8ae856c4dec/api/enterprise/networking/v1beta1/rate_limit_server_config.proto#L24

i.e., this first pass MVP does not propagate `AuthConfig`s across clusters, so even though each cluster has its own extauth service, only the local cluster has the server config

api/networking/v1/extauth/extauth.proto
```diff
@@ -2,11 +2,154 @@ syntax = "proto3";
...
-  // TODO: extauth
-  string todo = 1;
-}
\ No newline at end of file
+  // TODO(kdorosh): this is hard-coded. make configurable, allow health checks / mTLS, use EnvoyFilter?
+  // The upstream to ask about auth decisions
+  //.core.skv2.solo.io.ObjectRef extauthz_server_ref = 1;
+
+  // If this is set, communication to the upstream will be via HTTP and not GRPC.
+  HttpService http_service = 2;
```
EItanya, line 18: Does our ext-auth-service support this today, or would this be for a customer ext-auth-server?

api/networking/v1/extauth/extauth.proto
```diff
@@ -2,11 +2,154 @@ syntax = "proto3";
...
+  // Describes the transport protocol version to use when connecting to the ext auth server.
+  enum ApiVersion {
+    // Use v3 API.
+    V3 = 0;
+  }
+
+  // Determines the API version for the `ext_authz` transport protocol that will be used by Envoy
+  // to communicate with the auth server. Defaults to `V2`. For more info, see the `transport_api_version` field
+  // [here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto#extensions-filters-http-ext-authz-v3-extauthz).
+  ApiVersion transport_api_version = 9;
```
EItanya, line 66: I'm not sure we should expose this to the user unless they are brining their own server. Otherwise we should determine it for them.

api/networking/v1/extauth/extauth.proto
```diff
@@ -2,11 +2,154 @@ syntax = "proto3";
...
 
 // Configure the Extauth Filter on a Gateway
 message GatewayExtauth {
-  // TODO: extauth
-  string todo = 1;
-}
\ No newline at end of file
+  // TODO(kdorosh): this is hard-coded. make configurable, allow health checks / mTLS, use EnvoyFilter?
+  // The upstream to ask about auth decisions
+  //.core.skv2.solo.io.ObjectRef extauthz_server_ref = 1;
```
EItanya, line 15: @yuval-k, @ilackarms, @harveyxia this is a very interesting one. I think we should setup a meeting to talk about how exactly we want to connect to rate-limiting/extauth. Potentially we could piggy back on `mTLS` connections in istio, but that would involve "injecting" our services.

api/networking/v1/extauth/extauth.proto
```diff
@@ -2,11 +2,154 @@ syntax = "proto3";
...
+  bool pack_as_bytes = 3;
+}
+
+// Extauth configuration for a Route or TrafficPolicy. Configures extauth for individual HTTP routes
+message RouteExtauth {
+  oneof spec {
+    //  Set to true to disable auth on the route.
+    bool disable = 1;
+    // A reference to an AuthConfig. This is used to configure the Gloo Mesh Gateway extauth server.
+    .core.skv2.solo.io.ObjectRef config_ref = 2;
```
EItanya, line 130: @ilackarms likes to allow users to provide either a reference or to inline the config. Do you think that would be useful here as well.

api/networking/v1/extauth/extauth.proto
kdorosh: this would be for a customer ext-auth-server, akin to https://docs.solo.io/gloo-edge/latest/guides/security/auth/custom_auth/#creating-a-simple-http-authentication-service

api/networking/v1/extauth/extauth.proto
kdorosh: We should support the use case where they can bring their own server, so I'd like to leave it in. Will update the comment though since it's a bit dated.

api/networking/v1/extauth/extauth.proto
kdorosh: That wouldn't make much sense here, since the auth config we refer to here is the server config that the ext-auth-service picks up

EItanya: Yes and no, potentially we could translate this into an `AuthConfig` and output it to the remote cluster so the agent can read it.

kdorosh: if we do that, we'd also need to have the agent write it to the remote cluster, since the `AuthConfig` is picked up by ext-auth-service rather than gloo-mesh itself

this brings me back to https://github.com/solo-io/gloo-mesh/pull/1710#pullrequestreview-702385899 , which i had in mind for a follow up after the initial MVP

api/networking/v1/extauth/extauth.proto
kdorosh: opted to follow the istio model here for now: https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig-ExtensionProvider-EnvoyExternalAuthorizationGrpcProvider

i made this a oneof for now so we can add configuration in the future to allow for mesh-external user-provided configuration of the extauth server.

given that istio itself uses the service internal to the mesh for extauth, it seems safe to use similar logic in ours (and I plan to do so): https://github.com/istio/istio/blob/f393a2dde0910ed71749b1604b6e093caa723547/pilot/pkg/security/authz/builder/extauthz.go#L191

my biggest fear was if the `LookupCluster` function signature broke, but since istio already does this too, then we could just copy their upgrade path in the event they ever broke

api/networking/v1/extauth/extauth.proto
EItanya: I agree with that, and I think the agent writing it would be part of this

api/networking/v1/extauth/extauth.proto
EItanya: Just to be clear, this means we need to inject our extauth pod?

api/networking/v1/extauth/extauth.proto
EItanya: In gloo, the AuthConfig referenced here, inevitably ends up being sent to the ext-auth-service. Which is exactly what we should do here. Anything referenced or written here needs to be sent to the ext-auth-service by us. The user shouldn't be configuring the server themselves

api/networking/v1/extauth/extauth.proto
kdorosh: we do not actually, we're just using the discovered cluster from the istio service registry

api/networking/v1/extauth/extauth.proto
kdorosh: agreed, for first pass the agent can just duplicate it to the other clusters.

i can imagine in the future use cases where we want to accept a template of the `AuthConfig` and propagate it with slight changes

api/networking/v1/extauth/extauth.proto
EItanya: Interesting. I assumed that cluster would have mTLS enabled. That feels like most of the value prop of being "in mesh"

api/networking/v1/extauth/extauth.proto
```diff
@@ -2,11 +2,153 @@ syntax = "proto3";
...
+  //   Name: ext-auth-service-$POD_NAMESPACE-$CLUSTER_NAME
+  //   Namespace: $POD_NAMESPACE
+  .core.skv2.solo.io.ObjectRef extauthz_ref = 1;
+
+  // If this is set, communication to the upstream will be via HTTP and not GRPC.
+  HttpService http_service = 2;
+
+  // Timeout for the ext auth service to respond. Defaults to 2000ms. (OIDC requests to external IDPs can be slow)
+  // For latency critical applications, this value should be tuned much lower, as extauth is on the request path.
+  google.protobuf.Duration request_timeout = 3;
```
kdorosh, line 32: just calling out to the reviewer that this timeout has been bumped from gloo edge

.github/workflows/publish_docs.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
       run: |
         make install-go-tools
     - name: Gcloud Login
-      uses: google-github-actions/setup-gcloud@master
+      uses: GoogleCloudPlatform/github-actions/setup-gcloud@7159039b8c614f8ed6d9bd75af3f6113681a8ff4
```
kdorosh, line 35: preferring pinned deps so this doesn't break underneath us again

EItanya (approved):

api/networking/v1/extauth/extauth.proto
kdorosh: to close the loop here, this was updated to use gloo mesh destination abstraction instead

PR #1722, opened 2021-07-13
added short-names

EItanya:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/836


codegen/groups/groups.go
```diff
@@ -19,34 +19,34 @@ var (
...
+	{Kind: "Dashboard", ShortNames: []string{"dash", "dashes"}},
 })
 
 var GlooMeshDiscoveryGroup = makeGroup("discovery", "v1", []ResourceToGenerate{
-	{Kind: "Destination"},
-	{Kind: "Workload"},
-	{Kind: "Mesh"},
+	{Kind: "Destination", ShortNames: []string{"dest", "dests"}},
+	{Kind: "Workload", ShortNames: []string{"wkld", "wklds"}},
+	{Kind: "Mesh", ShortNames: []string{"m"}},
```
harveyxia, line 29: 4 chars is short enough, I don't think we need a shortname for Mesh

harveyxia (commented):
This LGTM, none of the short names overlap with those from Istio CRDs

harveyxia:
Wasn't able to find any k8s documentation on this behavior, but the following was determined through testing.

1. CRDs with a duplicate short name *with another CRD in the same group and version* fail to get installed, silently with exit code 0. This is probably a bug in Kubernetes.
2. CRDs with duplicate short names in *different group versions* will work. Operations referencing the short names unqualified by group/version appears to point at the CRD that is ordered first when listing all CRDs with `kubectl api-resources`

codegen/groups/groups.go
```diff
@@ -19,34 +19,34 @@ var (
...
 	{Kind: "Mesh"},
 })
 
 var GlooMeshNetworkingGroup = makeGroup("networking", "v1", []ResourceToGenerate{
-	{Kind: "TrafficPolicy"},
-	{Kind: "AccessPolicy"},
-	{Kind: "VirtualMesh"},
+	{Kind: "TrafficPolicy", ShortNames: []string{"tp", "tps"}},
+	{Kind: "AccessPolicy", ShortNames: []string{"ap", "aps"}},
+	{Kind: "VirtualMesh", ShortNames: []string{"vm", "vms"}},
```
ashleywang1, line 35: "vmesh" and "vmeshes"?

harveyxia (approved):

PR #1727, opened 2021-07-14
Ratelimit api changes

npolshakova:
Separate the rate limit server and client config in the RateLimitConfig from solo-apis. 

- The RateLimitServerConfig will hold the server side config (descriptors, etc.)
- New RateLimitClientConfig will hold the client config (actions, etc.)

There will be follow ups to add more features in an attempt to bring closer to feature parity with Gloo Edge ("basic" ratelimit API support)

BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/526


api/networking/v1/ratelimit/rate_limit.proto
```diff
@@ -28,132 +28,39 @@ message GatewayRateLimit {
     bool rate_limit_before_auth = 9;
```
kdorosh, line 28: not sure we want to support this in the MVP

api/networking/v1/ratelimit/rate_limit.proto
```diff
@@ -28,132 +28,39 @@ message GatewayRateLimit {
...
     }
 
     // Basic rate-limiting API
-    message BasicRateLimit {
-        // limits for authorized users
-        RateLimitRatio authorized_limits = 1;
-        // limits for unauthorized users
-        RateLimitRatio anonymous_limits = 2;
-
+    message BasicRateLimit {}
```
kdorosh, line 41: should we remove this until it's implemented?

kdorosh: will this become the same as our "basic" rate limit in gloo edge? if so, we should call out that this only works with enterprise and with extauth (also not in the mvp right now, since we need to use the sanitize filter which depends on some of @ilackarms work)

api/networking/v1/traffic_policy.proto
```diff
@@ -66,9 +66,11 @@ message TrafficPolicySpec {
 
         // Configure [outlier detection](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/outlier) on the selected destinations.
         // Specifying this field requires an empty `source_selector` because it must apply to all traffic.
+        // DEPRECATED: use DestinationPolicy to configure Outlier Detection
         OutlierDetection outlier_detection = 11;
```
kdorosh, line 70: let's also mark as deprecated in proto form, e.g. https://stackoverflow.com/questions/52781727/how-can-i-deprecate-whole-message-in-protocol-buffers

api/networking/v1/traffic_policy.proto
```diff
@@ -171,6 +173,7 @@ message TrafficPolicySpec {
 
         // Configure [outlier detection](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/outlier) on the selected destinations.
         // Specifying this field requires an empty `source_selector` because it must apply to all traffic.
+        // DEPRECATED: use DestinationPolicy to configure Outlier Detection
         message OutlierDetection {
```
kdorosh, line 177: same here https://stackoverflow.com/questions/52781727/how-can-i-deprecate-whole-message-in-protocol-buffers

api/networking/v1/ratelimit/rate_limit.proto
npolshakova: This is supposed to be the same as the "basic" ratelimiting in gloo that allows a user to specify anonymous/authenticated limits. If we need the sanitize filter we might need to remove this for now then.

api/enterprise/networking/v1beta1/rate_limit_client_config.proto
```diff
@@ -0,0 +1,35 @@
...
+  .ratelimit.networking.mesh.gloo.solo.io.RateLimitClient rate_limits = 1;
+
+}
+
+// The current status of the `RateLimitClientConfig`.
+message RateLimitClientConfigStatus {
+
+  // The most recent generation observed in the the RateLimiterClientConfig metadata.
+  // If the `observedGeneration` does not match `metadata.generation`,
+  // Gloo Mesh has not processed the most recent version of this resource.
```
kdorosh, line 26: has not processed, or has not successfully processed?

api/networking/v1/traffic_policy.proto
npolshakova: We're not deprecating OutlierDetection yet because it isn't getting moved to DestinationPolicy yet so it'll be configurable in the traffic policy.

api/enterprise/networking/v1beta1/rate_limit_client_config.proto
npolshakova: Has not processed- it's the same status as on the virtual host and route table.

api/networking/v1/traffic_policy.proto
npolshakova: This is no longer going to be deprecated- the OutlierDetection won't be set on DestinationPolicy, it'll still be set on the TrafficPolicy for now.

api/networking/v1/ratelimit/rate_limit.proto
```diff
@@ -22,138 +22,37 @@ message GatewayRateLimit {
...
-
-        // References to RateLimitConfig resources. This is used to configure the GlooE rate limit server.
-        // Only one of `ratelimit` or `rate_limit_configs` can be set.
-        .core.skv2.solo.io.ObjectRefList config_refs = 3;
+// The RateLimitClient specifies either a simplified, abstracted rate limiting model that allows configuring
+// the ratelimit Actions directly (Advanced).
+// The corresponding server config should be set in the RateLimitConfig.
+message RateLimitClient {
+    oneof config_type {
+        AdvancedRateLimit advanced = 1;
```
ilackarms, line 32: comment on this? i'm generally against putting api options out there which make an arbitrary distinction that isn't clear to an outsider. what does "advanced" mean in this context?

ilackarms: what would the alternatives in this oneof be?

api/enterprise/networking/v1beta1/rate_limit_client_config.proto
```diff
@@ -0,0 +1,35 @@
...
+
+import "github.com/solo-io/gloo-mesh/api/networking/v1/ratelimit/rate_limit.proto";
+import "google/protobuf/wrappers.proto";
+
+import "extproto/ext.proto";
+option (extproto.equal_all) = true;
+
+
+// RateLimitClientConfig contains the client configuration for the Gloo Rate Limiter
+message RateLimitClientConfigSpec {
```
ilackarms, line 13: what is the client responsible for in the Gloo Rate limiter? i think we need to clarify the comment here

api/networking/v1/ratelimit/rate_limit.proto
```diff
@@ -22,138 +22,37 @@ message GatewayRateLimit {
...
-        // - rateLimit:
-        //    requestsPerUnit: 20
-        //    unit: MINUTE
-        message RateLimitActions {
-            // TODO: come up with descriptive names and comments for these fields
-            repeated .ratelimit.api.solo.io.Action actions = 1;
-            repeated .ratelimit.api.solo.io.Action set_actions = 2;
-        }
+    // RateLimitConfig ref
+    .core.skv2.solo.io.ObjectRef ratelimit_config_ref = 1;
```
ilackarms, line 52: would it make sense to provide more than one config here?

ilackarms: is this a ref to the `RateLimitClientConfig` ?

ilackarms: if so we should clarify that in the comment

api/networking/v1/ratelimit/rate_limit.proto
npolshakova: The Basic ratelimit (same as the "basic" rate limit in gloo edge that sets the anonymous and authenticated limits) config is the other config that can go here inlined. The basic ratelimit API only works with enterprise and with extauth, but it's currently not in this PR since we need to use the custom envoy sanitize filter for it to work.

api/enterprise/networking/v1beta1/rate_limit_client_config.proto
npolshakova: The RateLimitClient is responsible for setting the client side configuration for the rate limit server (actions, setactions, etc.)

api/networking/v1/ratelimit/rate_limit.proto
npolshakova: It's a ref to the solo-apis RateLimitConfig which is picked up by the ratelimit server.

api/enterprise/networking/v1beta1/rate_limit_client_config.proto
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// The current status of the `RateLimitClientConfig`.
+message RateLimitClientConfigStatus {
+
+  // The most recent generation observed in the the RateLimiterClientConfig metadata.
+  // If the `observedGeneration` does not match `metadata.generation`,
+  // Gloo Mesh has not processed the most recent version of this resource.
+  int64 observed_generation = 1;
+
```
EItanya, line 29: We typically have a state enum, PENDING, FAILED, etc. do we want one here?

api/networking/v1/ratelimit/rate_limit.proto
EItanya: Follow up, if this can be multiple can we use labels?

codegen/groups/groups.go
```diff
@@ -36,13 +36,13 @@ var GlooMeshNetworkingGroup = makeGroup("networking", "v1", []ResourceToGenerate
...
-	{Kind: "WasmDeployment", ShortNames: []string{"wd", "wds"}},
-	{Kind: "RateLimiterServerConfig", ShortNames: []string{"rlsc", "rlscs"}},
-	{Kind: "VirtualDestination", ShortNames: []string{"vd", "vds"}},
-	{Kind: "VirtualGateway", ShortNames: []string{"vg", "vgs"}},
-	{Kind: "VirtualHost", ShortNames: []string{"vh", "vhs"}},
-	{Kind: "RouteTable", ShortNames: []string{"rt", "rts"}},
-	{Kind: "ServiceDependency", ShortNames: []string{"sd", "sds"}},
+	{Kind: "WasmDeployment"},
+	{Kind: "RateLimitClientConfig"},
+	{Kind: "VirtualDestination"},
```
EItanya, line 41: Readd short names

api/enterprise/networking/v1beta1/rate_limit_client_config.proto
```diff
@@ -0,0 +1,60 @@
...
+option (extproto.equal_all) = true;
+
+
+// RateLimitClientConfig contains the client configuration for the rate limit Actions that determine how Envoy
+// composes the descriptors that are sent to the rate limit server to check whether a request should be rate-limited
+message RateLimitClientConfigSpec {
+
+  // The RateLimitClient specifies the ratelimit Actions which the client (Envoy) will use to
+  // compose the descriptors that will be sent to the server to make a rate limiting decision.
+  .ratelimit.networking.mesh.gloo.solo.io.RateLimitClient rate_limits = 1;
```
EItanya, line 18: Why is this nested, is it used elsewhere?

npolshakova: It's used in the rate_limit.proto. The config can be set as a ref to the RateLimitClient

api/enterprise/networking/v1beta1/rate_limit_server_config.proto
```diff
@@ -2,50 +2,78 @@ syntax = "proto3";
...
-// RateLimiterConfig contains the configuration for the Gloo Rate Limiter, the external rate-limiting server used by
-// mesh proxies to rate-limit HTTP requests. One or more rate limiter servers may be deployed in order to
-// rate limit traffic across East-West and North-South routes. The RateLimiterConfig allows users to map
-// a single rate-limiter configuration to multiple rate-limiter server instances, deployed across managed clusters.
-message RateLimiterServerConfigSpec {
-    // The per-server rate limit config objects will be generated from the given config for each provided ref.
-    // Each rate limit server must be configured to read its server configuration from one of these refs.
-    repeated .core.skv2.solo.io.ObjectRef server_config_refs = 1;
+    // The set descriptors that will be applied to the server.
+    repeated .ratelimit.api.solo.io.SetDescriptor set_descriptors = 2 [(cue.opt).disable_openapi_validation = true];
```
ilackarms, line 28: it could be worth documenting that these fields are currently not structurally validated by k8s

ilackarms: also, nit: prefer to put nested messages below the fields in the message they're defined in

api/networking/v1/ratelimit/rate_limit.proto
ilackarms: i'm still hesitant to use the term `advanced` as it has no meaning to someone who's not already familiar with gloo edge & our implementation details. how about using `native_envoy` or `raw`? does it make sense that the name of the field we select here should would match what's currently named `RateLimitServerConfigSpec.raw` ? they seem to be matching conceptually. are those fields required to be aligned?

api/networking/v1/ratelimit/rate_limit.proto
```diff
@@ -22,138 +22,47 @@ message GatewayRateLimit {
...
-    message BasicRateLimit {
-        // limits for authorized users
-        RateLimitRatio authorized_limits = 1;
-        // limits for unauthorized users
-        RateLimitRatio anonymous_limits = 2;
+// Use this field if you want to inline the Envoy rate limits.
+// Note that this does not configure the rate limit server. If you are running Gloo Mesh, you need to
+// specify the server configuration via the appropriate field in the Gloo Mesh `RateLimitConfig` resource.
+// If you are running a custom rate limit server you need to configure it yourself.
+message AdvancedRateLimit {
```
ilackarms, line 40: same as above wrt name

api/networking/v1/ratelimit/rate_limit.proto
```diff
@@ -22,138 +22,47 @@ message GatewayRateLimit {
     google.protobuf.Duration request_timeout = 2;
```
ilackarms, line 23: these fields should be commented especially ratelimit_server_ref. do we have a default we should note there?

ilackarms (approved):

PR #1729, opened 2021-07-15
Add API for GM managed Istio installations

jahvon:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/848


api/enterprise/istio/v1/istio_installation.proto
```diff
@@ -0,0 +1,52 @@
...
+message IstioInstallationSpec {
+    // The cluster where the IstioOperators should be installed
+    string cluster_name = 1;
+
+    // References for the IstioOperators that should be installed in the managed cluster
+    repeated .core.skv2.solo.io.ObjectRef istio_operator_refs = 2;
+}
+
+message IstioInstallationStatus {
+    message IstioOperatorStatus {
```
jahvon, line 17: I was also considering adding `revision` to the status so that a user knows which revision the set of resources would have. Is this a good place to include that information?

api/enterprise/istio/v1/istio_installation.proto
```diff
@@ -0,0 +1,52 @@
...
+option go_package = "github.com/solo-io/gloo-mesh/pkg/api/istio.enterprise.mesh.gloo.solo.io/v1";
+
+import "github.com/solo-io/skv2/api/core/v1/core.proto";
+
+message IstioInstallationSpec {
+    // The cluster where the IstioOperators should be installed
+    string cluster_name = 1;
+
+    // References for the IstioOperators that should be installed in the managed cluster
+    repeated .core.skv2.solo.io.ObjectRef istio_operator_refs = 2;
```
jahvon, line 13: Alternative is using a label selector (`.core.skv2.solo.io. ObjectSelector`) here

api/enterprise/istio/v1/istio_installation.proto
```diff
@@ -0,0 +1,52 @@
+syntax = "proto3";
+package istio.enterprise.mesh.gloo.solo.io;
+
+option go_package = "github.com/solo-io/gloo-mesh/pkg/api/istio.enterprise.mesh.gloo.solo.io/v1";
+
+import "github.com/solo-io/skv2/api/core/v1/core.proto";
+
+message IstioInstallationSpec {
+    // The cluster where the IstioOperators should be installed
```
harveyxia, line 9: Let's use proper punctuation for doc strings

```suggestion
    // The cluster where the IstioOperators should be installed.
```

api/enterprise/istio/v1/istio_installation.proto
```diff
@@ -0,0 +1,52 @@
...
+    string cluster_name = 1;
+
+    // References for the IstioOperators that should be installed in the managed cluster
```
harveyxia, line 12:
```suggestion
    // References to the IstioOperator CRs whose installation parameters should be used when installing Istio to the managed cluster.
```

api/enterprise/istio/v1/istio_installation.proto
```diff
@@ -0,0 +1,52 @@
...
+
+message IstioInstallationStatus {
+    message IstioOperatorStatus {
+        // The state of a IstioOperator installation
+        enum State {
+            // Waiting for resources to be reconciled
+            PENDING = 0;
+
+            // In the process of reconciling Istio resources on to the managed cluster
+            RECONCILING = 1;
```
harveyxia, line 24: `RECONCILING` feels a bit implementation specific, maybe rename to something like `INSTALLING`?

api/enterprise/istio/v1/istio_installation.proto
```diff
@@ -0,0 +1,52 @@
...
+
+            // All Istio components were installed successfully and they are healthy
+            HEALTHY = 2;
+
+            // One or more installed Istio component(s) in an error state
```
harveyxia, line 29:
```suggestion
            // This Istio installation is in an error state.
```

api/enterprise/istio/v1/istio_installation.proto
harveyxia: Yeah this status would be the place for that kind of metadata. You can also add those additional data fields as they're needed down the line

harveyxia: nit: our convention is to place message definitions at the bottom of the enclosing message

api/enterprise/istio/v1/istio_installation.proto
```diff
@@ -0,0 +1,52 @@
+syntax = "proto3";
+package istio.enterprise.mesh.gloo.solo.io;
+
+option go_package = "github.com/solo-io/gloo-mesh/pkg/api/istio.enterprise.mesh.gloo.solo.io/v1";
+
+import "github.com/solo-io/skv2/api/core/v1/core.proto";
+
+message IstioInstallationSpec {
+    // The cluster where the IstioOperators should be installed
+    string cluster_name = 1;
```
ilackarms, line 10: we probably want to specify the namespace as well here

api/enterprise/istio/v1/istio_installation.proto
ilackarms: are we installing istio or the istio operator? im not sure istio operator has a revision

api/enterprise/istio/v1/istio_installation.proto
```diff
@@ -0,0 +1,52 @@
...
+        string message = 3;
+    }
+
+    // The most recent generation observed in the the IstioInstallation metadata.
+    // If the `observedGeneration` does not match `metadata.generation`,
+    // Gloo Mesh has not processed the most recent version of this resource.
+    int64 observed_generation = 1;
+
+    // The status of the installation for each IstioOperator that should be applied.
+    map<string, IstioOperatorStatus> statuses = 2;
```
ilackarms: what is key in this map?

api/enterprise/istio/v1/istio_installation.proto
```diff
@@ -0,0 +1,52 @@
...
+        // If the `observedGeneration` does not match `metadata.generation`,
+        // Gloo Mesh has not processed the most recent version of this resource.
+        int64 observed_generation = 1;
+
+        // The current state of the IstioOperator
+        State state = 2;
+
+        // A human readable message about the current state of the IstioOperator
+        string message = 3;
+    }
```
ilackarms: nit: we've been going with putting inline declarations at the bottom of the message where they're specified and putting the fields at the top. i think it's easier wrt readability, i.e. we see how the message is used before we see its contents (same as the ordering of the proto file itself, from the top of the file to the bottom the messages go from top-level to bottom-level)

api/enterprise/istio/v1/istio_installation.proto
jahvon: I didn't specify the namespace because I was just going to use the namespace that is defined on the individual IstioOperator CRs (see https://istio.io/latest/docs/reference/config/istio.operator.v1alpha1/). Does that make sense here?

api/enterprise/istio/v1/istio_installation.proto
jahvon: I was planning on mapping from the IstioOperator name to the operator status. Does that make sense here? I can also update the comment

api/enterprise/istio/v1/istio_installation.proto
```diff
@@ -0,0 +1,56 @@
+syntax = "proto3";
+package istio.enterprise.mesh.gloo.solo.io;
```
jahvon, line 2: Does this new `istio` package make sense? Or should this be encapsulated under something like discovery?

harveyxia: I think `istio` makes sense

api/enterprise/istio/v1/istio_installation.proto
EItanya: I definitely think we should favor label selectors whenever possible

api/enterprise/istio/v1/istio_installation.proto
```diff
@@ -0,0 +1,56 @@
...
+
+message IstioInstallationStatus {
+    // The most recent generation observed in the the IstioInstallation metadata.
+    // If the `observedGeneration` does not match `metadata.generation`,
+    // Gloo Mesh has not processed the most recent version of this resource.
+    int64 observed_generation = 1;
+
+    // The status of the installation for each IstioOperator that should be applied, where the
+    // key is the IstioOperator `metadata.name` and the value is the operator's status.
+    map<string, IstioOperatorStatus> statuses = 2;
```
EItanya, line 24: This map seems to represent the status of the IstioOperator CRs themselves, in the remote cluster. I think we may also want the status of this resource itself. If it was able to be translated, validate missing refs, etc.

api/enterprise/istio/v1/istio_installation.proto
ilackarms: i don't think `istio` captures the domain we're dealing with here. let's think about other CRs which may fit together with such a CR as this. likely they will be related to installation of other components, or even provisioning resources on external (cloud) environments. therefore i actually consider this to be a part of `admin` or `operations`

ilackarms: (all of our api's are `istio` apis) after all

api/enterprise/admin/v1/istio_installation.proto
```diff
@@ -19,9 +19,33 @@ message IstioInstallationStatus {
     // Gloo Mesh has not processed the most recent version of this resource.
     int64 observed_generation = 1;
 
-    // The status of the installation for each IstioOperator that should be applied, where the
-    // key is the IstioOperator `metadata.name` and the value is the operator's status.
-    map<string, IstioOperatorStatus> statuses = 2;
+    // The overall status of the Istio installation across the associated clusters.
+    IstioInstallationStatus status = 2;
```
EItanya, line 23: nit: This object should be flattened into `IstioInstallationStatus`

codegen/groups/groups.go
```diff
@@ -54,13 +54,18 @@ var GlooMeshEnterpriseRbacGroup = makeGroup("rbac.enterprise", "v1", []ResourceT
 	{Kind: "RoleBinding", ShortNames: []string{"gmrolebinding", "gmrolebindings"}},
 })
 
+var GlooMeshEnterpriseAdminGroup = makeGroup("admin.enterprise", "v1alpha1", []ResourceToGenerate{
+	{Kind: "IstioInstallation"},
```
EItanya, line 58: Can we leave a TODO here for short names, I don't think we need to come up with them now, but I want to signal for later

EItanya (approved):

EItanya (dismissed):

EItanya (approved):

PR #1730, opened 2021-07-16
[Gateway] Add weight and sort field to route tables

ryantking:


api/enterprise/networking/v1beta1/route.proto
```diff
@@ -131,10 +131,26 @@ message DirectResponseAction {
...
+        // After processing all routes, including additional route tables delegated to, the resulting routes are sorted
+        // by specificity to reduce the chance that a more specific route will be short-circuited by a general route.
+        // Matchers with exact path matchers are considered more specific than regex path patchers, which are more
+        // specific than prefix path matchers. Matchers of the same type are sorted by length of the path in descending
+        // order. Only the most specific matcher on each route is used.
+        ROUTE_SPECIFICITY = 1;
+    }
+
+    // How routes should be sorted
+    SortMethod sortMethod = 3;
```
ilackarms, line 155: to make sure the UI code generates properly, we need to use snake case (and should because it's the convention)

Sodman (approved):
LGTM! 🚀

PR #1735, opened 2021-07-19
Add pre-install check for validating relay-server-address

harveyxia:


pkg/meshctl/commands/install/install.go
```diff
@@ -210,6 +208,25 @@ func enterpriseCommand(ctx context.Context, installOpts *Options) *cobra.Command
...
+				0,
+				&checks.ServerParams{
+					RelayServerAddress: opts.RelayServerAddress,
+				},
+			)
+			if err != nil {
+				return err
+			}
+
+			if foundFailure := checkCtx.RunChecks(ctx, checks.Server, checks.PreInstall); foundFailure {
```
yuval-k, line 226: should we also add a `--ignore-pre-checks` flag in case we have a bug in a prehceck?

pkg/meshctl/validation/checks/server_parameters.go
```diff
@@ -0,0 +1,45 @@
...
+	result := &Result{}
+
+	// update result with all server paramater checks
+	isValidRelayServerIp(serverParams.RelayServerAddress, result)
+
+	return result
+}
+
+// return true is relay server IP exists and is a valid IP
+func isValidRelayServerIp(ip string, result *Result) {
```
yuval-k, line 36: realy server address can also be a dns name and also may have a port. so we should only check that it is a valid ip if it is not a dns name?
for example:
https://github.com/golang/go/blob/404127c30f3f66eb234da2dc1a78e6eea0ef4ee0/src/net/url/url.go#L619

we might even be able to just use url.Parse? (and verify that nothing but host is set?)
i would also add a unit test to verify this logic with test cases:
- `<ipv4>`
- `<ipv4>:port`
- `<ipv6>`
- `<ipv6>:port`
- `<dnsname>`
- `<dnsname>:port`
- `<invalid-dns-name>`

pkg/meshctl/validation/checks/interfaces.go
```diff
@@ -25,16 +25,38 @@ const (
...
+}
+func (c *CommonContext) Client() client.Client {
+	return c.Cli
+}
+
 type CheckContext interface {
 	Environment() Environment
 	Client() client.Client
 	AccessAdminPort(ctx context.Context, deployment string, op OperateOnAdminPort) (error, string)
+	Context() CommonContext
```
yuval-k, line 56: Why is CommonContext separated? will consuming code use it without CheckContext?

harveyxia: I'm placing any input parameters that need validation into CommonContext, such as [`ServerParams`](https://github.com/solo-io/gloo-mesh/pull/1735/files#diff-77fa5867996fa36312d65e6c73a5296c2bcf5c17351fc241e9969216e61b620aR42), open to other suggestions for passing these parameters though

pkg/meshctl/commands/install/install.go
harveyxia: Added a `--skip-checks` flag for meshctl as well as a `SKIP_CHECKS` env var for the helm hook

pkg/meshctl/validation/checks/server_parameters.go
harveyxia: updated with new verification method and unit tests

yuval-k (approved):

PR #1737, opened 2021-07-20
Add istio installation CRs to snapshot

jahvon:
WIP


EItanya:
Should these go in their own "component" like the cert issuer?
I'm not sure they fit into the networking translator. Especially because well be relying on the remote snapshot, and the networking snapshot is not great at that.
@ilackarms @harveyxia what do y'all think?

Issue #1719, opened 2021-07-12
Expose Istio AuthorizationPolicy's `(not)requestPrincipals` on AccessPolicy 


harveyxia:
Expose `(not)requestPrincipals` (maps to Istio AuthorizationPolicy's `requestPrincipals` and `notRequestPrincipals` fields, [see here](https://istio.io/latest/docs/reference/config/security/authorization-policy/#Source)) on AccessPolicy.

PR #1738, opened 2021-07-20
expose (not)requestPrinciples on access policy

ryantking:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1719


api/common/v1/selectors.proto
```diff
@@ -107,6 +107,12 @@ message IdentitySelector {
             When used in a Gloo Mesh Role, a wildcard (`"*"`) must be specified to match any cluster.
         */
         repeated string clusters = 2;
+
+        // An optional list of request identities (e.g. iss/sub claims) to the mTLS principal.
```
lgadban, line 111: This probably should be slightly reworded as this identity does not come from the mTLS peer but rather the JWT used for `RequestAuthentication`

lgadban:
So would the semantics be you could define an `AccessPolicy` with empty `sourceSelector` but a defined `destinationSelector` for JWT rules?

harveyxia:
> So would the semantics be you could define an AccessPolicy with empty sourceSelector but a defined destinationSelector for JWT rules?

The `sourceSelector` in the context of AccessPolicy refers to  the [identity of the incoming source](https://github.com/solo-io/gloo-mesh/blob/main/api/networking/v1/access_policy.proto#L20), not the source workload (perhaps we should have named the field `identitySelector` instead of `sourceSelector`.)

api/common/v1/selectors.proto
```diff
@@ -107,6 +107,12 @@ message IdentitySelector {
...
+        repeated string requestPrincipals = 3;
+
+        // An optional list of request identities to not match.
+        repeated string notRequestPrincipals = 4;
```
yuval-k, line 115: if both are provided who wins?

api/common/v1/selectors.proto
```diff
@@ -107,6 +107,12 @@ message IdentitySelector {
             When used in a Gloo Mesh Role, a wildcard (`"*"`) must be specified to match any cluster.
         */
         repeated string clusters = 2;
+
+        // An optional list of request identities (e.g. iss/sub claims) to the mTLS principal.
+        repeated string requestPrincipals = 3;
```
yuval-k, line 112: these are basically copy pasted into the istio auth policy?
let's in the doc comment how to use these and what values are valid, otherwise there is no way of knowing.

lgadban:
>The sourceSelector in the context of AccessPolicy refers to the identity of the incoming source, not the source workload (perhaps we should have named the field identitySelector instead of sourceSelector.)

@harveyxia okay understood, I misread that these new fields are being added to the `sourceSelector`. Seeing a full example would help.

More generally I see this API as slightly confusing, as the request principals are being added to the `KubeIdentityMatcher` message. The request principals are not really associated with k8s related identity as they come from a JWT.

Not accounting for the challenges of changing the API, IMO it would be a better fit to have a separate `IdentitySelector` that is specifically for JWT/request auth, for example something like `RequestIdentity` which would then contain the Istio AuthZ `[not]requestPrincipals`.

harveyxia:
> it would be a better fit to have a separate IdentitySelector

I agree, we should make a separate message to represent RequestIdentity and keep it distinct from the existing `KubeIdentity` cc @ryantking

api/common/v1/selectors.proto
```diff
@@ -92,6 +92,10 @@ message IdentitySelector {
     // Match request identities based on the Kubernetes service account of the request.
     KubeServiceAccountRefs kube_service_account_refs = 2;
 
+    // Match requests based on the identity of the source. If multiple fields are set, they are ANDed together.
```
harveyxia, line 95:
```suggestion
    // Match requests based on the identity of the request. If multiple fields are set, they are ANDed together.
```

api/common/v1/selectors.proto
```diff
@@ -92,6 +92,10 @@ message IdentitySelector {
     // Match request identities based on the Kubernetes service account of the request.
     KubeServiceAccountRefs kube_service_account_refs = 2;
 
+    // Match requests based on the identity of the request. If multiple fields are set, they are ANDed together.
```
lgadban, line 95: nit: since the request identity comes from a JWT (and only a JWT currently) it may make sense to call that out here somehow?

I'm also fine leaving it as is

lgadban:
One minor nit on the docs but otherwise LGTM

harveyxia (dismissed):

harveyxia (approved):

PR #1739, opened 2021-07-20
Support intermediate root ca

EItanya:


EItanya:
Follow up closed source PR https://github.com/solo-io/gloo-mesh-enterprise/pull/902

pkg/certificates/agent/reconciliation/cert_agent_reconciler.go
```diff
@@ -169,14 +176,20 @@ func (r *certAgentReconciler) reconcileIssuedCertificate(
 			return err
 		}
 
-		if err := r.translator.IssuedCertificateRequested(
+		wait, err := r.translator.IssuedCertificateRequested(
```
harveyxia, line 179: Pointing this out for context—without adding this `wait` condition, the previous behavior was to immediately create the PodBounceDirective once the issued certificate enters the issued state. This potentially explains the `CERTFICIATE_VERIFY_FAILED` CI flakes [documented here](https://github.com/solo-io/gloo-mesh-enterprise/issues/887), because workloads would be restarted before the new certificate was fully issued.

pkg/certificates/agent/utils/cert_gen.go
```diff
@@ -57,15 +61,28 @@ func GenerateCertificateSigningRequest(hosts []string, org string, privateKey []
 		return nil, eris.Wrapf(err, "unable to decode private key, currently only supporting PKCS1 encrypted keys")
 	}
 
+	var hostsStr string
```
harveyxia, line 64: this block isn't necessary (i.e. we can keep the old inlined invocation of join), `strings.Join(hosts, ",")` will just return the first element of `hosts` if it's of length 1

pkg/certificates/common/secrets/cert_secrets.go
```diff
@@ -10,62 +19,74 @@ const (
...
+	chains, err := cert.Verify(opts)
 
-func (c *RootCAData) ToSecretData() map[string][]byte {
-	return map[string][]byte{
-		RootPrivateKeyID: c.PrivateKey,
-		RootCertID:       c.RootCert,
+	if len(chains) == 0 || err != nil {
+		return eris.Wrap(
+			err,
+			"cannot verify the cert with the provided root chain and cert pool with error",
```
harveyxia, line 57:
```suggestion
			"failed to verify the cert with the provided root chain and cert pool",
```

pkg/certificates/issuer/translation/cert_issuer_translator.go
```diff
@@ -72,21 +75,30 @@ func (s *secretTranslator) Translate(
 		return nil, eris.Wrapf(err, "failed to find issuer's signing certificate matching issued request %v", sets.Key(issuedCertificate))
 	}
 
-	signingCA := secrets.RootCADataFromSecretData(signingCertificateSecret.Data)
+	signingCA := secrets.CADataFromSecretData(signingCertificateSecret.Data)
+	if err := signingCA.Verify(); err != nil {
+		return nil, eris.Wrapf(
+			err,
+			"Could not verify signing CA data from secret %s",
```
harveyxia, line 82: Can we make this consistent with the error below? `Provided CA (%s) is invalid", sets.Key(secret)`

pkg/certificates/agent/utils/cert_gen.go
```diff
@@ -57,15 +61,28 @@ func GenerateCertificateSigningRequest(hosts []string, org string, privateKey []
...
 		SignerPrivPem: privateKey,
 	})
 	if err != nil {
 		return nil, eris.Wrap(err, "CSR template creation failed")
 	}
 
+	// We add the cluster name to the new CSR template subject.
+	// This is extremely important to identify this new certificate as different from it's parent,
+	// and others it's communicating with.
+	if meshName == "" {
```
harveyxia, line 81: What can happen if we don't set this field?

pkg/mesh-networking/translation/istio/mesh/mtls/mtls_translator.go
```diff
@@ -67,9 +71,12 @@ var (
 	}
 )
 
+// IsSigningCert determines whether a cert can be used as a signing cert
 // used by networking reconciler to filter ignored secrets
+// Handle deprecated cert type.
```
harveyxia, line 76: Add a TODO here indicating the need to keep this updated with all future secret types representing signing secrets

harveyxia: Stepping back, the only purpose of this is to optimize our reconciler such that it can ignore any secrets that aren't related to our cert management. @ilackarms could we remove this check entirely, or would that cause too much of a performance hit?

pkg/mesh-networking/translation/istio/mesh/mtls/mtls_translator.go
```diff
@@ -212,6 +219,18 @@ func (t *translator) configureSharedTrust(
 			// Set deprecated field for backwards compatibility
 			issuedCertificate.Spec.SigningCertificateSecret = rootCaSecret
 		case *networkingv1.RootCertificateAuthority_Secret:
```
harveyxia, line 221: Would it be worth also verify our own generated secret? Or would errors be caught by our tests?

test/e2e/happypath/README.md
```diff
@@ -0,0 +1,24 @@
+# Istio E2e Tests
+
+The Istio End-to-End tests provide the most robust single test suite in the Gloo Mesh.
```
harveyxia, line 3: nit: this is a bit misleading, this test suite, as i understand it from the context, tests our cert management for managed service meshes, whereas "robust" implies that it's a comprehensive test of all GM functionality, generally speaking

pkg/mesh-networking/translation/istio/mesh/mtls/mtls_translator.go
EItanya: Errors would be caught by our tests, I think we can assume the certs are valid, if they cause a problem with istio e2e will catch it

pkg/certificates/agent/utils/cert_gen.go
EItanya: At a high level the certs will not generate properly, leading to a SSL_VERIFIED error in istio

test/e2e/happypath/README.md
EItanya: It's a copy pasta, I'll fix

pkg/certificates/agent/utils/cert_gen.go
EItanya: It may be length 0 though, we don't always need hosts

EItanya: Still seems to work properly with 0, I will revert

pkg/mesh-networking/reconciliation/networking_reconciler.go
```diff
@@ -322,16 +317,6 @@ func (r *networkingReconciler) syncSettings(ctx *context.Context, in input.Local
 	})
 }
 
-// returns true if the passed object is a secret which is of a type that is ignored by GlooMesh
-// TODO: this limits us to using our secret type, so products like cert-manager won't work.
-func isIgnoredSecret(obj metav1.Object) bool {
```
EItanya, line 327: I removed this secret predicate as it will most likely lead to confusing behavior with a provided root certificate.
1. Most people create secrets using `kubectl` which creates `Opaque` secrets. We need to be able to handle those.
2. Using a secret client isn't ideal here either, because if the user updates the secret to be valid/invalid, we want to make sure to run the networking translator again.

ilackarms (approved):

PR #1747, opened 2021-07-26
fix setup-gcloud Github Action

harveyxia:
see the [notice from the setup-gcloud repo](https://github.com/google-github-actions/setup-gcloud#-notice)


jahvon (approved):

kdorosh (commented):
fyi, for the future i think we should err on the side of pinning our dependencies like this where possible, rather than relying on bleeding edge. this leads to hermetic builds which makes things like supporting LTS branches much easier for devs

PR #1750, opened 2021-07-27
Use ingress gateway selector API

ryantking:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/857


api/enterprise/networking/v1beta1/virtual_gateway.proto
```diff
@@ -338,7 +322,7 @@ message VirtualGatewayStatus{
     // Any warnings found while processing this generation of the resource.
     repeated string warnings = 3;
 
-    repeated SelectedGatewayWorkload selected_gateways = 4;
+    repeated .discovery.mesh.gloo.solo.io.AppliedIngressGateway = 4;
```
Sodman, line 325: This needs a field name

```suggestion
    repeated .discovery.mesh.gloo.solo.io.AppliedIngressGateway selected_gateways = 4;
```

api/enterprise/networking/v1beta1/virtual_gateway.proto
```diff
@@ -322,6 +303,9 @@ message VirtualGatewaySpec {
         // For more info, check out the [Envoy docs](https://www.envoyproxy.io/docs/envoy/v1.17.1/api-v3/config/listener/v3/listener.proto)
         google.protobuf.UInt32Value per_connection_buffer_limit_bytes = 1;
 
+        // Only allow clients to connect from this address.
+        string bind_address = 2;
```
kdorosh, line 307: let's include the default values in the comments

Sodman: I think the original comment also was more accurate. This is the network interface that we're _serving_ traffic on, rather than a whitelist of addresses that are allowed to connect.

Sodman (commented):
LGTM! Let's hold off merging into main though until we have a GME PR that compiles with these proto changes.

api/common/v1/status.proto
```diff
@@ -26,3 +28,17 @@ enum ApprovalState {
     // the desired state.
     FAILED = 3;
 }
+
+message AppliedIngressGateway {
```
EItanya, line 32: This has been changed on main: https://github.com/solo-io/gloo-mesh/commit/35183e059feb5f3d30a51f3a02a0a587f5ae71bb#diff-ca467ebeadf4904622779ddf521625907a9c42b57405e2d592a1c02ff9b39cf2

api/discovery/v1/mesh.proto
```diff
@@ -191,21 +192,7 @@ message MeshStatus {
...
-        // The externally accessible address(es) for this ingress gateway Destination.
-        repeated string external_addresses = 2;
-
-        // The port on the ingress gateway Destination designated for receiving cross cluster traffic.
-        uint32 destination_port = 3;
-
-        // The port on the ingress gateway's backing Workload(s) designated for receiving cross cluster traffic.
-        uint32 container_port = 4;
-    }
+    repeated .common.mesh.gloo.solo.io.AppliedIngressGateway applied_east_west_ingress_gateways = 4;
```
EItanya, line 195: This has also been renamed on main

Sodman (commented):
The changes LGTM but can we update the existing gateway docs/examples to swap deployToIngressGateways => ingressGatewaySelectors, move bindAddress etc?

docs/content/guides/gateway/getting_started.md
```diff
@@ -54,16 +54,16 @@ metadata:
   name: demo-gateway
   namespace: gloo-mesh
 spec:
-  deployToIngressGateways:
-    gatewayWorkloads:
-    - kubeWorkloadMatcher:
+  ingressGatewaySelectors:
+  - portName: tls
```
Sodman, line 58: I think the port name we want here is probably `http2` rather than `tls`. Here are the default ports that istio ships with when you install the default istio-ingressgateway:

image

Can we also update L43 in this doc to something like "Note we're selecting the port named `http2`, which in our case corresponds to the port `8080` on our `istio-ingressgateway` service"

EItanya (dismissed):

install/helm/gloo-mesh-crds/crds/networking.enterprise.mesh.gloo.solo.io_v1beta1_crds.yaml
```diff
@@ -2317,9 +2260,85 @@ spec:
...
+                                  namespace:
+                                    description: namespace of the resource being referenced
+                                    type: string
+                                type: object
+                              type: array
+                          type: object
+                      type: object
+                    type: array
+                  portName:
+                    description: Specify by name the TLS port on the ingress gateway
```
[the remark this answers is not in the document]
EItanya: The comment is wrong, we can fix it to be more clear, as far as portName, we can add portNumber in a follow up for sure

install/helm/gloo-mesh-crds/crds/networking.enterprise.mesh.gloo.solo.io_v1beta1_crds.yaml
```diff
@@ -2317,9 +2260,85 @@ spec:
                     type: object
                 type: object
               type: array
+            ingressGatewaySelectors:
```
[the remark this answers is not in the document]
ryantking: We currently only support opening one port per service. If you want multiple ports, you'll need multiple services with the same backing workloads.

install/helm/gloo-mesh-crds/crds/networking.enterprise.mesh.gloo.solo.io_v1beta1_crds.yaml
```diff
@@ -2317,9 +2260,85 @@ spec:
...
+                  portName:
+                    description: Specify by name the TLS port on the ingress gateway
+                      destination. If not specified, will default to "tls".
+                    type: string
+                type: object
+              type: array
             options:
               description: Options applied to all clients who connect to this gateway
               properties:
+                bindAddress:
```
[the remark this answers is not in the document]
ryantking: The bind address is not yet implemented for NS traffic, but I believe when it is, the port is specified separately. What is the corresponding istio/envoy filled you're referring to?

Sodman (commented):
LGTM. Let's merge once we know the GME PR is in a good spot

test/e2e/istio/pkg/tests/federation.go
```diff
@@ -359,7 +359,7 @@ func FederationTest() {
 							},
 						},
 					},
-					PortName: "tls",
+					PortName: "http2",
```
EItanya, line 362: Why did this change? We still definitely want to use the `tls` port

docs/content/guides/gateway/getting_started.md
```diff
@@ -54,16 +54,16 @@ metadata:
   name: demo-gateway
   namespace: gloo-mesh
 spec:
-  deployToIngressGateways:
-    gatewayWorkloads:
-    - kubeWorkloadMatcher:
+  ingressGatewaySelectors:
+  - portName: http2
```
EItanya, line 58: Why are we defaulting to `http2` for these examples, we should switch to `http` as these examples are using plaintext

EItanya (changes requested):
Switch port names

api/common/v1/status.proto
```diff
@@ -26,3 +28,19 @@ enum ApprovalState {
...
+
+    // The Destination on the mesh that acts as an ingress gateway for the mesh.
+    .core.skv2.solo.io.ClusterObjectRef destination_ref = 1;
```
EItanya, line 35: Why did this become a ClusterObjectRef since destinations are always local to the management-cluster?

EItanya (approved):

docs/content/guides/gateway/getting_started.md
Sodman: We spoke offline about this - just to close the loop: Istio's default installation includes an istio-ingressgateway that uses a Service which names the port `http2`. Our selector is based on this Service's port's name, which is why we use http2 here. The actual protocol that will be used by the Envoy Listener is determined by whether the listener is HTTP or TCP, and whether SSLConfig is set (HTTPS/TLS).

PR #1751, opened 2021-07-28
Bump helm to latest

Sodman:
Makes Dependabot happy.


Sodman:
Looks like we'll need to bump a few other deps for compatibility, closing for now.

PR #1768, opened 2021-08-03
Add Prefix rewrite field to gateway API

Sodman:
Corresponding implementation - https://github.com/solo-io/gloo-mesh-enterprise/pull/969


api/enterprise/networking/v1beta1/route.proto
```diff
@@ -117,7 +123,6 @@ message RedirectAction {
     }
 }
 
-// TODO: Shuold we be using v4alpha now? https://github.com/envoyproxy/envoy/blob/5a8bfa20dc3c85ecb61826d122696ecaa75dffa0/api/envoy/config/route/v4alpha/route_components.proto#L1396
```
npolshakova, line 120: I think the reason we didn't use v4alpha is it isn't in the go control plane version we're currently using. It is now in the latest: https://github.com/envoyproxy/go-control-plane/blob/main/envoy/config/route/v4alpha/route_components.pb.go At some point we should upgrade so that body is a *v4alpha.DataSource instead of a string.

npolshakova (approved):

PR #1771, opened 2021-08-03
Gateway TLS API

Sodman:
See draft implementation [here](https://github.com/solo-io/gloo-mesh-enterprise/pull/980)


api/enterprise/networking/v1beta1/virtual_gateway.proto
```diff
@@ -371,29 +300,67 @@ message SelectedGatewayWorkload {
...
+
+        // SSLFiles reference paths to certificates which can be read by the proxy off of its local filesystem
+        SSLFiles ssl_files = 2;
+
+    }
+
+
+
+    // optional. the SNI domains that should be considered for TLS connections
+    repeated string sni_domains = 4;
```
EItanya, line 319: How does this field interact with ConnectionMatch.server_names?

api/enterprise/networking/v1beta1/virtual_gateway.proto
```diff
@@ -371,29 +300,67 @@ message SelectedGatewayWorkload {
...
-    string target_uri = 1;
-    oneof sds_builder {
-        // Call credentials.
-        CallCredentials call_credentials = 2;
-        // The name of the sds cluster in envoy
-        string cluster_name = 3;
+
+// SslConfig contains the options necessary to configure a virtual host or listener to use TLS
+message SslConfig {
+    oneof ssl_secrets {
```
EItanya, line 306:
```suggestion
    oneof certificates {
```

api/enterprise/networking/v1beta1/virtual_gateway.proto
```diff
@@ -123,13 +121,8 @@ message VirtualGatewaySpec {
...
 
             message TlsTerminationOptions {
-                // If specified, the route will match against whether or not a certificate is presented.
-                // If not specified, certificate presentation status (true or false) will not be considered when route matching.
-                google.protobuf.BoolValue presented = 1;
-
-                // If specified, the route will match against whether or not a certificate is validated.
-                // If not specified, certificate validation status (true or false) will not be considered when route matching.
-                google.protobuf.BoolValue validated = 2;
+                // SslConfig contains the options necessary to configure a virtual host or listener to use TLS
```
EItanya, line 124: Why do we have a nested object here, since SslConfig is all of the options, can't we flatten it out in this object?

api/enterprise/networking/v1beta1/virtual_gateway.proto
Sodman: Great question! The ConnectionMatch.server_names is used in the event of multiple connection handlers, to determine which connection handler should be used for the request. The connectionHandler.connectionMatch.tlsContext.sniDomains is meant to list which domains have TLS applied. I don't think we'd be able to support a subset though right? Ie connection match on foo.com & bar.com, but only have TLS configured on bar.com? I think we would need two different VirtualGateways for that? So perhaps this field is essentially redundant?

api/enterprise/networking/v1beta1/virtual_gateway.proto
Sodman: My original thinking was what if there were settings that we need for HTTP TLS that we didn't need for TCP TLS but I didn't end up with any fields in the end. I'm on board with flattening it! 👍

api/enterprise/networking/v1beta1/virtual_gateway.proto
ilackarms: yeah i'm thinking the same thing. i think we can remove it for now, and wait for a request to come for the ability to customize here. otherwise, we just use the server_names

Sodman: Removed sni_domains in favor of server_names at the connection match level.

api/enterprise/networking/v1beta1/virtual_gateway.proto
```diff
@@ -104,22 +98,28 @@ message VirtualGatewaySpec {
             // will be first matched against `www.example.com`, then `*.example.com`, then ``*.com`.
             //
             // Note that partial wildcards are not supported, and values like `*w.example.com` are invalid.
-            repeated string server_names = 2;
+            repeated string server_names = 1;
 
-            // TODO (add more fields from Envoy here)
+            // Optional, if set this will be used for the gateway, otherwise it will be inferred.
```
EItanya, line 103: I think we should explain the logic we use to infer

Sodman: Agreed

api/enterprise/networking/v1beta1/virtual_gateway.proto
```diff
@@ -104,41 +98,41 @@ message VirtualGatewaySpec {
...
             // Note that partial wildcards are not supported, and values like `*w.example.com` are invalid.
-            repeated string server_names = 2;
+            repeated string server_names = 1;
 
-            // TODO (add more fields from Envoy here)
+            // Optional, if set this will be used as the protocol for the gateway, otherwise it will be inferred based on the following logic:
+            // - If the connectionHandler is a HTTP handler and no SslConfig is set in the connectionOptions, use "HTTP"
+            // - If the connectionHandler is a HTTP handler and any SslConfig is set in the connectionOptions, use "HTTPS"
+            // - If the connectionHandler is a TCP handler and no SslConfig is set in the connectionOptions, use "TCP"
+            // - If the connectionHandler is a TCP handler and any SslConfig is set in the connectionOptions, use "TLS"
```
harveyxia, line 107: This doesn't render as a bulleted list, see [here](https://gloo-mesh--pr1771-gateway-tls-laoj31zz.web.app/gloo-mesh/latest/reference/api/github.com.solo-io.gloo-mesh.api.enterprise.networking.v1beta1.virtual_gateway/#virtualgatewayspecconnectionhandlerconnectionmatch). We can just reformat as sequential sentences

api/enterprise/networking/v1beta1/virtual_gateway.proto
```diff
@@ -84,41 +74,45 @@ message VirtualGatewaySpec {
...
+            // Optional, if set this will be used as the protocol for the gateway, otherwise it will be inferred based on the following logic:
+            //
+            // - If the connectionHandler is a HTTP handler and no SslConfig is set in the connectionOptions, use "HTTP"
+            //
+            // - If the connectionHandler is a HTTP handler and any SslConfig is set in the connectionOptions, use "HTTPS"
+            //
+            // - If the connectionHandler is a TCP handler and no SslConfig is set in the connectionOptions, use "TCP"
+            //
+            // - If the connectionHandler is a TCP handler and any SslConfig is set in the connectionOptions, use "TLS"
+            string transport_protocol = 2;
```
EItanya, line 88: Why is this in the match? It doesn't have to do with how the traffic gets filtered by envoy

EItanya (approved):

PR #1794, opened 2021-08-10
remove usages of deprecated `Mesh.Spec.IngressGateways`

harveyxia:


pkg/mesh-networking/translation/istio/mesh/mtls/mtls_translator.go
```diff
@@ -462,6 +480,13 @@ func getPodsToBounce(
 	allWorkloads.List(func(workload *discoveryv1.Workload) bool {
 		kubeWorkload := workload.Spec.GetKubernetes()
 
+		// skip ingress gateway workloads as those have already been added
```
EItanya, line 483: Do we even need to check both the gateways and the workloads separately anymore? The ingress gateways should be captured by this block as well?

harveyxia: I couldn't recall whether we had a reason for bouncing the ingress gateway workloads first—I'll test without that ordering

EItanya: Oh because of ordering, I don't think it matters because we set to 0, but that is a good point. Checking the labels felt confusing to me so I'm wondering if there's a simpler way to check unique

EItanya (approved):

PR #1795, opened 2021-08-11
Include revision in Istio configmap discovery

jahvon:


ci/setup-kind.sh
```diff
@@ -90,7 +90,7 @@ else
   kubectl --context kind-${mgmtCluster} create namespace bookinfo
   kubectl --context kind-${mgmtCluster} label ns bookinfo istio-injection=enabled --overwrite
   kubectl --context kind-${remoteCluster} create namespace bookinfo
-  kubectl --context kind-${remoteCluster} label ns bookinfo istio-injection=enabled --overwrite
+  kubectl --context kind-${remoteCluster} label ns bookinfo istio.io/rev=${istio_revision} --overwrite
```
EItanya, line 93: Can you just comment why this is necessary now

ci/setup-funcs.sh
```diff
@@ -388,6 +388,7 @@ EOF
 function install_istio_1_8() {
   cluster=$1
   eastWestIngressPort=$2
+  revision=$3
```
EItanya, line 391: Can we be consistent with this var naming, either `istioRevision`, `revision`, or `istio_revision`

EItanya (approved):

PR #1800, opened 2021-08-11
decouple workloads from pods

ilackarms:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/989


pkg/mesh-discovery/translation/mesh/detector/istio/istio_mesh_detector.go
```diff
@@ -114,7 +116,7 @@ func (d *meshDetector) detectMesh(
 
 	agent := getAgent(
 		deployment.ClusterName,
-		in.Pods(),
+		in.Deployments(),
```
ilackarms, line 119: this fix will also ensure that detecting the agent workloads will not depend on pods

pkg/mesh-discovery/translation/workload/detector/istio/istio_workload_detector.go
```diff
@@ -0,0 +1,333 @@
...
+	if name == constants.IstioSystemNamespace {
+		return false
+	}
+	if util.IsSystemNamespace(resource.Namespace(name)) {
+		return false
+	}
+
+	injectionLabel := namespace.Labels[util.InjectionLabelName]
+
+	// TODO(ilackarms): store revision instead of a boolean when supported
```
EItanya, line 122: Jahvon has a PR going in right now to support revision, if this goes in will revision stop working?

ilackarms: no, the point is that this will need to be updated when we add support for multiple meshes per cluster (which relies on revisions)

pkg/common/reconciliation/record_reconcile.go
```diff
@@ -109,6 +177,75 @@ func (r *Recorder) RecordReconcileResult(
...
+) {
+	counterName := opts.CounterName
+	labelValues := opts.LabelValues
+	r.customCountersLock.RLock()
+	defer r.customCountersLock.RUnlock()
+	counter, counterRegistered := r.customCounters[counterName]
+
+	if !counterRegistered {
+		// internal error, should never happen
+		contextutils.LoggerFrom(ctx).Warnf("no cunter registered with name %v", counterName)
```
harveyxia, line 215:
```suggestion
		contextutils.LoggerFrom(ctx).Warnf("no counter registered with name %v", counterName)
```

pkg/common/reconciliation/record_reconcile.go
```diff
@@ -131,3 +268,37 @@ func makeReconcileLabels(inputName string, inputGvks []schema.GroupVersionKind,
...
+
+// RecorderFromContext returns the recorder stored int the context.
+// if none is available, a no-op recorder is returned
+func RecorderFromContext(ctx context.Context) Recorder {
+	if ctx != nil {
+		if recorder, ok := ctx.Value(ctxKey{}).(Recorder); ok {
+			return recorder
+		}
+	}
+	return noopRecorder{}
```
harveyxia, line 287: what's the benefit to returning a noopRecorder vs. throwing an error here?

pkg/mesh-discovery/translation/workload/detector/istio/istio_workload_detector.go
```diff
@@ -0,0 +1,333 @@
...
+	istioMesh *v1.MeshSpec_Istio,
+	configMaps corev1sets.ConfigMapSet,
+) (*inject.Config, error) {
+	injectorCm, err := configMaps.Find(&skv1.ClusterObjectRef{
+		Name:        getInjectorConfigMapName(""), // TODO(ilackarms): support mesh revisions here
+		Namespace:   istioMesh.GetInstallation().GetNamespace(),
+		ClusterName: istioMesh.GetInstallation().GetCluster(),
+	})
+	if err != nil {
+		// TODO(ilackarms): no injector configured for this mesh, explore other options for workload detection
```
harveyxia, line 211: we'll want to return an error here so we can track occurrences of this scenario

pkg/common/reconciliation/record_reconcile.go
ilackarms: this is to prevent parts of the code from panicking because they call this on the wrong context

pkg/mesh-discovery/translation/mesh/detector/istio/istio_mesh_detector.go
ilackarms: should not

pkg/mesh-discovery/translation/workload/detector/istio/istio_workload_detector.go
ilackarms: i don't think it's technically an error. istio is able to start up (and will use a default config) for the sidecar injector in this case. we do the same - if no config is found, we default it in a way that mirrors istio

EItanya (approved):

Issue #1791, opened 2021-08-10
meshctl install CRD race condition


harveyxia:
Our OSS e2e tests will sometimes fail during setup due to:

```
error: installing gloo-mesh: installing helm chart: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "Settings" in version "settings.mesh.gloo.solo.io/v1"
```

This is caused by a race condition—the kube apiserver needs some time to install the new CRDs, so we need to wait for that condition before installing our Helm chart (which contains a Settings object). See [this issue](https://github.com/helm/helm/pull/5112) for a description (and fix) of the same problem.

ashleywang1:
Needs backport

PR #1828, opened 2021-08-18
Wait for CRDs to be established in meshctl install

ashleywang1:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh/issues/1791


pkg/meshctl/commands/cluster/register/cluster_register.go
```diff
@@ -44,7 +44,7 @@ The edition registered must match the edition installed on the management cluste
 	)
 
 	opts.addToFlags(cmd.PersistentFlags())
-	cmd.MarkFlagRequired("remote-context")
+	cmd.MarkPersistentFlagRequired("remote-context")
```
ashleywang1, line 47: This was previously not propagating down to the `mehsctl register enterprise` command.

pkg/meshctl/commands/check/agent_check.go
```diff
@@ -97,7 +97,9 @@ func runAgentChecks(ctx context.Context, opts *agentOpts, inCluster bool) error
 		}
 	}
 
-	return nil
+	// Some helm checks need to be run in a pod, and not a job, leaving behind Pods in a completed state.
+	// This can affect future runs of `meshctl check agent` and therefore are cleaned up here.
+	return utils.MustCleanupCompletedPodsInNamespace(ctx, opts.kubeconfig, opts.kubecontext, opts.namespace)
```
ashleywang1, line 102: Some helm checks need to be run in a pod, and not a job, leaving behind Pods in a completed state.
This can affect future runs of `meshctl check agent` and therefore are cleaned up here.

pkg/meshctl/utils/k8s.go
```diff
@@ -19,3 +20,15 @@ func EnsureNamespace(ctx context.Context, kubeClient client.Client, namespace st
 		Spec: corev1.NamespaceSpec{Finalizers: []corev1.FinalizerName{"kubernetes"}},
 	})
 }
+
+func MustCleanupCompletedPodsInNamespace(ctx context.Context, kubeconfig, kubecontext, namespace string) error {
```
harveyxia, line 24: "must" implies termination of the program if any error occurs, whereas in this case you return the error, so I would just call this function `CleanupCompletedPodsInNamespace`

pkg/meshctl/install/helm/installer.go
```diff
@@ -247,7 +250,7 @@ func upsertCrds(ctx context.Context, kubeClient client.Client, chartObj *chart.C
 			}
 		}
 	}
-	return nil
+	return utils.WaitUntilCRDsEstablished(ctx, kubeClient, time.Minute, crdNames)
```
harveyxia, line 253:
```suggestion
        // wait until CRDs are in the "established" status to prevent race conditions when subsequently creating CRs
	return utils.WaitUntilCRDsEstablished(ctx, kubeClient, time.Minute, crdNames)
```

pkg/meshctl/utils/wait.go
```diff
@@ -0,0 +1,56 @@
...
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/rotisserie/eris"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func WaitUntilCRDsEstablished(ctx context.Context, kubeClient client.Client, timeout time.Duration, crdNames []string) error {
```
harveyxia, line 13:
```suggestion
// block until CRD is in the "established" phase to prevent subsequent race conditions when attempting to create CRs
func WaitUntilCRDsEstablished(ctx context.Context, kubeClient client.Client, timeout time.Duration, crdNames []string) error {
```

pkg/meshctl/utils/wait.go
```diff
@@ -0,0 +1,56 @@
...
+			if len(notYetEstablished) == 0 {
+				return nil
+			}
+		}
+	}
+}
+
+func crdEstablished(ctx context.Context, kubeClient client.Client, crdName string) (bool, error) {
+	existingCrd := &v1beta1.CustomResourceDefinition{}
+	if err := kubeClient.Get(ctx, client.ObjectKey{Name: crdName}, existingCrd); err != nil {
```
harveyxia, line 40: can we use the `CustomResourceDefinitionClient` from [external-apis](https://github.com/solo-io/external-apis/blob/ecc2e32be1368de440160a21b0e4ccab8957970e/pkg/api/k8s/apiextensions.k8s.io/v1beta1/clients.go#L129)?

pkg/meshctl/utils/wait.go
```diff
@@ -0,0 +1,56 @@
...
+		return false, err
+	}
+	for _, cond := range existingCrd.Status.Conditions {
+		switch cond.Type {
+		case v1beta1.Established:
```
harveyxia, line 45: can a CRD have an "established" condition and also a naming conflict?

harveyxia: If all we care about is that the CRD exists, then we need to remove L50-53

harveyxia (approved):

PR #1830, opened 2021-08-18
add guide for dashboard authentication

ryantking:


docs/content/guides/accessing_enterprise_ui.md
```diff
@@ -93,6 +93,48 @@ The **Debug** section allows you to view and download the full configuration of
 
 You can use this view to quickly ascertain information about a particular mesh or to capture the current configuration of a virtual mesh.
 
+## Securing the Dashboard
+
+The dashboard supports OIDC authentication. Common providers include Google, Okta, and Auth0 among many others.
+Its is configurable via Helm values.
```
ashleywang1, line 99:
```suggestion
It is configurable via Helm values.
```

docs/content/guides/accessing_enterprise_ui.md
```diff
@@ -93,6 +93,48 @@ The **Debug** section allows you to view and download the full configuration of
...
+A full list of Values can be found [here]({{% versioned_link_path fromRoot="/reference/helm/gloo_mesh_enterprise/latest/dashboard_helm_values_reference/" %}}).
+
+Below is a basic set of Helm valuer to install Gloo Mesh Enterprise with the secure dashboard enabled.
```
ashleywang1, line 102:
```suggestion
Below is a basic set of Helm values to install Gloo Mesh Enterprise with the secure dashboard enabled.
```

docs/content/guides/accessing_enterprise_ui.md
```diff
@@ -93,6 +93,48 @@ The **Debug** section allows you to view and download the full configuration of
...
+
```
artberger, line 103: Maybe switch the order of these lines. 

```suggestion

Copy and fill out the following YAML file to set up OIDC for the dashboard. To customize other parts of the dashboard, review the full list of Helm values [here]({{% versioned_link_path fromRoot="/reference/helm/gloo_mesh_enterprise/latest/dashboard_helm_values_reference/" %}}).

```

artberger: A couple other questions... 
* How would they use the file...by passing this in during the Helm installation? 
* If so, does the file have to have a particular name? (I did try it with the `helm install -f test-meshdb.yaml...` command  but the pods didn't spin up for me)
* Can they do this if they already have Gloo Mesh Enterprise installed, maybe by the `helm upgrade` command?

Depending on the answer, how about:
```suggestion


Copy and fill out the following YAML file to set up OIDC for the dashboard. To customize other parts of the dashboard, review the full list of Helm values [here]({{% versioned_link_path fromRoot="/reference/helm/gloo_mesh_enterprise/latest/dashboard_helm_values_reference/" %}}).

After filling out the YAML file with your OIDC provider details, install Gloo Mesh Enterprise with the Helm file. For example, after setting the Kubernetes context to your cluster, you can run the following command. Include `[--install]` if you do not already have Gloo Mesh Enterprise installed in the cluster. 

`helm upgrade -f oidc.yaml gloo-mesh-enterprise gloo-mesh-enterprise/gloo-mesh-enterprise --namespace gloo-mesh [--install]`

```

docs/content/guides/accessing_enterprise_ui.md
```diff
@@ -93,6 +93,48 @@ The **Debug** section allows you to view and download the full configuration of
...
+    enabled: true
+    backend: oidc
+    oidc:
+      clientId: # From the OIDC provider
+      clientSecret: # From the OIDC provider (will be stored in secret)
+      clientSecretRef:
+        name: dashboard
+        namespace: gloo-mesh
+      issuerUrl: # URL of the issuer, usually something like https://<domain>.<provider url>/
+      appUrl: # URL the dashboard will be accessible at, will be used for redirects
```
artberger, line 118: Does this also come from the OIDC provider, or how do they find the app URL?

docs/content/guides/accessing_enterprise_ui.md
```diff
@@ -93,6 +93,48 @@ The **Debug** section allows you to view and download the full configuration of
...
+  auth:
+    enabled: true
+    backend: oidc
+    oidc:
+      clientId: # From the OIDC provider
+      clientSecret: # From the OIDC provider (will be stored in secret)
+      clientSecretRef:
+        name: dashboard
+        namespace: gloo-mesh
+      issuerUrl: # URL of the issuer, usually something like https://<domain>.<provider url>/
```
artberger, line 117:
```suggestion
      issuerUrl: # The issuer URL from the OIDC provider, usually something like https://<domain>.<provider url>/
```

docs/content/guides/accessing_enterprise_ui.md
```diff
@@ -93,6 +93,48 @@ The **Debug** section allows you to view and download the full configuration of
...
+      appUrl: # URL the dashboard will be accessible at, will be used for redirects
+```
+
+Optionally, Redis can be used as a session backend if the ID tokens returned by the OIDC provider are too large to be stored in a cookie.
+The following values can be added to the ones above to enable Redis. The example values assume the Redis that ships as part of the chart is used,
+but users can disable it and adjust the values accordingly to use their own.
```
artberger, line 123: I wasn't sure if the last two sentences meant that they could use things other than Redis, or just they could set it to their own Redis (but it has to be Redis)?

If it means something other than Redis, how about:

```suggestion

**Setting a session backend for large ID tokens**

If the ID tokens that the OIDC provider returns are too large to be stored in a cookie, you can set a session backend to a data store such as Redis. To set a session backend, the data store must be part of the Helm chart that you use. Add the following `session` values to the auth YAML file that you previously created, replacing Redis with your own host values.

```

artberger (commented):
Hi @ryantking thanks for opening this PR and tagging us for review. I left some questions and suggestions from a docs perspective, but please let me know if this is not the sort of feedback you were looking for, as this is the first review I've done here!

docs/content/guides/accessing_enterprise_ui.md
ryantking: We support passing custom Helm values into the meshctl installation command so either can be used.

artberger (approved):

PR #1836, opened 2021-08-20
Doc updates for snake_case to camelCase

artberger:
For https://github.com/solo-io/gloo/issues/5151, follow up to https://github.com/solo-io/gloo/pull/5198

Updated snake case to camel case for fields in Gloo resources. Method I took was:

1. Search doc folder for YAMLs with a solo.io API version `(^apiVersion).*(solo)` so I'd only find Gloo resources.
2. Search those YAML snippets for underscores to see snake_case fields.
3. If any snake cases were found, compare with the CRDs or the -o yaml output of the resource in my cluster to double-check that they should be camelCase.


Sodman (approved):
LGTM! 👍

PR #1837, opened 2021-08-20
Virtual gateway integration tests

nmnellis:


Sodman:
This doesn't feel like the right repo for these tests to live, considering these gateway features under test here require Gloo Mesh Enterprise or Gloo Mesh Gateway to work, and can't run in an OSS Gloo Mesh install.

PR #1847, opened 2021-08-24
Peer auth api

mlholland:
API changes to allow peer auth controls via GME


api/networking/v1/traffic_policy.proto
```diff
@@ -212,6 +212,8 @@ message TrafficPolicySpec {
             message Istio {
 
                 // TLS connection mode
+                // Note: If this is set to STRICT, and settings.spec.peerAuth.peerAuthTlsMode is UNSET (or vise verse), translation will fail, since
+                // we cannot both mandate and disable TLS at different junctures.
```
artberger, line 216: I'm not sure if these comments are used to generate docs, but if so, I was wondering what the "translation" and "juncture" parts mean...and how they could get it to work? Maybe something like this:
```suggestion
                // Note: If this setting is set to STRICT, but settings.spec.peerAuth.peerAuthTlsMode is UNSET (or vice versa), the connection fails because
                // mutual TLS cannot be enforced in one place but unenforced in another place. Make sure these settings match.
```

api/settings/v1/settings.proto
```diff
@@ -28,8 +28,45 @@ message SettingsSpec {
...
+
+    // Enable the creation of PeerAuthentications on meshes to manage TLS connections.
+    // Defaults to true.
+    bool enabled = 1;
+
+    // The default mutualTls mode for automatically generated Peer Authentications to use.
+    // The enum values correspond to the values listed [by istio](https://istio.io/latest/docs/reference/config/security/peer_authentication/#PeerAuthentication-MutualTLS-Mode)
+    // Defaults to UNSET, which behaves like PERMISSIVE.
+    // Note: If this is set to STRICT, and settings.mtls.istio.tlsMode is UNSET (or vise verse), translation will fail, since
+    // we cannot both mandate and disable TLS at different junctures.
```
artberger, line 49: Similar comment:
```suggestion
    // Note: If this setting is set to STRICT, but settings.mtls.istio.tlsMode is UNSET (or vice versa), the connection fails because
    // mutual TLS cannot be enforced in one place but unenforced in another place. Make sure these settings match.
```

api/settings/v1/settings.proto
```diff
@@ -28,8 +28,45 @@ message SettingsSpec {
...
+    MutualTLS peer_auth_tls_mode = 2;
+
+    // Istio MutualTLS settings
+    enum MutualTLS {
+
+        // Inherit from parent, if has one. Otherwise treated as PERMISSIVE.
```
artberger, line 55: Is it clear what the parent setting is? If not, maybe we can just say what it is. 

Also is it just the number that is inherited, or like the entire setting of UNSET vs DISABLE vs PERMISSIVE vs STRICT?
```suggestion
        // Inherited from parent, if any. Otherwise treated as PERMISSIVE.
```

api/settings/v1/settings.proto
```diff
@@ -28,8 +28,45 @@ message SettingsSpec {
...
+        UNSET = 0;
+
+        // Connection is not tunneled.
```
artberger, line 58: Is this the same thing as the plaintext mentioned in the PERMISSIVE?
```suggestion
        // Connection is plaintext, and not encrypted through an mTLS tunnel.
```

api/settings/v1/settings.proto
```diff
@@ -28,8 +28,45 @@ message SettingsSpec {
...
+        DISABLE = 1;
+
+        // Connection can be either plaintext or mTLS tunnel.
+        PERMISSIVE = 2;
+
+        // Connection is an mTLS tunnel (TLS with client cert must be presented).
```
artberger, line 64:
```suggestion
        // Connection is encrypted through an mTLS tunnel (TLS with client cert must be presented).
```

changelog/v1.2.0-beta2/peer-auth-api.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/gloo-mesh-enterprise/issues/833
+    resolvesIssue: false
+    description: Add api changes for peer auth creation.
```
artberger, line 5: Would it be important to mention the matching TLS settings?
```suggestion
    description: Add API changes for creating peer authentication, including matching mTLS settings for Istio mesh connections.
```

artberger (commented):
Thanks for opening this PR! I left some comments from a docs perspective.

Sodman:
this CRD was moved to GME

PR #1851, opened 2021-08-25
Docs - Rephrasing upgrade so link is out of note

artberger:
Because formatting doesn't work in the note


artberger:
Confirmed it built in the preview doc:

image

Rachael-Graham (approved):

PR #1852, opened 2021-08-25
Upgrade docs pre v1.0.0

ashleywang1:


docs/content/operations/upgrading.md
```diff
@@ -5,7 +5,11 @@ weight: 100
 description: Best practices for upgrading Gloo Mesh
 ---
 
-Gloo Mesh is pre-1.0 and under active development. There may be breaking API changes with every minor version bump.
+{{< notice warning >}}
+This upgrade process is for Gloo Mesh, from a pre-1.0.0 state up to v1.0.0.
```
artberger, line 9: Does it also apply to Enterprise or just the OSS?
```suggestion
This upgrade process is for Gloo Mesh versions less than v1.0.0 up to v1.0.0.
```

ashleywang1: Just the OSS - the enterprise version did not have any upgrade docs that I'm aware of.

docs/content/operations/upgrading.md
```diff
@@ -5,7 +5,11 @@ weight: 100
...
+{{< /notice >}}
+
+Gloo Mesh is under active development. There may be breaking API changes with every minor version bump.
```
Sodman, line 12: This is no longer true. Post 1.0.0 we are committed to not having breaking API changes between minor versions!

```suggestion
Gloo Mesh is under active development.
```

docs/content/operations/upgrading.md
```diff
@@ -5,7 +5,11 @@ weight: 100
 description: Best practices for upgrading Gloo Mesh
 ---
 
-Gloo Mesh is pre-1.0 and under active development. There may be breaking API changes with every minor version bump.
+{{< notice warning >}}
+This upgrade process is for Gloo Mesh versions less than v1.0.0 up to v1.0.0.
```
ashleywang1, line 9:
```suggestion
This upgrade process is for Gloo Mesh versions less than v1.0.0 up to, but not including, v1.0.0.
```

PR #1904, opened 2021-09-30
Rewrite community focused meshctl

ryantking:
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/1030
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/1927

PR #1922, opened 2021-11-16
openmeshctl- fix e2e tests

arianaw66:
Branches off https://github.com/solo-io/gloo-mesh/pull/1904 to fix tests.

High level changes:
- removes testing on Istio 1.7
- updates `openmeshctl register/deregister` to use cluster name as context name by default, as documented
- ensures `install`, `register`, `deregister`, and `uninstall` all occur in the intended kubeContext
- updates `helm` to use secrets to keep track of releases, to persist across commands
- fixes several ignored flags

BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/1927


pkg/openmeshctl/commands/utils/utils.go
```diff
@@ -0,0 +1,32 @@
...
+	"os/exec"
+
+	"github.com/gobuffalo/packr"
+	"github.com/rotisserie/eris"
+)
+
+var box = packr.NewBox("./scripts")
+
+func SwitchContext(kubeCtx string) error {
+	if err := runScript(box, os.Stdout, "context_switch.sh", kubeCtx); err != nil {
```
Sodman, line 15: Longer term I'm a little worried about using this to switch kube contexts as it could potentially affect other processes running on the system relying on the default kube context, but this is how it worked before this PR so updating that is out of scope here.

Sodman (approved):
LGTM! 🚀

pkg/openmeshctl/commands/utils/utils.go
arianaw66: Absolutely agree with that. I spent a bit of time trying to get the helm install to pay attention to the kubeContext passed in but didn't succeed and went with the approach we were already using.

arianaw66: Issue to address this down the line: https://github.com/solo-io/gloo-mesh/issues/1925

PR #1929, opened 2021-12-14
Switch to forms for new issues

kcbabo:
Moving to the beta forms support in GitHub to increase the consistency and quality of issue reports across all types. You can preview what the new issue L&F here:
https://github.com/kcbabo/test-issues/issues/new/choose


.github/ISSUE_TEMPLATE/FEATURE-ISSUE.yml
```diff
@@ -0,0 +1,36 @@
...
+    attributes:
+      label: Version
+      description: What version of Gloo Mesh are you using?
+      options:
+        - 1.3.x 
+        - 1.2.x
+        - 1.1.x
+        - 1.0.x
+    validations:
+      required: true
```
Sodman, line 16: Version might not matter for feature requests, I think we could remove this as a required field? (We typically only backport bugfixes, building new features for the latest release branch).

.github/ISSUE_TEMPLATE/BUG-ISSUE.yml
```diff
@@ -0,0 +1,49 @@
...
+      required: true
+  - type: textarea
+    id: additional-context
+    attributes:
+      label: Additional Context
+      placeholder: |
+        - Customer or prospect
+        - Gloo Mesh version
+        - Kubernetes version
+        - service mesh and version (if applicable, e.g. Istio 1.7)
```
Sodman, line 49: Can we maybe add "Istio Version" as it's own separate (optional) dropdown field? This is becoming increasingly relevant as Istio adds new features that we can't support in older versions, or when users are reporting bugs with our Istio Lifecycle development features etc. For Istio version the granularity can just be minor version, eg 1.9.x, 1.10.x, etc.

.github/ISSUE_TEMPLATE/DOCS-ISSUE.yml
```diff
@@ -0,0 +1,40 @@
...
+  - type: dropdown
+    id: version
+    attributes:
+      label: Version
+      description: What version of Gloo Mesh are you using?
+      options:
+        - 1.3.x 
+        - 1.2.x
+        - 1.1.x
+        - 1.0.x
```
Sodman, line 14: If we're talking about the docs site specifically here, I think it makes sense for the options to mirror the site, which would currently be:
- main
- 1.2.x (latest stable)
- 1.1.x

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,8 @@
+blank_issues_enabled: false
+contact_links:
+  - name: Talk with Us on Slack
+    url: https://slack.solo.io
+    about: For a fast response, consider asking your question on the Solo slack.
+  - name: Solo.io Documentation
+    url: https://docs.solo.io/gloo-mesh-enterprise/latest/operations/troubleshooting
+    about: Check out the Troubleshooting session of the Gloo Mesh docs.
```
Sodman, line 8: typo?
```suggestion
    about: Check out the Troubleshooting section of the Gloo Mesh docs.
```

.github/ISSUE_TEMPLATE/BUG-ISSUE.yml
```diff
@@ -0,0 +1,49 @@
...
+      placeholder: When I performed x, I expected y to happen
+    validations:
+      required: true
+  - type: textarea
+    id: additional-context
+    attributes:
+      label: Additional Context
+      placeholder: |
+        - Customer or prospect
+        - Gloo Mesh version
```
Sodman, line 47: We can probably remove Gloo Mesh version from this list if we have it above.

.github/ISSUE_TEMPLATE/BUG-ISSUE.yml
```diff
@@ -0,0 +1,49 @@
...
+labels: ["Type: Bug"]
+body:
+  - type: dropdown
+    id: version
+    attributes:
+      label: Version
+      description: What version of Gloo Mesh are you using?
+      options:
+        - 1.3.x 
+        - 1.2.x
```
Sodman, line 12: What do you think about annotating these to make it clear to users, just incase they assume the top one is the latest one they installed from default via meshctl?
```suggestion
        - 1.3.x (beta)
        - 1.2.x (latest stable)
```

Sodman (commented):
Left a few minor comments but overall LGTM!

.github/ISSUE_TEMPLATE/FEATURE-ISSUE.yml
kcbabo: Sounds good to me. Some considerations in my mind for keeping it required:

- It's pretty painless in terms of data entry. Takes a couple seconds to select the version from a dropdown.
- It's always good to have background info on the version that a user is running. 
- There may be cases where a feature exists on a newer version and the user is on a prior version. Allows for quick turnaround for resolution: "upgrade to x.y.z". :-)

100% agree that we don't want to set an expectation that features/enhancements will be targeted at a specific version other than latest. I'm good with version being required or optional - just throwing out some additional considerations.

.github/ISSUE_TEMPLATE/BUG-ISSUE.yml
kcbabo: I like that idea. We should probably come up with a README or something that talks about what needs to change when a release is cut (adding new version, adjusting annotations, etc.). Ideal world, that would be automated as part of a release. :-)

.github/ISSUE_TEMPLATE/BUG-ISSUE.yml
kcbabo: Added

kcbabo:
Thanks for the review and great feedback, @Sodman. I believe I have everything incorporated with the latest commits on my PR branch.

.github/ISSUE_TEMPLATE/FEATURE-ISSUE.yml
Sodman: Sure it doesn't hurt to start with it required. We can always adjust later if it becomes a nuisance.

Sodman (approved):

PR #1954, opened 2022-12-15
Fix run script for meshctl installation to get latest

Sodman:


ashleywang1 (approved):

hack/install.sh
```diff
@@ -14,7 +14,7 @@ else
...
+  GLOO_MESH_VERSIONS=$(curl -s https://storage.googleapis.com/storage/v1/b/meshctl/o?pageToken=CiB2Mi4xLjAtYmV0YTIvbWVzaGN0bC1saW51eC1hbWQ2NA | any_python -c "import sys; from distutils.version import StrictVersion, LooseVersion; from json import loads as l; obj = l(sys.stdin.read()); items = obj['items'];  all_releases = [release['name'].split('/')[0] for release in items]; releases = []; [releases.append(n) for n in all_releases if n not in releases]; filtered_releases = list(filter(lambda release_string: len(release_string) > 0 and StrictVersion.version_re.match(release_string[1:]) != None, releases)); filtered_releases.sort(key=LooseVersion, reverse=True); print('\n'.join(filtered_releases))")
```
fernandosanchezs, line 17: Is this fetching a specific page (hence the constant pageToken) to get this version?

nrjpoddar: How stable is the pagination token?

Sodman: @fernandosanchezs - This is currently page 2, yes.

@nrjpoddar I can't find any documentation that clarifies either way. In hindsight perhaps a better route would be to do something like `?maxResults=2000` (default 1000) What do you think?

nrjpoddar: yes
No magic tokens ever in a script..

